PR #2, opened 2017-11-15
graph entities clean up, refactor, formatted with scalafmt

MarcoLotz:
Reduced the redundant code on Graph entities
Refactored some method
Changed some log management to be tail-recursive
added formating

Please take a look at the comments in the commit and check if the solution still performs as expected - it looked fine here :)


miratepuffin:
RE facebook chat, because of the use of the Kill/revive functions (that have been renamed) in other classes if these could be changed back I can merge it. Once that is done we can change the names back or agree on some new ones that better fit :)

PR #59, opened 2019-11-04
Entity Storage Revamp

miratepuffin:
Removed Edge Map Monolith - All edges now stored in their source and destination nodes.
Turned Entity Storage into a class that workers are allocated -- separates all workers and means a lot less shared state.
Removed WorkerID and Router ID being passed to all functions to insure no incorrect maps were accessed.


miratepuffin:
Closes #40

PR #119, opened 2021-01-06
Fix bugs in Router

wuliaososhunhun:
- change behaviour of `Spout` so it will send `DataFinished` once to each `Router`. It no longer sends `NoWork` after finished so that `Route` no longer asks `WorkPlease` again once data finished
- Before a `Router` finished, when received `DataSync` from other completed `Router`, it will hold best time as `restRouterNewestTime` in state. When it finishes, the `Router` will consume the `restRouterNewestTime`[ Useless after this step as we will use `newestTime` in state as the source] and publish `DataSync` to all `Router`s and let `Writer` know if necessary. After that, when received `DataSync` from other completed `Router` again, it will update `newestTime` if necessary and then update `Writer`s.
- other minor optimisations


miratepuffin:
Wicked thanks Yanyang, seems to solve some of the issues, I will merge in and then see how the deletions and so on react!

PR #120, opened 2021-01-20
adding utility scripts to generate queries and to extract the results

niallroche:
added some scripts that I found useful for generating the curl commands for different time windows and algorithms for analysis and then extracting them from mongoDB


miratepuffin:
Looks Great Niall, Cheers for adding these in!

PR #127, opened 2021-02-01
extract sending message logic out from EntityStorage + watchdog refactoring

wuliaososhunhun:


miratepuffin:
Individual commits included via cherry-pick due to unfortunate divergence with the base code.

PR #131, opened 2021-02-12
Clean some messages

wuliaososhunhun:
- move some system messages to different locations
- optimise startUp of RouterWorker
- rename effectMessage to GraphEffect
- remove few unused message
- unify PartitionsCount related message


miratepuffin:
Compared to Citation and all looking good!

PR #132, opened 2021-02-17
Clean and refactor GraphUpdate

wuliaososhunhun:


miratepuffin:
New API is looking clean! Thank as every @wuliaososhunhun

PR #138, opened 2021-03-03
refactor and move reader related messages

wuliaososhunhun:


miratepuffin:
Thanks as every Yangyang, have checked through everything and will be including in the next release.

PR #147, opened 2021-03-25
Refactor api

wuliaososhunhun:


miratepuffin (approved):
Great stuff @wuliaososhunhun, Everything seems to be working as expected. There is an issue with the new exportResults function, but this is one my side if you are trying to submit jobs and not getting results.

Issue #124, opened 2021-01-22
Move analytical state into the vertex visitors


miratepuffin:
Once #123 is complete then the vertex visitors it creates may store the analytical state and messages for each algorithmic run and then be cleanly dropped afterwards stopping them from building up over time. We can also at this point change the way the keys work for analytical state removing the large string component.

Issue #123, opened 2021-01-22
Generate semi permanent flattenings 


miratepuffin:
Currently flattening's are generated for each superstep and dropped as soon as it is done. This is done as the generation process is fairly quick and it seemed a better idea than storing all of the visitors. However it appears that generating them over and over is creating a lot of junk in the heap and by keeping a semi-permanent object until the end of an algorithm we can cleanly drop all temporary state.

This may make running multiple jobs at once a bit more complicated, as the flattening's will need to be stored separately, but we will just have to cross that bridge when we come to it.

Issue #122, opened 2021-01-22
Reduce the number of message queues each vertex has 


miratepuffin:
Following issue #121 we will be able to remove the need to specify the window or timestamp components of the vertex messaging queues, drastically cutting the number of arrays which are left hanging in the vertex multi queue.

Issue #121, opened 2021-01-22
Run windows for batches to completion serially


miratepuffin:
Currently batched window jobs are run for all windows at each superstep, this should be each window to completion and then start on the next. This will reduce the complexity of the message handling in the vertex, reduce the amount of messages stored at any one point and reduce the number of required mailboxes.

PR #149, opened 2021-04-05
refactor analysis

wuliaososhunhun:
Too many changes to summarise...


miratepuffin (commented):
Just had an initial scroll through and ticked off a couple of the smaller affected files. Bigger changes looks great based on the discussions we had over the weekend. I will pull it all down now and start testing. Imagine we may need to pull this into a separate branch for a little while to work back and forth on it.

miratepuffin:
Oh also as a secondary point, we should probably enumerate the changes at some point, just to track them all for the next major version going into master.

miratepuffin:
**Bits I fixed:**
- Fixed a couple of minor issues with the graph lens to make sure it returned a set of vertices. (Map to collect) 
- Fixed the RaphtoryGraph App so that it would actually accept Analysers of any type.

Few problems noted:

- For some reason the Analysis Tasks are receiving info about the number of partitions. Not really an issue, but they should handle this correctly if we need to them to.
> `01:17:19.151 [Citation-system-akka.actor.default-dispatcher-3] ERROR c.r.c.a.A.T.V.ViewAnalysisTask - Not handled message in check reader worker: SubscribeAck(Subscribe(/partitionsCount,None,Actor[akka://Citation-system/user/ViewAnalysisTask_com.raphtory.algorithms.DegreeBasic_1617841039133#-172809655]))`
- Batches of windows not currently working, but this is because we turned it off.
- Algorithms all need to be updated with the `new` API -- To be decided if we fully change over or keep this in-between state for a while.
- Output seems inconsistent at the moment, digging into this in the morning.

miratepuffin:
Running the `AllCommandsTest` class until all data is ingested (your laptop stops screaming) then submitting the following query:
>curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","analyserName":"com.raphtory.algorithms.ConnectedComponents","start":1,"end":300000,"jump":10000,"windowSet":[1000000,100000,10000,1000,100]}' 127.0.0.1:8081/RangeAnalysisRequest

- Seems that the above issue of partition count blocks the analysis going forward.
>01:42:59.346 [Raphtory-akka.actor.default-dispatcher-28] ERROR c.r.c.a.A.T.R.RangeAnalysisTask - Not handled message in check reader worker: SubscribeAck(Subscribe(/partitionsCount,None,Actor[akka://Raphtory/user/RangeAnalysisTask_com.raphtory.algorithms.ConnectedComponents_1617842579104#1078059270]))
01:43:14.292 [Raphtory-akka.actor.default-dispatcher-33] ERROR c.r.c.a.A.T.R.RangeAnalysisTask - Not handled message in check time: PartitionsCount(10)
01:44:14.314 [Raphtory-akka.actor.default-dispatcher-18] ERROR c.r.c.a.A.T.R.RangeAnalysisTask - Not handled message in check time: PartitionsCount(10)
01:45:14.328 [Raphtory-akka.actor.default-dispatcher-27] ERROR c.r.c.a.A.T.R.RangeAnalysisTask - Not handled message in check time: PartitionsCount(10)
01:46:14.347 [Raphtory-akka.actor.default-dispatcher-25] ERROR c.r.c.a.A.T.R.RangeAnalysisTask - Not handled message in check time: PartitionsCount(10)

-interestingly this doesn't always seem to happen, got the results out for a Degree Basic and one Connected Components. Though these results were empty i.e. had no vertices or edges included.

src/main/scala/com/raphtory/core/actors/PartitionManager/Workers/AnalysisSubtaskWorker.scala
```diff
@@ -0,0 +1,126 @@
...
+
+  private def work(state: State): Receive = {
+    case StartSubtask(_, timestamp, window) =>
+      log.debug(s"Job [$jobId] belonging to ReaderWorker [$workerId] Reader [$managerId] is SetupTaskWorker.")
+      val beforeTime = System.currentTimeMillis()
+      val initStep   = 0
+      val graphLens  = GraphLens(jobId, timestamp, window, initStep, workerId, storage)
+      analyzer.sysSetup(graphLens, workerId)
+      val messages = analyzer.view.getAndCleanMessages()
+      messages.foreach(m => mediator ! new DistributedPubSubMediator.Send(getReader(m.vertexId, state.managerCount), m))
```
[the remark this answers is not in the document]
miratepuffin: This refers to line 45, not sure why it added so much.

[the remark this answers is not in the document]
wuliaososhunhun: we are sending the message need to be sent. We cannot improve this by magic. It just send during the `analyzer` setup or analyse. Technically, we just move the from there to the last step. There is no difference from performance perspective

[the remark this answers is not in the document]
miratepuffin: My point here is that for these messages we have to do some IO, which going to be slower than pure processing. If we can push this into the background whilst still processing other vertices the messages can be travelling across the network `for free'. In the instance above we are blocking all IO until the processing is done and, I imagine, will have to wait for it to synch before continuing. I believe this is the same manner for many distributed systems, i.e. the shuffle in hadoop is happening alongside completing map tasks.

src/main/scala/com/raphtory/core/actors/PartitionManager/Workers/AnalysisSubtaskWorker.scala
```diff
@@ -41,6 +41,7 @@ final case class AnalysisSubtaskWorker(
       val initStep   = 0
       val graphLens  = GraphLens(jobId, timestamp, window, initStep, workerId, storage)
       analyzer.sysSetup(graphLens, workerId)
+      analyzer.setup()
```
miratepuffin, line 44: Hahaha yeah this is probably an important one to have in :D

src/main/scala/com/raphtory/core/actors/PartitionManager/Workers/AnalysisSubtaskWorker.scala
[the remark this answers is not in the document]
miratepuffin: I can see where you are coming from, but I would be surprised if it is faster. Lets put a pin in it for the second, get everything working and then we can test it to see which works better. Then the winner can gloat 😄

miratepuffin:
Fixed Below as by setting the reader workers to create a new copy of the analyser. As in the Raphtory Graph everything is passed by reference they were all accessing the same one and therefore the graph lens for one worker was accessible to all.
>java.util.NoSuchElementException: key not found: -1799065806
	at scala.collection.parallel.ParMapLike.default(ParMapLike.scala:40) ~[scala-library-2.12.4.jar:na]
	at scala.collection.parallel.ParMapLike.default$(ParMapLike.scala:40) ~[scala-library-2.12.4.jar:na]
	at scala.collection.parallel.mutable.ParTrieMap.default(ParTrieMap.scala:36) ~[scala-library-2.12.4.jar:na]
	at scala.collection.parallel.ParMapLike.apply(ParMapLike.scala:46) ~[scala-library-2.12.4.jar:na]
	at scala.collection.parallel.ParMapLike.apply$(ParMapLike.scala:44) ~[scala-library-2.12.4.jar:na]
	at scala.collection.parallel.mutable.ParTrieMap.apply(ParTrieMap.scala:36) ~[scala-library-2.12.4.jar:na]
	at com.raphtory.core.analysis.GraphLenses.GraphLens.receiveMessage(GraphLens.scala:76) ~[raphtory_2.12-latest.jar:latest]

miratepuffin:
Seemed to have fixed the issues with getting stuck in the `checkMessage` loop. 
LOTR example now outputs:
>{"top5":[76,3,2,2],"total":5,"totalIslands":1,"proportion":0.9047619,"clustersGT2":2}
Which I have no idea if is correct, but is better than nothing!

src/main/scala/com/raphtory/core/actors/AnalysisManager/Tasks/AnalysisTask.scala
```diff
@@ -86,7 +86,7 @@ abstract class AnalysisTask(
     case ClassMissing => log.error("unexpected missing")
     case ClassExists =>
       if (readyCount + 1 == managerCount) {
-        messageToAllReaderWorkers(LoadPredefinedAnalyser(jobId, analyser))
+        messageToAllReaderWorkers(LoadPredefinedAnalyser(jobId, analyser, args))
```
wuliaososhunhun, line 89: why you need this? you just pass the the build analyser instance directly to the target. You do not need to build another one I think

src/main/scala/com/raphtory/core/analysis/GraphLens.scala
```diff
@@ -73,5 +73,13 @@ final case class GraphLens(
   }
   def vertexVoted(): Unit = voteCount += 1
   def nextStep(): Unit    = superStep += 1
-  def receiveMessage(msg: VertexMessage): Unit = vertexMap(msg.vertexId).receiveMessage(msg)
+  def receiveMessage(msg: VertexMessage): Unit = {
+    try{
```
wuliaososhunhun, line 77: It works but the main question is why this happens? why it can receive a message should have the vertex but the vertex is not there

src/main/scala/com/raphtory/core/analysis/entity/Vertex.scala
```diff
@@ -106,6 +106,6 @@ final case class Vertex(
 
   // todo hide
   def receiveMessage(msg: VertexMessage): Unit = {
-    multiQueue.receiveMessage(view.superStep, msg)
+    multiQueue.receiveMessage(view.superStep, msg.data)
```
wuliaososhunhun, line 109: this makes sense

src/main/scala/com/raphtory/core/analysis/GraphLens.scala
miratepuffin: This can be deleted, I thought I removed this after testing.

miratepuffin:
As a side note whilst inserting the initial version of the serialiser. I think some additional information would be good reported back at the moment. (I need to double check if this is just at a different logging level) but elements like if a job is submitted, if it is waiting for the view to become available, when it is completed etc.

miratepuffin:
New Aggregate Serialiser has been added in. Needs a little bit of tidy up as well as the example mongo one.
I need to:
-  pass the real view time to the serialiser as currently being saved only to kamon
- test the REST API works with the serialisers 
- Run the all commands test and results compare now that the output of this branch is comparable.

miratepuffin:
Added the view time quickly before bed.  Run a quick test on the gab graph with the following query: 
  
> rg.rangeQuery(ConnectedComponents(),new` MongoSerialiser, start= 1470797917000L,end = 1476113868000L,increment = 86400000L,windowBatch=List(3600000L,86400000L,604800000L,2592000000L,31536000000L),arguments)

Seemed to run correctly (in that all timestamps and windows are executed. Results seem slightly off, going to compare to master in the morning. Example of Results below:

`{"proportion":0.8576058,"clustersGT2":28,"totalIslands":317,"total":394,"top5":[4499,159,26,20,13],"viewTime":3517,"time":1476068317000,"windowSize":2592000000}`

`{"proportion":0.8148559,"clustersGT2":16,"totalIslands":218,"total":266,"top5":[2205,163,10,6,5],"viewTime":608,"time":1476068317000,"windowSize":604800000}`

`{"proportion":0.8155803,"clustersGT2":7,"totalIslands":88,"total":133,"top5":[1026,43,6,6,5],"viewTime":163,"time":1476068317000,"windowSize":86400000}`

`{"proportion":0.6037736,"clustersGT2":3,"totalIslands":2,"total":21,"top5":[64,4,4,2,2],"viewTime":68,"time":1476068317000,"windowSize":3600000}`

miratepuffin:
- Fixed the messaging errors by removing the message queue, which wasn't thread safe for the partriemap. 
- Results are now consistent but still wrong for the all commands test
- Trying to compare the state to see if the random generator is outputting a different graph, but this has some deadlock it appears
- Fixing state test now and will hopefully lead to the source of the incorrect connected components.

miratepuffin:
image
Seems the Random generator is slightly different now, in a new compare with the master branch the results are almost the same. The views are working the same, but messaging seems to be broken.

src/main/scala/com/raphtory/core/analysis/GraphLens.scala
```diff
@@ -48,29 +49,32 @@ final case class GraphLens(
...
+      case (k, vertex) if vertex.hasMessage() => vertex
+    }
     viewTimer.update(System.currentTimeMillis() - startTime)
+    vertexCount.set(result.size)
     result
   }
 
-  def getVertices(): ParIterable[Vertex] = vertexMap.map(x=>x._2)
-
-  def checkVotes(): Boolean =
```
miratepuffin: This was a pretty poorly optimised function and also would have been returning the wrong results as after the vertex pulls its queue it would have been emptied.

miratepuffin:
Seems to still be some issues with the smaller windows, namely that the components of 2/3 in size aren't being reported as connected. It seems a lot worst now that I have fixed the voteCount, which either means that I have actually broken more things or the above was hiding a larger issue.

src/main/scala/com/raphtory/core/analysis/GraphLens.scala
```diff
@@ -48,29 +49,32 @@ final case class GraphLens(
...
   //TODO hide away
   def sendMessage(msg: VertexMessage): Unit = messageHandler.sendMessage(msg)
 
 
-  def vertexVoted(): Unit = voteCount += 1
-  def nextStep(): Unit    = superStep += 1
+  def vertexVoted(): Unit = voteCount.incrementAndGet()
+  def nextStep(): Unit    = {
+    voteCount.set(0)
+    vertexCount.set(0)
```
wuliaososhunhun, line 75: this does not make sense. vertex count should be = static all the time

miratepuffin: If the user is calling .getMessagedVertices() this number will change each superstep so needs to be reset.

miratepuffin:
Couple of notes:

- When submitting an analyser with the wrong name still get a message saying it was submitted successfully 

- Is there any need to forward messages to the sub task worker, once they have been created could they not just be left running.
1.  On this point we need to kill subtask workers after the view has concluded
2.  Secondly, the subtask worker knows the current job ID and superstep so I believe if we messaged directly these could be removed from the messsage

- The sending of messages seems to be correct in the count of sent and received, suggesting that the storage or sending is not working correctly

miratepuffin:
Current state of focused issues: 

- Slow down of analysis over a Range of flattenings #82 requires testing. Similarly for issue #84 we need to check if the vertex messaging is indeed faster. 
- Issues #101 and #121 for the running of batched jobs serially is now fixed 
-  Issues #123 for semi permanent flattenings, issue #124 for moving the analytical state inside of visitors and #122 for reducing the number of message queues is now done 
- Issue #85 and #83 for clearing of messages and analytical state still needs to be implemented by killing analysis subtask workers

Issues touched during this PR:

- Issue #93 for run time compile of analysers has been tided but still not fully working.

- Issue #99 for a standard process for retrieving results has been implemented in the form of the new serialisers. This needs to be expanded for distributed serialisation when storing a value for all nodes (issue #97).

miratepuffin:
Analysis results all seem correct now! Just need to kill off workers and remove the strings from vertex messages and I think this PR can be finally merged :) 

For the testing to see if we have reduced the slow down, I think this may come under some of the work planned by @imanehaf  and @iamsmkr to fix up the logging and profiling. Perhaps this should become a separate element we work on as comparisons are still being made to Master. What are your thoughts @wuliaososhunhun & @alnaimi-

PR #157, opened 2021-07-29
Tidy up build file

jamesalford:


miratepuffin:
All seems great for me, just had to bring the global settings into their own function call as it was complaining about it. @imanehaf would you mind just double checking this works for you and then I will merge.

imanehaf (approved):
Tested locally -- Everything seems in order.

richardclegg:
I am on annual leave until 7th August. For most of this time I will be at sea and the sea has notoriously poor WiFi.

PR #163, opened 2021-09-15
Feature/partition clean up

miratepuffin:
- Currently have abstracted the graph partition so that we can work on the storage analysis and ingestion separately. They were very intertwined before.
- Have removed the concept of a local partition/partitioned shared state as this means we can move partitions round a lot easier and support other non-object based partitions i.e. arrow.
- Had a spring clean on all function names for ingestion and the Akka messages they pass. 
- Turned the object based entity visitors and graph lens into interfaces so that we can easily see what a user will have access to + wrapper functions. This also further separates the storage and analysis. 
- Reworked all the `worker classes` (router/writer/reader) to have a unique name and not reference the machine they are on.
- Removed all passing of the Partition Manager count around which is now a single value inside the Raphtory Actor class
- Started to break vars such as this out into a separate conf to allow users to set all configuration elements easier. 
- Removed all the old compile at runtime code as we will be adding compiled jars to the class path.
- Removed a load of old Kamon logging code which caused deadlocking issues 
- Removed the Router Manager and fully renamed the router to graph builder throughout
- Deleted a load of unneeded utils and and actors, including the original seed actor.
- Refactored the project structure to be more appropriate for parallel development
- Turned the actor lookup functions within the RaphtoryActor into lazy vals so that they do not reinitialise the arrays every time.


richardclegg:
I am on leave until 19th September. I will look at email but will respond very slowly.

miratepuffin:
- Currently have abstracted the graph partition so that we can work on the storage analysis and ingestion separately. They were very intertwined before.
- Have removed the concept of a local partition/partitioned shared state as this means we can move partitions round a lot easier and support other non-object based partitions i.e. arrow.
- Had a spring clean on all function names for ingestion and the Akka messages they pass. 
- Turned the object based entity visitors and graph lens into interfaces so that we can easily see what a user will have access to + wrapper functions. This also further separates the storage and analysis.

PR #165, opened 2021-10-05
Actually adding in the new Query Engine

miratepuffin:


Haaroon:
i have tested this with a custom spout and the lotr example and both work.

Haaroon:
TODO refractor Raphtory/Examples

miratepuffin:
Yeah that is fine, we can do that once I have finished with the new Queries Engine and Fixed the deployment classes as both with affect the examples.

PR #169, opened 2021-10-14
Query expansion and fixes

miratepuffin:
**Direct Query**
- Removed the need for Row(Array()) is now just Row()
- Added getStateOrElse to vertex
- Shorted then job names to remove the full path which was a bit ugly and unneeded
- Allow jobs such as the GraphState check to go strait into table form without any graph operations
- Added global vertex count availability to algorithms
- Added Explode Table operation to turn one row into many
- Added Explode Edges and helpers to view Temporal Edges as singular entities (one for each update)
- Changed property access to take a class tag instead of having .asInstanceOf[T] everywhere
 
**Query Management**
- Fixed the Raphtory Client connection so queries aren't dropped
- Confirmed the Query Manager and Analysis Manager return the same results
- Purged the Analysis Manager

**Misc Spout**
- Removed the environment variables from the  file spout, to do the same from all other areas
- Added a generic parquet spout

### DIFF
--- a/src/main/scala/com/raphtory/spouts/ParquetSpout.scala
+++ b/src/main/scala/com/raphtory/spouts/ParquetSpout.scala
@@ -1,6 +1,6 @@
 package com.raphtory.spouts
 
-import com.github.mjakubowski84.parquet4s.ParquetReader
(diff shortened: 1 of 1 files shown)


Haaroon:
parquet spout was based on a previous version i had written and previously tested, explode function has been tested and works as has getorelse,

PR #170, opened 2021-10-16
Feature/triangle count

narnolddd:
Converted Triangle Count algorithm to new API. This works with self-loops and duplicate edges (i.e. pretends they don't exist)
Ben added getNeighbours functionality -- this is distinct from getEdges.


miratepuffin:
Also added here message self function for @Haaroon

PR #171, opened 2021-10-16
Feature/degree

narnolddd:
Added functionality for:
Degree, which returns in/out/total degree of each vertex
Average neighbour degree, which returns for each vertex the average degree of its neighbours.


miratepuffin:
Also fixes bug with message Queues which could not be seen if going straight from step into select

PR #176, opened 2021-10-25
ported out degree

Haaroon:
ported out degree code into new algorithm format, fixed as performing count in step


src/main/scala/com/raphtory/algorithms/newer/OutDegreeAverage.scala
```diff
@@ -0,0 +1,26 @@
+package com.raphtory.algorithms.newer
+
+import com.raphtory.core.model.algorithm.{GraphAlgorithm, GraphPerspective, Row}
+
+class OutDegreeAverage(output:String) extends GraphAlgorithm {
+  override def algorithm(graph: GraphPerspective): Unit = {
+    var nodeCount = 0L
+    graph
+      .step({v => nodeCount += 1})
```
miratepuffin, line 9: You can go back to using the node count function as this should now be fixed and the above won't work

miratepuffin (changes requested):
I am not sure if this algorithm makes sense without an aggregate step. This is more along the lines of the proportion of edges that a node has instead of an overall average.

Haaroon:
it doesnt make sense, but we can commit it for now and aggregate when thats ready

PR #177, opened 2021-10-26
wattscascade rewritten

Haaroon:
Rewrote the old Watts Cascade algorithm into the new algorithm format. 
Added everything as command-line arguments to simplify the algorithm.


src/main/scala/com/raphtory/algorithms/newer/WattsCascade.scala
```diff
@@ -0,0 +1,53 @@
...
+            if (degree > 0 && !vertex.getStateOrElse[Boolean]("infected", false))
+              (vertex.messageQueue[Double].sum/degree > threshold_choice)
+            else
+              vertex.getState[Boolean]("infected")
+          if (newLabel != vertex.getState[Boolean]("infected")) {
+            vertex.messageAllNeighbours(1.0)
+            vertex.setState("infected",true)
+          } else
+            vertex.voteToHalt()
+      }, 100)
```
miratepuffin, line 40: will need to add in the new `messagedOnly` argument as we will probably merge that first.

miratepuffin (commented):
All looks good other than adding `messagedOnly` to your iterate step

PR #178, opened 2021-10-26
Feature/degree distribution

Haaroon:
Degree Distribution code re-written for new algorithm format

This calculates the following for a graph 
total number of vertices, 
maximum degree of any node, 
average Squared degree  
average Squared in degree  
average Squared out degree


miratepuffin (changes requested):
I think for this we need to add some aggregation component to the graph algorithm. Trying to aggregate in this manner won't work unfortunately.

Haaroon:
> I think for this we need to add some aggregation component to the graph algorithm. Trying to aggregate in this manner won't work unfortunately.

unfortunately u are correct

PR #179, opened 2021-10-26
LPA implementation without aggregation

Haaroon:
LPA implementation from existing code converted to new algorithm format but without aggregation


miratepuffin (commented):
This all looks good to me, would be great if @imanehaf can confirm happy with it as well.

imanehaf:
This is ready to go, functionally speaking. The only hiccup for me is you're spitting out a label per vertex -- You might want to wait til you aggregate results before pushing this up.

Haaroon:
thanks, aggie gna take too long, will do that once its pushed

PR #180, opened 2021-10-28
Feature/two hop neighbour

Haaroon:
added algorithm for two hop neighbour, given a node or the entire graph, find and return all two hops neighbours as well as the path.


src/main/scala/com/raphtory/algorithms/newer/twoHopNeighbour.scala
```diff
@@ -0,0 +1,81 @@
+package com.raphtory.algorithms.newer
+
+import com.raphtory.core.model.algorithm.{GraphAlgorithm, GraphPerspective, Row}
+
+import scala.collection.mutable
+
+/*
+  Description: Two hop neighbours.
+  given a start node, or not given the start_node, this will
```
miratepuffin, line 9: This is a little confusing

src/main/scala/com/raphtory/algorithms/newer/twoHopNeighbour.scala
```diff
@@ -0,0 +1,81 @@
...
+  override def algorithm(graph: GraphPerspective): Unit = {
+    graph
+      .step(
+        vertex =>
+          if (nodeID == -1){
+            vertex.getEdges().foreach(edge => edge.send(("twoHopRequest", vertex.ID, 0)))
+          }
+          else if (vertex.ID() == nodeID){
+            vertex.getEdges().foreach(edge => edge.send(("twoHopRequest", vertex.ID, 0)))
+          }
```
miratepuffin, line 34: Could merge these into an or

src/main/scala/com/raphtory/algorithms/newer/twoHopNeighbour.scala
```diff
@@ -0,0 +1,81 @@
...
+            }
+          }
+          if (responses.nonEmpty) {
+            vertex.setState("twoHopResponse", true)
+            var twoHops = vertex.getOrSetState("twoHops", mutable.ListBuffer[(Long, Long)]())
+            responses.foreach(response => twoHops.append((response._2, response._3)))
+            twoHops = twoHops.distinct
+            vertex.setState("twoHops", twoHops)
+          }
+        }, 100
```
miratepuffin, line 56: If this is 2 hop, do we need so many default steps (100)

src/main/scala/com/raphtory/core/implementations/objectgraph/algorithm/ObjectGraphPerspective.scala
```diff
@@ -26,7 +26,7 @@ class ObjectGraphPerspective(vertices:Int)  extends GraphPerspective{
 
   override def step(f: Vertex => Unit): GraphPerspective = {
 
-    graphOpps.enqueue(Step(FuncSerialiser.converter(f)))
+    graphOpps.enqueue(Step(f))
```
miratepuffin, line 29: Cheers, didn't realise I left this in!

miratepuffin (changes requested):
Couple of changes noted, but all quite minor.

src/main/scala/com/raphtory/algorithms/newer/twoHopNeighbour.scala
Haaroon: yes

PR #182, opened 2021-10-29
remove degree ranking

Haaroon:
totally useless


miratepuffin (approved):
Less old algorithms the better

PR #184, opened 2021-10-29
Minor query features and ingestion batching - Correct Merge

miratepuffin:
This PR adds:

Batching:
- Batching of messages between the spout and the Builder.
- The size of these batches can be set in the conf or via a RAPHTORY_BATCH env variable. Raphtory will handle when there is not enough data to fill this batch. 
- This was added to reduce the amount of garbage being sent between the spout and the builder which was causing a large amount of GC overhead.

Query Improvements:

- Added an equivalent function to `assignID` inside of the graphAlgorithm called `checkID` this is a helper function which allows the user to feed queries with the strings that exist in the raw data and convert these to the internal long ID.
- Fixed the `graph.nodeCount` function which now returns the correct size of the whole graph.
- Added a `executeMessagedOnly` flag to iterate functions which allows the user to specify if they only want vertices which have received a message in the last superstep to run of if all vertices should always execute.

Misc:

- Removed all the old docker conf as this is now redundant and parametrised the last of the deployment arguments.


Haaroon:
- tested ingestion and graphbuilding. its deffo faster than before. watermark responds much earlier with a new timestamp, cant test analysis as they are sadly killed before they can be run. 
 done go for it

PR #185, opened 2021-11-01
Offer alternative off-heap vertex storage

miratepuffin:
Testing out using chronicle map (https://github.com/OpenHFT/Chronicle-Map) as an alternative to a full in memory graph object. This should solve some of the issues we are having with datasets much larger than memory.


miratepuffin:
This branch has unfortunately not solved a huge amount. ChronicleGraph appears to require you to only have one map and reput (re-serialise) complex data types i.e. vertices in our case. Made a couple of package refactors and other misc updates so merging those in.

Issue #153, opened 2021-05-25
Need new method for seedNode availability in docker


miratepuffin:
`nc -l seedNode -p 1600 -w 1` No longer works with Akk Artery, need some new method to check when the seednode has come online.

miratepuffin:
Fixed with 0.4.0 Release

Issue #142, opened 2021-03-18
Convert Router Worker to FSM actor


miratepuffin:
This should been a state in the Worker. Have a look at FSM Actors. They allow an Actor to respond to messages differently depending on their internal state.

_Originally posted by @Alnaimi- in https://github.com/Raphtory/Raphtory/pull/136#discussion_r586268387_

miratepuffin:
Fixed with 0.4.0 Release

Issue #141, opened 2021-03-18
Externalise deployment variables


miratepuffin:
`dockerRepository := Some("miratepuffin")` -> Docker repository needs to default to a Raphtory Repo, but should be customisable external to the build.
`val seedLoc = s"127.0.0.1:$port" ` & `    implicit val system: ActorSystem = initialiseActorSystem(seeds = List("127.0.0.1:1600"))
`-> Need to externalise the seed location within the Raphtory component

miratepuffin:
Fixed with 0.4.0 Release

Issue #117, opened 2020-12-14
Improve usability of pulling latest jar from Raphtory issues


Haaroon:
Currently, in order to download the platform one must specify the exact version number in the both the path of the url and path of the jar file

`https://github.com/Raphtory/Raphtory/releases/download/0.1/raphtory-v0.1.jar` 

Rather than this, it would be more usable to remove the version number from the jar file (as seen below), this way a user can be guaranteed to always pull the latest jar without worrying about version numbers, for the version they can simply change the `latest` value

`https://github.com/Raphtory/Raphtory/releases/latest/download/raphtory.jar`

Haaroon:
need to add this as well https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/linking-to-releases

miratepuffin:
Fixed with 0.4.0 Release

PR #194, opened 2021-11-15
0.4.0 Release

miratepuffin:
0.4.0 is a drastic overhaul over almost every component within Raphtory. This includes: 

1. A new analysis engine with a clearer more intuitive API; 
2. A full rework of Raphtory orchestration, allowing for clean local and distributed deployment; 
3. Batching and backpressure added between all components, improving stability and providing 10-100x speed up in ingestion and analysis conversion; 
4. Integration tests for all update types, different analysis tasks and windowed perspectives; 
5. CI/CD built on top of the tests for automated testing of branches, publishing of nightly builds and release tagging.

### Analysis Control Overhaul 
- Added the `Query Manager` and `Query Handler` to deal with the organisation of a query and replace the Analysis Manager/Task Handlers.
- Added `PerspectiveControllers` which look after all `perspectives` (combinations of timestamp + windows) within a job.
- Added the `QueryExecutor` inside of each Partition to run submitted algorithms -  replacing the `AnalyserExecutor`.
- Centralised `safetime` checking within the watermarker instead of asking all the Executors independently.
- Removed all REST api and replaced with a programatic `RaphtoryClient` which can be run via a compiled Jar or via the scala REPL.

### New Graph Algorithm Structure
- Added `Graph Algorithm`, `Graph Perspective` and `Table` traits as the components of our new algorithm API.
- Graph algorithms consist of calling `Step` and `Iterate` on the graph perspective - these take a function to run on each vertex. This massively expands the analytical facilities of Raphtory allowing the chaining of multiple algorithmic steps. 
- Once an algorithm has been defined via step/iterate the user may call `select` which converts each vertex to a row and returns a table abstraction. This table may be `filtered`, `exploded` (turning one row into N rows as defined by a user given function) and then written to disk via `writeTo`. 
- Global aggregation is currently removed as it was causing several issues in the previous analytical model. Elements such as counting, groupBy, topK, etc. will be added in the next minor version.

### New Analysis Features
- Added Explode Edges and helpers to view Temporal Edges as singular entities (one for each update).
- Added a global vertex count within Graph Algorithms via `graph.nodeCount` which changes based upon the perspective.
- Added an equivalent function to assignID inside of the graphAlgorithm called checkID this is a helper function which allows the user to feed queries with the strings that exist in the raw data and convert these to the internal long ID.
- Changed property access to take a class tag (i.e. `getState[String]()`) instead of requiring `.asInstanceOf[T]`
- Added `getStateOrElse` to vertex visitor.

### Deployment Overhaul
- Moved away from docker fully, allowing Raphtory to be compiled into a Jar and deployed on bare metal.
- Organised deployment classes into `RaphtoryGraph` for local deployment and `RaphtoryService` for distributed deployments.
- Added a `RaphtoryClient` for users to submit jobs to either Graph or Service.
- Converted the Raphtory Component (top level akka handler of spout, builder, partition, etc) into a Component Factory Object used by all deployment classes i.e. the `RaphtoryGraph` and `RaphtoryService`
- Moved creation of Spout and Builder away from Scala reflection to allow them to have multiple parameters. 

### Message Batching
- Batched the messages between the spout and the builders to minimise the amount of akka messages sent between them. The size of this is configurable by `RAPHTORY_BUILDER_BATCH_SIZE`.
- Added a outgoing queue map for the builder (one queue for each partition) which will be pulled when the partition is ready for more data.
- Added `RAPHTORY_BUILDER_MAX_CACHE` for the total amount of messages a builder will hold before it stops pulling data from the spout (to stop it becoming memory overloaded itself). 
- Added `RAPHTORY_PARTITION_MIN_QUEUE` which is number of messages in the queue of each partition actor below which the Partition Manager will request more data from the builders.
- Batched effect sync messages between partitions which are pushed out every second.
- Batched vertex messages between Readers during analysis, which are flushed after each superstep has concluded on all vertices in the perspective.
- Changed ChannelID between the builder/partition to be Ints instead of strings and only send 1 per batch.

### Message Handling
- Swapped to Twitter Chill library for kryo serialisation instead of altoo-ag. This removed the need to specify each case class manually in the conf and allows users to send their own types.
- Moved all actors onto the large message queue - allowing heartbeats etc to use the normal queue unopposed -  shrinking the queue size to 100k, but increasing each message frame to 10MB.
- Bumped up heartbeat monitoring to stop akka complaining when actors are looping through a batch of messages.
- Added a new custom mailbox which tracks the amount of messages in each actors queue to get a better idea of workload. This is how the Partition Manager knows how much data each partition has to process.
- Swapped the vertex message queue to use ArrayBuffer instead of ListBuffer
- Swapped the watermarker to ask the watchdog when the cluster is ready before probing for timestamps to stop fake dead letters.

### Testing:
- Added a new version of the all commands test, which is more extensible than the prior and works with the new API. This also doesn't require the `golden standard` data to be available, instead comparing hashes. The data has also been removed from the repo and is now pulled into /tmp within a users first run.
- Added a Raphtory pseudo distributed deployment class (`RaphtoryPD`) which simulates a real distributed environment for testing.
- Adding speed logging to All commands test for ingestion and queries
- Added speed logging to the Query Manager for all jobs.

### CI/CD
- Added CI pipeline workflow via GitHub actions which will - Run the all commands test on push to any branch that isn’t master - Run a nightly build of develop branch and publish a tag and release on successful build - Run a build on push to master branch, bump semantic version, create a tag and release on successful build
- Added in badges to readme to show latest build run status for push and scheduled events
- Added in badges to readme to show latest tag and release versions (SemVer only as to only show published release rather than nightly builds)

### Partition Overhaul
- Have abstracted the graph partition so that we can work on the storage analysis and ingestion separately. They were very intertwined before.
- Have removed the concept of a local partition/partitioned shared state as this means we can move partitions round a lot easier and support other non-object based partitions i.e. arrow.
- Turned the object based entity visitors and graph lens into interfaces so that we can easily see what a user will have access to + wrapper functions. This also further separates the storage and analysis. 
- Current version of the entity storage implements these and has been renamed as `POJOGraph` to make it distinct from later implementation in frameworks such as arrow.
- Reworked all the `worker classes` (router/writer/reader) to have a unique name and not reference the machine they are on.
- Turned the actor lookup functions within the RaphtoryActor into lazy vals so that they do not reinitialise every call. 
- Removed all ParTrieMaps (our last parallel data structure) as these were causing resource contention.

### Partition Memory Management:
- Swapped all multiple.treemaps to ArrayBuffers and moved to using arrays in analysis were possible as these are apparently more efficient based on recent benchmarks (https://www.lihaoyi.com/post/BenchmarkingScalaCollections.html).
- Added a state deduplication step run periodically in the partitions to remove any state which is stored more than once (say in the instance of a vertex added at the same time as some of its edges). 
- Removed the HistoricOrdering object as it is no longer used without the trees and one was being instantiated for each entity in the graph taking huge amounts of memory. 

### Spout
- Deleted the multiline file spout as not used.
- Added a mongoDB spout.
- Added a parquet spout.
- Fixed the Kafka spout to work with the overhaul from 0.3.0

### Config
- Added `RAPHTORY_LEADER_ADDRESS`, `RAPHTORY_LEADER_PORT`, `RAPHTORY_BIND_ADDRESS`, `RAPHTORY_BIND_PORT` for specifying where a Raphtory service should be binding to and where to look for the leader of a deployment.
- Added `RAPHTORY_DATA_HAS_DELETIONS` flag to only run the extra steps for handling deletions which such elements exist in the data.

### Misc clean up
- Removed all env variables throughout the code, notably in spouts and algorithms - these are all now taken as class arguments. 
- Tided up the Raphtory components and placed previously duplicated akka code such as the mediator into the top level Raphtory Actor.
- Added a `Windows` type as a wrapper to List[Long] to better explain what is happening when you submit a query
- Shortened Raphtory job names, removing the full algorithm path. 
- Deleted env-setter as no longer using raphtory docker image.
- Deleted old docker settings inside of `build.sbt` as we no longer build straight into docker.
- Removed all the old compile at runtime code as now depreciated.
- Removed old Kamon logging code which caused deadlocking issues. 
- Removed the Router Manager and fully renamed the router to graph builder throughout.
- Deleted  unneeded utils and and actors, including the original seed actor.
- Deleted the old Snapshot Manager
- Increased the frequency of watermarking 
- Added logging info of the IP of each Partition joining a Raphtory cluster to better discern where an issue arrises from.


miratepuffin:
Fixes #153 , Fixes #153 , Fixes #142 , Fixes #141 , Fixes #117 , Fixes #100 , Fixes #98 , Fixes #94 , Fixes #91 , Fixes #89 , Fixes #84 , Fixes #82 , Fixes #77

Haaroon:
i have tested this build by ingesting the ethereum blockchain multiple times. since this has greatly improved the speed of ingestion, building and analysis.

Haaroon:
please change log back to info from warn, as warn produces really nothing

Haaroon:
if no other comments tis is ready

imanehaf (approved):
This looks good to go!

Haaroon (approved):

Haaroon (approved):

PR #203, opened 2021-11-16
Feature/weighted page rank

narnolddd:
Added the weighted page rank algorithm which either uses a "weight" tag or the size of the edge history as a weight. It's producing "sensible" looking output but if anyone is aware of other ways of sanity checking this, let me know! :)


Haaroon:
rejected until you add docs  under format

* Description
* Parameters
* Returns

PR #205, opened 2021-11-16
Adding generic taint algorithm

Haaroon:
adding generic taint algorithm


src/main/scala/com/raphtory/algorithms/GenericTaint.scala
```diff
@@ -0,0 +1,138 @@
...
+Parameters
+  startTime Long : Time to start spreading taint
+  infectedNodes Set[Long] : List of nodes that will start as tainted
+  stopNodes Set[Long] : If set, any nodes that will not propogate taint
+  output String : Path where the output will be written
+
+Returns
+  Infected ID (Long) : ID of the infected vertex
+  Edge ID (Long) : Edge that propogated the infection
+  Time (Long) : Time of infectino
```
miratepuffin, line 21: Gone Italian

PR #206, opened 2021-11-16
0.4.0 Readme

miratepuffin:
New version of the readme for 0.4.0. I need to create the page for distributed deployment so that is currently the same as the one above. Please have a read and make any suggestions.


narnolddd:
Looks great, I like the short TLDR version! Just a few things: 
- "Installation" and "Write your own analysis is out of date" on the github.io page is out of date for this version, the latter could be replaced with something similar to the blog post I did? 
- Distributed deployment points to "write your own analysis"
- "You will then see Raphtory build the graph and execute an algorithm relevant to your chosen dataset." I'm a bit confused, is this an out of the box example that it uses like lotr? 
- Are we telling users to use the sbt compile + run workflow or running the main deployment class workflow?

narnolddd:
Ahhhh ignore me! Didn't realise the .io page hadn't been merged, makes total sense now :)

jamesalford (approved):
Looks good

PR #208, opened 2021-11-23
Feature/distinctiveness centrality

narnolddd:
Implemented Distinctiveness Centrality and checked that this produces correct results.

I have been using doubles here since just a single messaging step occurs and the mathematical functions like math.pow and math.log10 use doubles, is it cool to send doubles over the wire? (seem to remember that floats are preferred but this is not messaging heavy as an algo).

Finally, as part of this I added a getEdge() fn in the PojoExVertex class, it seems fine but I am nervous of editing the I N T E R N A L S

### DIFF
--- a/src/main/scala/com/raphtory/core/implementations/pojograph/entities/external/PojoExVertex.scala
+++ b/src/main/scala/com/raphtory/core/implementations/pojograph/entities/external/PojoExVertex.scala
@@ -45,6 +45,8 @@ class PojoExVertex(private val v: PojoVertex,
   def getOutEdge(id: Long,after:Long=0L,before:Long=Long.MaxValue): Option[Edge] = individualEdge(internalOutgoingEdges,after,before,id)
   //In edges individual
   def getInEdge(id: Long,after:Long=0L,before:Long=Long.MaxValue): Option[Edge]  = individualEdge(internalIncomingEdges,after,before,id)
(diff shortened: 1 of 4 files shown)


src/main/scala/com/raphtory/core/implementations/pojograph/entities/external/PojoExVertex.scala
```diff
@@ -45,6 +45,8 @@ class PojoExVertex(private val v: PojoVertex,
   def getOutEdge(id: Long,after:Long=0L,before:Long=Long.MaxValue): Option[Edge] = individualEdge(internalOutgoingEdges,after,before,id)
   //In edges individual
   def getInEdge(id: Long,after:Long=0L,before:Long=Long.MaxValue): Option[Edge]  = individualEdge(internalIncomingEdges,after,before,id)
+  // edge individual
+  def getEdge(id: Long,after:Long=0L,before:Long=Long.MaxValue): Option[Edge]  = individualEdge(internalIncomingEdges++internalOutgoingEdges,after,before,id)
```
miratepuffin, line 49: The only issue I could see here is that if there is an incoming and outgoing edge we may return different objects each time. If the user is using History etc this could be a problem. But as long as we give reasonable warning I think this is fine!

miratepuffin (approved):

PR #209, opened 2021-11-23
Wordsem

narnolddd:
I think `++` wasn't concatenating `neilabs` in-place so replaced it with `neilabs = neilabs ++ ...`, @imanehaf could you possibly ok this? Thanks!


miratepuffin:
Ello, I believe the issue you have here is that 
image
this function returns an immutable map. If you want to fix this you can create a mutable map within the function and return that.  Then you can go back to your previous appending of data. Also this is merging into master, would you mind recreating and merging into Develop.

PR #210, opened 2021-11-25
Possible fix to MultilayerLPA

narnolddd:
Pushing to the correct place this time.
(neilabs wasn't getting appended with labels from self from other layers I don't think)


narnolddd:
Have changed neilabs to an appendable data structure rather than needing to make multiple lists and concatenate :)

miratepuffin (approved):

PR #214, opened 2021-12-14
New community detection algorithm

narnolddd:
Added SLPA algorithm (more relaxed version of LPA) based on https://arxiv.org/pdf/1109.5720.pdf for overlapping communities.


miratepuffin (approved):
The algorithm looks very interesting! Love the rule case classes.

PR #215, opened 2021-12-16
New HDFS Spout

calzadaharo:
This is the first version of an HDFS Spout. It can ingest both individual or partitioned files.

### DIFF
--- a/src/main/scala/com/raphtory/spouts/HdfsSpout.scala
+++ b/src/main/scala/com/raphtory/spouts/HdfsSpout.scala
@@ -5,6 +5,8 @@ import com.raphtory.core.components.spout.Spout
 import java.io.{BufferedReader, InputStreamReader}
 import java.net.URI
 
+import com.typesafe.scalalogging.LazyLogging
+
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.{FileSystem, Path}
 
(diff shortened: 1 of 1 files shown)


src/main/scala/com/raphtory/spouts/HdfsSpout.scala
```diff
@@ -0,0 +1,55 @@
...
+
+    def nextLine(): String = {
+      bufferReader.readLine()
+    }
+
+    def finished(line: String): Boolean = {
+      if (line == null) {
+        if(status.length > 0) {
+          hdfsManager = HDFSManager()
+          println("Still loading Spout...")
```
miratepuffin, line 44: Hi Rodrigo, thanks for adding this in! Would you mind changing these prints to use loggers and I shall merge it in :)

calzadaharo: Done! I hope it is what you were hoping for.

miratepuffin: Fab! Will merge it in now 🎉

miratepuffin (approved):

src/main/scala/com/raphtory/spouts/HdfsSpout.scala
calzadaharo: Nice! :)

PR #216, opened 2021-12-21
Added weighted degree algorithm.

narnolddd:
Weighted degree algorithm which returns the sum of in-edge, out-edge and total edge weights per node. If edges don't have a numerical weight property it just returns the number of in/out edges and the total of these.


miratepuffin (approved):
Great stuff Naomi! We need to swap this to the new API, but that can be done within @ljeub-pometry 's refactor

PR #217, opened 2022-01-12
Make Algorithm API composable

ljeub-pometry:
I've started working on making the Algorithm API composable: 

The main idea is to break up the algorithm into three stages:

1. graphStage: takes a graph as input and returns a graph (can be composed)
2. tableStage: takes a graph and runs select to return a table (not composable)
3. write: takes table as input and writes out results (not composable)

This also implements an AlgorithmChain which takes a sequence of algorithms as input and runs the graph stages in sequence and then runs the table and write out of the final algorithm in the chain.

Suggestions for method names and comments very welcome!

### DIFF
--- a/src/main/scala/com/raphtory/algorithms/Ancestors.scala
+++ b/src/main/scala/com/raphtory/algorithms/Ancestors.scala
@@ -9,7 +9,7 @@ class Ancestors(path:String,
                 directed:Boolean=true)
   extends GraphAlgorithm{
 
-  override def algorithm(graph: GraphPerspective): GraphPerspective = {
+  override def apply(graph: GraphPerspective): GraphPerspective = {
     graph.step({
       vertex =>
         if (vertex.ID() == checkID(seed)) {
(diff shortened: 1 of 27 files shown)


ljeub-pometry:
New suggestion for naming the three stages:

graphStage -> run
tableStage -> tabularize
write -> write

ljeub-pometry:
All existing algorithms except for CBOD should now be composable. 

CBOD needs some rethinking, my thought is that it should just take a string label as input and not run any community detection itself. Thus using it would require making use of the new composable interface.

miratepuffin (approved):
These changes make a lot of sense and are one of the things the algorithm API really needed. I have changed the function names to `algorithm`, `tabularise` and `write`. What was previously `algorithm` is now called `run`.

miratepuffin:
As a final change I have made `algorithm` into `apply` so that all functions are verbs. I have also added an `->` operator to build chains of algorithms, for example:
image

PR #218, opened 2022-01-13
Feature/optionally include properties

ljeub-pometry:
Implement a flag to include values defined as properties when looking up state on vertices. This is useful in conjunction with the  composable API (see #217) as it means that an algorithm can declare that it does not care whether a value was the result of computations done by a previous algorithm in the chain or something that was defined when the data was ingested.


miratepuffin (approved):
Improvements make sense, especially with the new composable algorithm API. I have fixed the merge conflicts and will merge into dev

PR #220, opened 2022-01-17
Feature/fix weighted degree

ljeub-pometry:
Fix the weighted degree algorithm which got broken by merging in the new algorithm api


miratepuffin (approved):
Nice Spot @ljeub-pometry

PR #222, opened 2022-01-17
Reorganise algorithm folder

ljeub-pometry:
This is an attempt at organising the algorithms folder. This first splits the algorithms into "generic" algorithms which just look at vertices and edges, ignoring the temporal properties of the data and "temporal" algorithms which actually explore the history of vertices and edges.

Within these two groups I've categorised the algorithms into

- centrality: compute node centrality values
- community: community detection algorithms
- dynamic: simulated dynamical process

There also remain a few unclassified algorithms. Any suggestions and comments welcome!


ljeub-pometry:
image

Here is what the new organisation looks like

miratepuffin (approved):

PR #225, opened 2022-01-24
"name" function for vertices

narnolddd:
This is to try and make it easier/more standardised to output a vertex's name in algorithm results. 

`vertex.name(nameProperty:String)` is the same as `vertex.getPropertyOrElse(nameProperty, vertex.ID)` so that the vertex's assigned ID (long) is returned if it doesn't have a name property. The input `nameProperty` defaults to `"name"` meaning that `vertex.name()` becomes a nice shorthand if we assume that the name property is called "name".

I don't know whether this is the correct/best way of doing this, let me know what you think! :)

### DIFF
--- a/src/main/scala/com/raphtory/core/implementations/pojograph/entities/external/PojoExVertex.scala
+++ b/src/main/scala/com/raphtory/core/implementations/pojograph/entities/external/PojoExVertex.scala
@@ -17,8 +17,6 @@ class PojoExVertex(private val v: PojoVertex,
 
   override def ID() = v.vertexId
 
-  def name(nameProperty:String="name"): String = getPropertyOrElse(nameProperty,ID.toString)
-
   private val multiQueue: VertexMultiQueue = new VertexMultiQueue() //Map of queues for all ongoing processing
   private var computationValues: Map[String, Any] = Map.empty //Partial results kept between supersteps in calculation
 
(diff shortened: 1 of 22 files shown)


src/main/scala/com/raphtory/core/implementations/pojograph/entities/external/PojoExVertex.scala
```diff
@@ -17,6 +17,8 @@ class PojoExVertex(private val v: PojoVertex,
 
   override def ID() = v.vertexId
 
+  def name(nameProperty:String="name"): String = getPropertyOrElse(nameProperty,ID.toString)
```
ljeub-pometry, line 20: You could have put this implementation directly into the Vertex trait, any reason not to?

ljeub-pometry (approved):
I like it. I've been meaning to add more of these convenience methods (especially for degree) but haven't gotten around to it yet. I would move the implementation to the Vertex trait though as it doesn't need access to any of the internal state...

src/main/scala/com/raphtory/core/implementations/pojograph/entities/external/PojoExVertex.scala
miratepuffin: I agree with @ljeub-pometry here, we can put quite a few of these directly into the vertex trait so they are available for all storage models :) 

Are you ok to make the change @narnolddd? I can then merge it in.

narnolddd: Good call, I will move it over to Vertex Trait and let you know when I've pushed it up. Thanks for checking this out! :)

narnolddd: pushed revised version now :) thanks @ljeub-pometry @miratepuffin!

ljeub-pometry (approved):

PR #226, opened 2022-01-31
Feature/multilayer explode

narnolddd:
I made a utility algorithm which turns a temporal network into an "exploded" multiplex network. Nodes are given copies of themself for the snapshots/layers in which they are present and edges are added between them are given accordingly. Finally, an edge is given between two copies of the same node if they are present in consecutive layers. The final network is written out as a csv file.


ljeub-pometry (approved):
This is probably fine to merge as is, though it may be nice to make it chaining friendly?

PR #232, opened 2022-02-22
Reenabled file polling via scheduler

miratepuffin:


miratepuffin:
Reenabled the file polling by the file spout so that new files added to the directory are sent through to the graph builder.

Issue #92, opened 2020-11-22
Global Analytical State


miratepuffin:
For many algorithms such as pagerank and Hub/authority their needs to be a small global state maintained. This should be completed by allowing the analysis manager to aggregate and report back with the next analyse superstep request.

ljeub-pometry:
Addressed by #236

PR #236, opened 2022-02-23
Feature/graph accumulator

ljeub-pometry:
_Edited by @miratepuffin _

**What changes were proposed in this pull request?**

- Working implementation of graph accumulators.

**Why are the changes needed?**

- Currently, there is no way to establish any information about the whole network.

**Does this PR introduce any user-facing change?**

- This adds several new functions to the graph analysis API but does not affect any old functions/algorithms - nor any older deployments.

**How was this patch tested?**

- This was tested locally via the full test suite and has an additional accumulator test added in to examine the new features.

**Still to do**

- Add documentation
- More convenience methods for creating common accumulators


ljeub-pometry:
- added convenience methods for min/max/sum accumulators
- documented GraphState and GraphPerspective with docstrings

- performance issue still remains (not sure what the root cause is yet)

ljeub-pometry:
Note this addresses issue #92

PR #238, opened 2022-02-24
New Query API

ricopinazo:
Add Window interface to allow managing variable intervals with temporal units


ricopinazo:
Just realized that forgot to use the new interface to express the increments for range and live queries. I'm gonna rename it to 'Interval'

ricopinazo:
The last commit adds the new Query definition, replacing the older PointQuery, RangeQuery and LiveQuery case classes. In order to handle it within the different classes of the incoming class hierarchy (not yet here), a new immutable class QueryBuilder has been created as a wrapper of a Query object. This object is able to set all the parameters of the query and submit it. The RaphtoryClient has been changed to use this class to define and submit queries instead of doing it by itself.

ricopinazo:
It also modifies GenericGraphPerspective and GenericTable so they are immutable objects that use internally a QueryBuilder to progressively define the query and eventually submit it. They are not used anymore inside the QueryHandler. The idea is that RaphtoryGraph will eventually extend GraphPerspective to add the new methods (slice, from, until) on top of them.

ricopinazo:
Another option would be that GenericGraphPerspective and GenericTable extend QueryBuilder instead of having it as a member, but the only way that comes to my mind to do this is using F-bounded polymorphism, which leads to a not very readable implementation. There should be a way to do it using a type class, not sure how.

ricopinazo:
Just committed a few more changes. After a review by some of you, I think the current state could be a good baseline to merge in.

The main changes are:

- Created two user facing traits: 
    - `RaphtoryGraph`: a collection of views over a graph the user can apply operations on (filter, step, etc). It returns a Table when select is called, and then writeTo causes the submission of the query, returning a `QueryProgresTracker` object.
    - `TemporalGraph`: a special case of `RaphtoryGraph` with only one view, whose timeline can be manipulated and projected into a RaphtoryGraph

- Renamed `RaphtoryGraph` class to `GraphDeployment` (is this ok?).
- Added two new methods to `Raphtory` object:
    - `deployGraph`: creates a `GraphDeployment` internally and returns a `TemporalGraph` attached to it. This path doesn’t provide a way to stop the components, which probably needs to be tackled later on.
    - `getGraph`: returns a `TemporalGraph`  attached to the deployment specified in the configuration.
- The `TemporalGraph` methods provide versions accepting `String` instead of `Long`  to express:
    - Timestamps: the format is set in the configuration: `raphtory.query.timeFormat = "yyyy-MM-dd HH:mm:ss[.SSS]"`
    - Intervals: the user can introduce thinks like "2 days" , "1 month" , "1 minutes and 3 seconds"  and a parser tries to translate it to ISO8601 format for intervals to further using the standard java parser for intervals.
Internally, an Interval trait is used to hold interval values, which can be discrete or time intervals. `Long` is still the type we use to store times, interpreted as milliseconds from the epoch by default, but an implicit conversion has been introduced to allow add and subtract interval objects to longs.
- New `Query`  definition introduced, replacing the three previous versions `PointQuery` , `RangeQuery` and `LiveQuery` to enable expressing all the new things the new API is going to allow using optional values. There is also a new class `QueryBuilder` that acts as a wrapper for the `Query` and handles the submission of these.

Important considerations:
- The old API and the new one coexist, using different methods as entry points from the `Raphtory` object.
- The `GraphPerspective` remains as a completely different trait from `RaphtoryGraph` . Right now it doesn’t provide more functionality than `RaphtoryGraph` does. If this doesn’t change in the future I would consider that they both merge into a single trait to avoid redundancy.
- Although GraphAlgorithm still exists, it’s important to notice that the new API allows the user to define algorithms as functions `RaphtoryGraph => RaphtoryGraph` or even `RaphtoryGraph => Table` which is probably even more interesting if we target python users.
`nodeCount` method has been removed from `GraphPerspective` expecting that this is addressed using accumulators shortly. `Distinctiveness` algorithm has stopped working as a result, which needs to be solved.

I am aware of the discussion around some names and semantics. I think that is good to have this PR as a baseline to iteratively keep improving it with further feedback and when the functionality we want to provide becomes more clear.

ricopinazo:
I have added a second test using the new API inside testAlgorithm method of BaseAlgorithmTest. Everything seems to work fine.

Besides, I have reviewed the timestamps of the logs to spot any regression in execution time since the last merge request into 0.5.0 branch. From the time the query progress tracker is created to the time the hash is generated for the results the elapsed time is still 3004+-1 milliseconds for the degree test.

ricopinazo:
I have added the GraphOperations trait that describes all graph operations (filter, step, iterate, etc) and that returns a generic class extending GraphOperations for every operation so it is easier to reuse the code between RaphtoryGraph and GraphPerspective. I think it is worth it to include it in this PR because the code is way cleaner. It could be useful in the future so a TemporalGraph object is able to return a TemporalGraph after calling, e.g., filter (right now is returning a RaphtoryGraph because currently time handling takes places before the execution of graph operations internally)

ricopinazo:
With the last changes, the starting point to define the queries is a `TemporalGraph`. Now, the definition of the perspectives is carried out in three different steps, where the user can:
1. Filter the timeline, obtaining a new `TemporalGraph`.
2. Define a set of points over the timeline, obtaining a `DottedGraph`.
3. Decide where to look from each point (to the past, to the future, or within a window aligned with the point in an arbitrary way), obtaining a `RaphtoryGraph`, which is effectively a collection of perspectives over the graph.

From the `TemporalGraph`, the users have access to the graph operations the same way they have from the `RaphtoryGraph`. In this case, a perspective is set from the start of the timeline to the end of it. In contrast, from a `DottedGraph` the users don't have access to the graph operations, i.e., they cannot write something like `graph.range(...).execute(algo)`. Instead, they need to specify some window method from the `DottedGraph` class. This allows us to decide the default windowing strategy later on.

PR #241, opened 2022-02-25
Feature/weighted random walk

ljeub-pometry:
Implement random walk that takes edge weights into account


ljeub-pometry:
This should be functionally working though can be made more efficient in terms of sampling

PR #252, opened 2022-03-03
Feature/edge weight

miratepuffin:
Bringing over from #227 -- @narnolddd's old message:

I've added some functionality here regarding the edge weight after noting myself repeating the same sort of code over and over when writing algorithms, namely edge.getPropertyOrElse("weight",edge.history().size). Will list them in bullet points.

The main one in the edge trait is weightOrHistory in the Edge trait. This looks for a property called "weight" (a user can supply a different string for this if desired and adds up the values of this in the history. E.g. in a transaction type setting it would be the total amount of money transferred over the edge during the time period. If there is no weight property, it returns the number of occurrences of that edge in the view.
Second is various concepts of node weight which is some functions of the edge weights coming in and out. These node weight functions involve a merge strategy which defines how you want to combine the edge weights. The default and I think most relevant one would be sum (adding the edge weights) but I've included product, min, max.
In a separate function I've included "difference" as a strategy for getting a node's incoming - outgoing weight (equivalent of "balance" in a financial network probably).
Quite a few classes have been touched by this so I'd appreciate any guidance about whether this is the best way of doing things. Also any naming suggestions are welcome, not sure I'm keen on weightOrHistory 😅 Thanks!


miratepuffin:
Seems that the failing tests have an empty dir, which suggests we are swallowing an error somewhere where we 100% should not be doing so.

@ljeub-pometry If you wouldn't mind having a double check of this before we merge it in (once we fix this issue).

ljeub-pometry:
Hi, I've done a major refactor of this feature. There are two key changes:

- Moved the MergeStrategy functionality to the property level 
  * this means that one can now control aggregation when accessing properties for both vertices and edges
  * default behaviour is unchanged (i.e., return latest value)
  * different strategies are implemented an can be dropped in
- Edge weight functionality is build on top of this MergeStrategy implementation and can use generic numeric types
  * unlike property access, MergeStrategy defaults to computing the sum of values

Still to do:

- Document everything (this will be part of separate pull request for adding docstrings to all of the graph api)

src/test/scala/com/raphtory/algorithms/WeightedGraphBuilder.scala
```diff
@@ -0,0 +1,30 @@
...
+class WeightedGraphBuilder extends GraphBuilder[String] {
+
+  override def parseTuple(tuple: String): Unit = {
+    val line       = tuple.split(",")
+    val sourceNode = line(0)
+    val srcID      = sourceNode.toLong
+    val targetNode = line(1)
+    val tarID      = targetNode.toLong
+    val timeStamp  = line(2).toLong
+    val weight     = line(3).toLong
```
narnolddd, line 17: Good call adding a weighted graph builder! Just a thought, I remember seeing something in a networks library (possibly graph-tool?) for reading in weighted networks from csv where the user could specify which columns in their csv correspond to `source`, `target`, `weight` defaulting to 0, 1, 2, respectively, in case the csv is not already in this format. Might be an idea to do something similar here, possibly also a separator defaulting to `,` to allow for tab separated files?

narnolddd, line 17: Good call adding a weighted graph builder! Just a thought, I remember seeing something in a networks library (possibly graph-tool?) for reading in weighted networks from csv where the user could specify which columns in their csv correspond to `source`, `target`, `weight` defaulting to 0, 1, 2, respectively, in case the csv is not already in this format. Might be an idea to do something similar here, possibly also a separator defaulting to `,` to allow for tab separated files?

src/main/scala/com/raphtory/core/graph/visitor/Vertex.scala
```diff
@@ -1,11 +1,17 @@
 package com.raphtory.core.graph.visitor
 
+import PropertyMergeStrategy.PropertyMerge
+import VertexMergeStrategy.VertexMerge
+import com.raphtory.core.graph.visitor.EdgeDirection.Direction
+
 import scala.reflect.ClassTag
 
 trait Vertex extends EntityVisitor {
```
narnolddd, line 9: Thank you for taking up this beast of a function overloading implentation 🥲 does the trick and I don't think there is a better way of doing it really, @miratepuffin ?

src/main/scala/com/raphtory/core/graph/visitor/EntityVisitor.scala
```diff
@@ -11,15 +11,43 @@ abstract class EntityVisitor {
...
+
+  def getPropertyOrElse[A, B](key: String, otherwise: B, mergeStrategy: PropertyMerge[A, B]): B =
+    getProperty[A, B](key, mergeStrategy) match {
+      case Some(value) => value
+      case None        => otherwise
+    }
+
+  def getPropertyOrElse[A](key: String, otherwise: A): A =
+    getPropertyOrElse(key, otherwise, PropertyMergeStrategy.latest[A])
+
```
narnolddd, line 36: I like this a LOT! 🙌

narnolddd (approved):

src/test/scala/com/raphtory/algorithms/WeightedGraphBuilder.scala
ljeub-pometry: Making a generic EdgeListGraphBuilder which would support these things has been in the back of my mind for a while but haven't had a chance to work on it yet. This was just a quick one for testing everything works.

PR #253, opened 2022-03-04
add a test to check that vote to halt is working

ljeub-pometry:
Small test for checking vote-to-halt functionality


src/test/scala/com/raphtory/algorithms/TestIteration.scala
```diff
@@ -0,0 +1,51 @@
...
+import com.raphtory.core.algorithm.GraphPerspective
+import com.raphtory.core.algorithm.Row
+import com.raphtory.core.algorithm.Table
+import com.raphtory.core.graph.visitor.Vertex
+
+class CountIterations(num_iters_before_vote: Int, num_iters: Int) extends GraphAlgorithm {
+
+  override def apply(graph: GraphPerspective): GraphPerspective =
+    graph
+      .setGlobalState { graphState =>
```
miratepuffin, line 13: I like the casual dropping in of the global state as well!

miratepuffin (approved):

PR #256, opened 2022-03-04
clean up Raphtory.createGraph API

ljeub-pometry:
- Only one generic versions of createGraph
- Remove unused TypeTags and Schemas


miratepuffin:
All looks good to me! Just rebased on top of the new changes for the filespout and analysis

PR #258, opened 2022-03-07
Moved prisoners dilemma to 0.5.0 version

narnolddd:
I have moved across the prisoners dilemma code from my old 0.4.0 version branch to the new one. Still to do is:
* Change the comments at the top of the class to Sphinx docstring format
* Make a test for it 
Once I have figured out how to do these I will take this out of draft mode 😃

### DIFF
--- a/src/test/scala/com/raphtory/lotrtest/LotrTest.scala
+++ b/src/test/scala/com/raphtory/lotrtest/LotrTest.scala
@@ -148,13 +148,6 @@ class LotrTest extends BaseRaphtoryAlgoTest[String] {
     assert(true)
   }
 
-  test("Prisoners Dilemma Test") {
-    assert(
-      algorithmTest(PrisonersDilemma(proportionCoop = 0.9f, benefit = 1.05f, cost = 0.05f,seed = 42),outputFormat,1, 32674, 10000, List(500, 1000, 10000))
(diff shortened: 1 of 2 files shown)


narnolddd:
I've added docstrings to the Prisoners Dilemma algorithm and made it take a seed as a parameter (for the random initialisation of "cooperator" nodes), so there is now a test for it in LOTR test.

src/test/scala/com/raphtory/lotrtest/LotrTest.scala
```diff
@@ -146,6 +147,13 @@ class LotrTest extends BaseRaphtoryAlgoTest[String] {
     assert(true)
   }
 
+  test("Prisoners Dilemma Test") {
+    assert(
+      algorithmTest(PrisonersDilemma(proportionCoop = 0.9f, benefit = 1.05f, cost = 0.05f,seed = 42),outputFormat,1, 32674, 10000, List(500, 1000, 10000))
```
miratepuffin, line 152: Algo and doc strings look great! We are having some issues at the moment where the seeded algorithms will return different results with different numbers of partitions, but this is a wider problem and will do the trick for the time being 😄

miratepuffin (approved):

src/test/scala/com/raphtory/lotrtest/LotrTest.scala
narnolddd: Ahhh ok yeah that makes a lot of sense! 😃

PR #260, opened 2022-03-08
docs: serialisers

rutujasurve94:
- [x] `PulsarSerialiser`
- [x] `Avro`
- [x] `Marshal`

### DIFF
--- a/src/main/scala/com/raphtory/serialisers/Marshal.scala
+++ b/src/main/scala/com/raphtory/serialisers/Marshal.scala
@@ -9,19 +9,19 @@ package com.raphtory.serialisers
   *
   * ## Methods
   *
-  *    {s}`dump[A](o: A)(implicit t: ClassTag[A]): Array[Byte]`
-  *      : Serialise to ByteArray
+  *    {s}`dump[A: ClassTag](o: A): Array[Byte]`
+  *      : Serialise to byte array
   *
   *        {s}`o: A`
-  *           : serialise object `A` to `Array[Byte]`
+  *           : object to serialise
   *
-  *    {s}`load[A](buffer: Array[Byte])(implicit expected: ClassTag[A])`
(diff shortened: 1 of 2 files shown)


src/main/scala/com/raphtory/serialisers/Marshal.scala
```diff
@@ -1,5 +1,50 @@
...
+  *      : Deserialise from Array[Byte]
+  *
+  *        {s}`buffer: Array[Byte]`
+  *           : deserialise from byte buffer to object
+  *
+  *    {s}`deepCopy[A](a: A)(implicit m: reflect.ClassTag[A])`
+  *      : Deep copy
+  *
+  *        {s}`a: ClassTag[A]`
+  *           : deserialise and persist as deep copy by serialising
```
ljeub-pometry, line 28: This is wrong, `a` is the object to copy, not a ClassTag

src/main/scala/com/raphtory/serialisers/Marshal.scala
```diff
@@ -1,5 +1,50 @@
...
+  * : support serialisation (`dump`) and deserialisation (`load`) for `ByteArray` input and output streams
+  *
+  *
+  * ## Methods
+  *
+  *    {s}`dump[A](o: A)(implicit t: ClassTag[A]): Array[Byte]`
+  *      : Serialise to ByteArray
+  *
+  *        {s}`t: ClassTag[A]`
+  *           : serialise object to Array[Byte]
```
ljeub-pometry, line 16: `t` is not the object being serialised, in fact `t` should almost never be given explicitly as it is a ClassTag that normally should be passed in automatically by the compiler

ljeub-pometry (changes requested):

Haaroon (approved):

PR #274, opened 2022-03-15
Fix bug causing the ConfigHanlder to ignore user input

ricopinazo:
I have kept the resolving step at the end of `createConf()` because it is used to lazily read the value of `"raphtory.deploy.distributed"` in the constructor. It seems that there are no problems adding two more values and resolving again at `local()` method.


ljeub-pometry:
The resolving again doesn't work. It means that the file spout does not pick up the changed config properly which breaks salting, i.e., some values in the config depend on the deploymentID and if you resolve too early those get hard-coded and do not pick up the change.

PR #279, opened 2022-03-16
add timeout to github ci test and build

Haaroon:
small timeout to prevent tests wasting resources if stuck in loop


Haaroon:
test showed the timeout work, its going in

PR #288, opened 2022-03-18
More sensible producer config

ljeub-pometry:
attempt to fix blocking producers bug by setting non-zero maxPendingMessages

### DIFF
--- a/src/main/scala/com/raphtory/core/config/PulsarController.scala
+++ b/src/main/scala/com/raphtory/core/config/PulsarController.scala
@@ -116,9 +116,9 @@ class PulsarController(conf: Config) {
       .topic(topic)
       .enableBatching(true)
       .batchingMaxPublishDelay(1, TimeUnit.MILLISECONDS)
-      .batchingMaxMessages(Integer.MAX_VALUE)
+      .batchingMaxMessages(10000)
       .blockIfQueueFull(true)
-      .maxPendingMessages(0)
+      .maxPendingMessages(1000)
       .create()
 
   def deleteTopic(topic: String) =


src/main/scala/com/raphtory/core/config/PulsarController.scala
```diff
@@ -78,36 +78,60 @@ class PulsarController(conf: Config) {
...
-  ): Consumer[T] =
+      topics: Seq[String]
+  ): ConsumerBuilder[T] =
     client
       .newConsumer(schema)
       .priorityLevel(0)
       .topics(topics.toList.asJava)
       .subscriptionName(subscriptionName)
-      .subscriptionType(SubscriptionType.Shared)
+      .subscriptionType(SubscriptionType.Exclusive)
```
miratepuffin, line 91: If this is being set below we can probably just remove the line here to be clearer

src/main/scala/com/raphtory/core/config/PulsarController.scala
```diff
@@ -116,9 +140,9 @@ class PulsarController(conf: Config) {
       .topic(topic)
       .enableBatching(true)
       .batchingMaxPublishDelay(1, TimeUnit.MILLISECONDS)
-      .batchingMaxMessages(Integer.MAX_VALUE)
+      .batchingMaxMessages(10000)
```
miratepuffin, line 143: I believe this was set based on pulsar best practices.
The max Pending Messages I believe can just be set back to default though? - @Alnaimi- can you confirm.

src/main/scala/com/raphtory/core/config/PulsarController.scala
```diff
@@ -78,36 +78,60 @@ class PulsarController(conf: Config) {
...
+  ): Consumer[T] =
+    configuredConsumerBuilder(subscriptionName, schema, topics)
       .subscriptionType(SubscriptionType.Shared)
-      .subscriptionInitialPosition(SubscriptionInitialPosition.Earliest)
-      .maxTotalReceiverQueueSizeAcrossPartitions(Integer.MAX_VALUE)
-      .receiverQueueSize(200_000)
-      .poolMessages(true)
+      .subscribe()
+
+  def createExclusiveConsumer[T](
```
miratepuffin, line 128: No reason not to have it, but this isn't used anywhere at the moment is it?

src/main/scala/com/raphtory/core/config/PulsarController.scala
ljeub-pometry: The `maxPendingMessages` change is because @Haaroon actually saw a deadlock after I changed the flushing of vertex messages to make sure they are actually sent before sending the `GraphFunctionComplete` message, i.e., the flushAsync call would never complete. With this set to default value, the old behaviour of looping in CheckMessages forever came back before I set the subscription to exclusive.

src/main/scala/com/raphtory/core/config/PulsarController.scala
ljeub-pometry: Not used but felt very strange not to have it here...

src/main/scala/com/raphtory/core/config/PulsarController.scala
ljeub-pometry: removed

PR #289, opened 2022-03-18
fix batchLoading errors

ljeub-pometry:
- StaticGraphSpout was broken when batch loading
- graph.stop() raised NotImplementedError when batch loading


Haaroon:
this test did not pass :(

ljeub-pometry:
#288 did pass though. This is an unrelated bug fix, would not expect it to make any difference to the message checking problem.

PR #291, opened 2022-03-20
[docs] builder executor

rutujasurve94:
Docs for `BuilderExecutor`


src/main/scala/com/raphtory/core/components/graphbuilder/BuilderExecutor.scala
```diff
@@ -15,7 +15,27 @@ import scala.reflect.ClassTag
 import scala.reflect.runtime.universe
 import scala.reflect.runtime.universe._
 
-/** @DoNotDocument */
```
ljeub-pometry, line 18: This is not part of the public API, leave this tag in place!

ljeub-pometry (changes requested):

rutujasurve94:
Closing the PR since this is not to be documented

PR #293, opened 2022-03-21
Bug/bail out on failure

ljeub-pometry:
- Make QueryHandler skip current perspective in case of error
- Add check to make sure CheckMessages is not send in an infinite loop

### DIFF
--- a/src/test/scala/com/raphtory/generic/FailureTest.scala
+++ b/src/test/scala/com/raphtory/generic/FailureTest.scala
@@ -1,8 +1,12 @@
 package com.raphtory.generic
 
 import com.raphtory.algorithms.BaseCorrectnessTest
+import com.raphtory.algorithms.BasicGraphBuilder
 import com.raphtory.core.algorithm.GraphAlgorithm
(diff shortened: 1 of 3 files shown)


src/test/scala/com/raphtory/generic/FailureTest.scala
```diff
@@ -12,9 +16,16 @@ class FailingAlgo extends GraphAlgorithm {
...
+class FailureTest extends AnyFunSuite {
   test("test failure propagation") {
-    assert(correctnessTest(new FailingAlgo, "./blank.csv", "./blank.csv", 0))
+    val graph = Raphtory.streamGraph(graphBuilder = BasicGraphBuilder())
+    val query = graph.pointQuery(new FailingAlgo, FileOutputFormat("/tmp"), 0)
+    for (i <- 1 to 10 if !query.isJobDone())
+      Thread.sleep(1000)
+    assert(
+            query.isJobDone()
+    ) // if query failed to terminate after 10 seconds, assuming infinite loop
```
miratepuffin, line 27: Nice!

PR #296, opened 2022-03-23
Spout metrics

rutujasurve94:
Adds `counter` metrics for spout

- On processing each file, increments counter `total_processed_files` counter
- On rescheduling spout, increments counter `spout_reschedule`
- Adds a `gauge` for total lines read per file
- Adds a `counter` to track number of errored files

### DIFF
--- a/src/main/scala/com/raphtory/core/client/RaphtoryGraph.scala
+++ b/src/main/scala/com/raphtory/core/client/RaphtoryGraph.scala
@@ -10,7 +10,9 @@ import com.raphtory.core.config.ZookeeperIDManager
 import com.typesafe.config.Config
 import monix.execution.Scheduler
 import org.apache.pulsar.client.api.Schema
+import io.prometheus.client.exporter.HTTPServer
 
+import java.io.IOException
 import scala.language.postfixOps
 import scala.reflect.ClassTag
 import scala.reflect.runtime.universe._
@@ -79,6 +81,12 @@ private[core] class RaphtoryGraph[T: ClassTag: TypeTag](
 
   logger.info(s"Created Graph object with deployment ID '$deploymentID'.")
   logger.info(s"Created Graph Spout topic with name '$spoutTopic'.")
+
+  try {
+    new HTTPServer(8899)
+  } catch {
+    case e: IOException => e.printStackTrace()
+  }
 
   def stop(): Unit = {
     partitions.writers.foreach(_.stop())
--- a/src/main/scala/com/raphtory/core/client/RaphtoryGraph.scala
+++ b/src/main/scala/com/raphtory/core/client/RaphtoryGraph.scala
@@ -57,6 +57,7 @@ private[core] class RaphtoryGraph[T: ClassTag: TypeTag](
 
   private val deploymentID: String = conf.getString("raphtory.deploy.id")
   private val spoutTopic: String   = conf.getString("raphtory.spout.topic")
+  private val prometheusPort: Int = conf.getInt("raphtory.prometheus.server")
 
   private val zookeeperAddress: String = conf.getString("raphtory.zookeeper.address")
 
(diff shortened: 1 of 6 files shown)


src/main/scala/com/raphtory/core/client/RaphtoryGraph.scala
```diff
@@ -80,6 +82,12 @@ private[core] class RaphtoryGraph[T: ClassTag: TypeTag](
   logger.info(s"Created Graph object with deployment ID '$deploymentID'.")
   logger.info(s"Created Graph Spout topic with name '$spoutTopic'.")
 
+  try {
+    new HTTPServer(8899)
```
Haaroon, line 86: can we please make the port be pulled from the config? Is there also any reason why we are not storing the HTTP server object? what if we want to shut it down?

src/main/scala/com/raphtory/core/components/spout/SpoutExecutor.scala
```diff
@@ -41,6 +43,7 @@ class SpoutExecutor[T](
     executeSpout()
 
   private def executeSpout() = {
+    SpoutTelemetry.totalSpoutReschedules.inc()
```
Haaroon, line 46: the counter objects cant throw an exception right? i checked the source and i couldn't see anything, if they cant then this is fine.

src/main/scala/com/raphtory/core/config/telemetry/SpoutTelemetry.scala
```diff
@@ -0,0 +1,30 @@
+package com.raphtory.core.config.telemetry
+
+import io.prometheus.client.{Counter, Gauge}
+
```
Haaroon, line 4: Please document. What is this file. What does it do. What does it depend on.

src/main/scala/com/raphtory/core/config/telemetry/SpoutTelemetry.scala
```diff
@@ -0,0 +1,30 @@
...
+object SpoutTelemetry {
+  val totalFilesProcessed: Counter =
+    Counter
+      .build
+      .namespace("spout")
```
Haaroon, line 9: namespaces should be from the config.

src/main/scala/com/raphtory/spouts/FileSpout.scala
```diff
@@ -86,6 +87,7 @@ class FileSpout[T: TypeTag](val path: String = "", val lineConverter: (String =>
 
   private def processFile(file: File) = {
     logger.info(s"Spout: Processing file '${file.toPath.getFileName}' ...")
+    SpoutTelemetry.totalFilesProcessed.inc()
```
Haaroon, line 90: is there anyway we can adjust this slightly, the counter should only increment after the getlines? Or it can increment where you have put it, but then decrement if the spout throws an error. 
Can we then add an error case, where we increment how many file errors the spout throws?

src/main/scala/com/raphtory/core/config/telemetry/SpoutTelemetry.scala
```diff
@@ -0,0 +1,30 @@
...
+      .register
+
+  val totalLinesParsed: Gauge =
+    Gauge
+      .build
+      .namespace("spout")
+      .name("total_lines_parsed")
+      .help("Total lines parsed")
+      .register
+
```
Haaroon, line 29: add counter to count errors, how many times the spout failed to process a file.

Haaroon (changes requested):
Thank you for this, great work. Please can you take a look at my comments.

src/main/scala/com/raphtory/core/components/spout/SpoutExecutor.scala
rutujasurve94: This won't throw an exception since we are only incrementing

src/main/scala/com/raphtory/core/client/RaphtoryGraph.scala
rutujasurve94: Sounds great, I've stored and stopped it post execution as well

src/main/scala/com/raphtory/spouts/FileSpout.scala
rutujasurve94: This would make the iterator not work. I've instead just placed it before calling `getLines`, we are also tracking errors in processed files now so we should be good to calculate.

Haaroon (approved):

PR #297, opened 2022-03-23
Telemetry for partition

rutujasurve94:
Adds metrics for `Partitions`:

- [x] - Metrics for `Readers`
- [x] - Metrics for `StreamWriter` 
- [x] - `BatchWriter`
- [x] - `LocalBatchHandler`

Counters for `vertex`, `edge`, `watermarks` created, messages received and sent

PR #301, opened 2022-03-25
Metrics for `GraphBuilder`

rutujasurve94:
Add metrics for `GraphBuilder`.

### DIFF
--- a/src/main/scala/com/raphtory/core/components/graphbuilder/GraphBuilder.scala
+++ b/src/main/scala/com/raphtory/core/components/graphbuilder/GraphBuilder.scala
@@ -178,7 +178,6 @@ trait GraphBuilder[T] extends Serializable {
     batchWriters = writers
     batching = true
     totalPartitions = partitions
-    GraphBuilderTelemetry.totalBuilderPartitions.set(partitions)
   }
 
   protected def addVertex(updateTime: Long, srcId: Long): Unit = {
--- a/src/main/scala/com/raphtory/core/components/graphbuilder/GraphBuilder.scala
+++ b/src/main/scala/com/raphtory/core/components/graphbuilder/GraphBuilder.scala
@@ -1,6 +1,7 @@
 package com.raphtory.core.components.graphbuilder
 
 import com.raphtory.core.components.partition.BatchWriter
+import com.raphtory.core.config.telemetry.GraphBuilderTelemetry
 import com.typesafe.scalalogging.Logger
 import net.openhft.hashing.LongHashFunction
 import org.slf4j.LoggerFactory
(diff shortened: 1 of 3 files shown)


src/main/scala/com/raphtory/core/config/telemetry/GraphBuilderTelemetry.scala
```diff
@@ -0,0 +1,68 @@
...
+  * {s}`GraphBuilderTelemetry`
+  *  : Adds metrics for {s}`GraphBuilder` using Prometheus Client
+  *
+  *    Exposes Counter and Gauge stats for tracking number of vertices and edges added and deleted, total partitions created
+  *    by the graph builder
+  *    Statistics are made available on http://localhost:8899 on running tests and can be visualised using Grafana dashboards
+  *
+  *
+  */
+object GraphBuilderTelemetry {
```
miratepuffin, line 16: Not sure if the names here are needed to add in the builderID. Might be an idea so we can see if one is lagging behind etc.

src/main/scala/com/raphtory/core/config/telemetry/GraphBuilderTelemetry.scala
```diff
@@ -0,0 +1,68 @@
...
+
+  val totalGraphBuilderUpdates: Counter =
+    Counter
+      .build
+      .namespace(conf.getString("raphtory.prometheus.builder_namespace"))
+      .name("graph_builder_total_updates")
+      .help("Total graph builder updates")
+      .register
+
+  val totalBuilderPartitions: Gauge =
```
miratepuffin, line 60: Can be removed I think

miratepuffin (changes requested):

src/main/scala/com/raphtory/core/components/graphbuilder/GraphBuilder.scala
```diff
@@ -176,6 +178,7 @@ trait GraphBuilder[T] extends Serializable {
     batchWriters = writers
     batching = true
     totalPartitions = partitions
+    GraphBuilderTelemetry.totalBuilderPartitions.set(partitions)
```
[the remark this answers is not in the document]
rutujasurve94: Added this for convenience - to extract `application.conf` properties out as metrics, if helpful. I'm removing this, if redundant

src/main/scala/com/raphtory/core/config/telemetry/GraphBuilderTelemetry.scala
rutujasurve94: I've added a `builder_id` var to this object that could then be set as required.

miratepuffin:
#297 #301 #302 and #339 to be merged into one PR off of the new Development branch so they can bee all tested together and merged in.

PR #309, opened 2022-03-29
Resource leaks and cleaned up BaseRaphtoryAlgoTest.scala file

Alnaimi-:


core/src/test/scala/com/raphtory/BaseRaphtoryAlgoTest.scala
```diff
@@ -1,61 +1,64 @@
...
-  val conf             = graph.getConfig()
-  val pulsarController = new PulsarController(conf)
 
-  val pulsarAddress: String =
-    conf.getString("raphtory.pulsar.broker.address") //conf.getString("Raphtory.pulsarAddress")
-  val deploymentID: String = conf.getString("raphtory.deploy.id")
+  val conf: Config          = graph.getConfig()
+  val pulsarAddress: String = conf.getString("raphtory.pulsar.broker.address")
+  val deploymentID: String  = conf.getString("raphtory.deploy.id")
+  val pulsarController      = new PulsarController(conf)
```
ljeub-pometry, line 46: If we are cleaning things, it may be good to wrap this stuff into a `beforeAll` so the graph is only initialised when tests start (right now all graphs are being initialised when tests are discovered).

core/src/test/scala/com/raphtory/BaseRaphtoryAlgoTest.scala
```diff
@@ -1,61 +1,64 @@
...
-  implicit val schema: Schema[Array[Byte]] = Schema.BYTES
-  val testDir                              = "/tmp/raphtoryTest" //TODO CHANGE TO USER PARAM
+  before {
+    setup()
+
+    Raphtory.createClient(
+            deploymentID = "deployment123",
+            customConfig = Map(("raphtory.pulsar.endpoint", "localhost:1234"))
+    )
+  }
```
ljeub-pometry, line 55: What is up with the client here? What is it used for?

Also, isn't this run once per test? The `setup()` function currently is only run once.

core/src/test/scala/com/raphtory/BaseRaphtoryAlgoTest.scala
```diff
@@ -1,61 +1,64 @@
...
+
+  after {
+    if (deleteResultAfterFinish)
+      FileUtils
+        .cleanDirectory(new File(outputDirectory))
+  }
```
ljeub-pometry, line 61: Is there a way to make the cleanup only run for successful tests? This would require making the cleanup more specific as we only want to delete the folder associated with the current test, not the whole output directory). Would be good to be able to inspect the output of failed tests...

core/src/test/scala/com/raphtory/BaseRaphtoryAlgoTest.scala
```diff
@@ -73,18 +76,33 @@ abstract class BaseRaphtoryAlgoTest[T: ClassTag: TypeTag]
...
+      finally source.close()
+    }.sorted
+
+    val hash = Hashing
+      .sha256()
+      .hashString(results.mkString, StandardCharsets.UTF_8)
+      .toString
+
     logger.info(s"Generated hash code: '$hash'.")
+
```
ljeub-pometry, line 105: factor this out into a separate function as it is used again below for `algorithmPointTest`

ljeub-pometry (commented):
- I've got some doubts about the use of `before` actually doing what we want

- It would be really good if we can keep the output for failed tests for easier debugging

- Minor suggestions for further cleanup

PR #318, opened 2022-03-30
Fix error due to calling head on empty list in perspective controller

ljeub-pometry:
- No more exception
- Still have failing test due to missing data (race condition?)


ricopinazo:
I am changing that

miratepuffin (approved):
Great job lads!

PR #321, opened 2022-03-30
Bug/fix failure test

ljeub-pometry:
fixes failure test and hang in case of failing algorithms

### DIFF
--- a/core/src/main/scala/com/raphtory/spouts/SequenceSpout.scala
+++ b/core/src/main/scala/com/raphtory/spouts/SequenceSpout.scala
@@ -1,0 +1,22 @@
+package com.raphtory.spouts
+
+import com.raphtory.components.spout.Spout
+
+/**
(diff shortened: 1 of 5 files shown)


core/src/main/scala/com/raphtory/spouts/SequenceSpout.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import com.raphtory.components.spout.Spout
+
+/**
+  * {s}`SequenceSpout[T](seq: T*)`
+  *  : output input messages to Raphtory
+  *
+  * This spout is mainly useful for testing functionality
+  */
+class SequenceSpout[T](seq: Seq[T]) extends Spout[T] {
```
miratepuffin, line 11: This is a very cool new class.

miratepuffin (approved):

PR #325, opened 2022-03-31
Feature/apply algorithms to raphtory graph

ljeub-pometry:
- make RaphtoryGraph inherit GraphPerspective
- internal implicit conversion from GraphPerspective to the generic type means we can use a GraphAlgorithm to implement a method of DefaultGraphOperations or its subclasses
- massively simplifies the `transform` method and will enable porting over the new Vertex filtering code


ricopinazo:
My main concern here is that all this modifies the class hierarchy in a way that is maybe confusing for the user. I think an alternative is to implement the implicit conversion like this:

```scala
implicit protected def graphOperationsFromPerspective(graph: GraphPerspective): G = {
    val graphImpl = graph.asInstanceOf[GenericGraphPerspective]
    newGraph(graphImpl.query, graphImpl.querySender)
  }
```

This way we don't need to change anything in the class hierarchy.

ljeub-pometry:
How does your suggestion address the problem that RaphtoryGraph is not a GraphPerspective and can thus not be used as the input to an algorithm? We could maybe fix that with another implicit conversion but I think making it actually extend GraphPerspective is clearer...

In any case, I don't think the new hierarchy is that confusing?

ricopinazo:
The intention behind having a trait `GraphOperations` is precisely to completely separate the concepts of `RaphtoryGraph` and `GraphPerspective`, as they are different things. `GraphOperations` states the operations the user can apply over a graph. `RaphtoryGraph` allows those operations and the same `GraphPerspective` does, but a `GraphPerspective` could potentially add some functionality that we don't want to allow from a `RaphtoyGraph`. That is my inital idea and of course it can be changed, but I would like to understand the problem to be sure there isn't another way.

Regarding the intention of this PR, if we want to use the logic of an algorithm to implement a new method for `RaphtoryGraph` or `graphOperations`, as the `transform` method exist, we can do something like:
```scala
def functionality(parameter: Parameter, ...) = transform(Functionality(parameter, ...))
```
Is this something helpful in some way? I would like to know more details to be able to help, because I am probably missing several things. For instance, I think I am not aware of how the vertex filtering is being implemented.

PR #328, opened 2022-03-31
Feature/apply algorithms to raphtory graph recreated

miratepuffin:


ljeub-pometry:
After chatting with @ricopinazo, we think there is a better approach to this problem. Going to close this for now and there will most likely be a future pull request with some deeper restructuring of the API.

PR #329, opened 2022-03-31
Fix infinite loop bug in Binary Diffusion, simplify algorithm

miratepuffin:


miratepuffin:
@Haaroon Can you add some documentation to this. please.

PR #332, opened 2022-04-01
Dev 191 - Vertex Filtering

ljeub-pometry:
This is work by @rachchan that I rescued across the great divide of the project restructure and API changes

- Working implementation of filtering
- Still need to implement resetting the filter (future work)
- Still need to implement edge filter (future work)


core/src/main/scala/com/raphtory/algorithms/api/DefaultGraphOperations.scala
```diff
@@ -13,10 +13,10 @@ abstract class DefaultGraphOperations[G <: GraphOperations[G]](
 ) extends GraphOperations[G] {
   override def setGlobalState(f: GraphState => Unit): G = addFunction(Setup(f))
 
-  override def filter(f: (Vertex) => Boolean): G = addFunction(VertexFilter(f))
+  override def filter(f: (Vertex) => Boolean): G = step(vertex => if (!f(vertex)) vertex.remove())
```
miratepuffin, line 16: This is 200iq

core/src/main/scala/com/raphtory/components/partition/QueryExecutor.scala
```diff
@@ -104,6 +103,13 @@ class QueryExecutor(
           graphLens.receiveMessage(msg)
           receivedMessageCount.addAndGet(1)
 
+        case msg: FilteredEdgeMessage                                         =>
+          logger.trace(
+                  s"Job '$jobID' at Partition '$partitionID': Executing 'FilteredEdgeMessage', '$msg'."
+          )
+          graphLens.receiveMessage(msg)
+          receivedMessageCount.addAndGet(1)
```
miratepuffin, line 111: Was the counting moved outside because have the two different types of messages now?

core/src/main/scala/com/raphtory/components/querymanager/AnalysisMessages.scala
```diff
@@ -69,8 +69,23 @@ case object RecheckTime                 extends QueryManagement
...
+sealed trait GenericVertexMessage extends QueryManagement {
+  def superstep: Int
+  def vertexId: Long
+}
+
+case class VertexMessage[+T](superstep: Int, vertexId: Long, data: T) extends GenericVertexMessage
+
+case class VertexMessageBatch(data: Array[GenericVertexMessage]) extends QueryManagement
+
+case class FilteredEdgeMessage(superstep: Int, vertexId: Long, sourceId: Long)
```
miratepuffin, line 81: Is this needed when we have In and Out? or is this for both In/Out?

core/src/main/scala/com/raphtory/graph/visitor/Vertex.scala
```diff
@@ -379,6 +379,15 @@ trait Vertex extends EntityVisitor {
   def isInNeighbour(id: Long): Boolean  = inNeighbourSet.contains(id)
   def isOutNeighbour(id: Long): Boolean = outNeighbourSet.contains(id)
 
+//  def removeInNeighbour(id: Long)
```
miratepuffin, line 382: These a TODO for later?

core/src/main/scala/com/raphtory/storage/pojograph/entities/external/PojoExVertex.scala
```diff
@@ -37,44 +41,64 @@ class PojoExVertex(
     queue
   }
 
+  private var incomingEdgeDeleteMultiQueue: VertexMultiQueue = new VertexMultiQueue()
+  private var outgoingEdgeDeleteMultiQueue: VertexMultiQueue = new VertexMultiQueue()
+  private var filtered                                       = false
+
+  def executeEdgeDelete(): Unit = {
+    internalOutgoingEdges --= outgoingEdgeDeleteMultiQueue
```
miratepuffin, line 49: Not a point for now, but we should have a look at how slow --= is

core/src/main/scala/com/raphtory/components/querymanager/AnalysisMessages.scala
miratepuffin: Can't see it being used, but I could be totally blind.

core/src/main/scala/com/raphtory/components/partition/QueryExecutor.scala
ljeub-pometry: Counting was always here for optimising the VertexMessageBatch case?

core/src/main/scala/com/raphtory/components/querymanager/AnalysisMessages.scala
ljeub-pometry: I think we put this in initially but then didn't use it. Maybe we should actually use it as we could reduce the number of messages sent.

core/src/main/scala/com/raphtory/graph/visitor/Vertex.scala
ljeub-pometry: These are Stubbs from thinking about options for doing edge filtering (though probably not how we actually should implement it)

core/src/main/scala/com/raphtory/components/partition/QueryExecutor.scala
miratepuffin: Sorry yes you are correct I was just misremembering - thought I had moved it inside with the prior refactor

core/src/main/scala/com/raphtory/components/querymanager/AnalysisMessages.scala
miratepuffin: Ok cool, we can revisit this then when we do the expansion of the filtering you mention in the PR notes

miratepuffin (approved):
Very happy with all of this. Certainly a lot we can do with this as noted in the comments on the PR, but can merge this in for now. Is great to have this functionality in.

PR #333, opened 2022-04-01
Update docs for the graph classes

ricopinazo:


miratepuffin (approved):
Just had a read through @ricopinazo and it looks great. I think we should take a portion of this and turn it into its own top level tutorial file for the docs in-line with the Raphtory Algorithm explanation.

PR #356, opened 2022-04-11
Feature/multilayer view algorithms

ljeub-pometry:
Make use off and test multilayer views (WIP)


ljeub-pometry:
- algorithms that call `multilayerView` can now make use of `ExplodedVertex` methods without using `asInstanceOf`
- `GraphFunction`s no longer have explicit `Vertex` type to allow passing along functions using `ExplodedVertex` or `Vertex`
- The fact that `RaphtoryGraph` directly exposes graph operations means that we require a lot of abstract types. 
- `RaphtoryGraph` also does not support multilayer views properly which means using `asInstanceOf` is necessary when using `step`/`iterate` from `RaphtoryGraph` directly

ljeub-pometry:
Documentation needs to be updated for the new type parameters once this is settled

core/src/main/scala/com/raphtory/algorithms/temporal/community/MultilayerLPA.scala
```diff
@@ -66,12 +66,12 @@ class MultilayerLPA(
     layerSize: Long,
     omega: Double = 1.0,
     seed: Long = -1
-) extends GraphAlgorithm {
+) extends Generic {
```
miratepuffin, line 69: Can't wait to see this one properly rewritten!

core/src/main/scala/com/raphtory/api/algorithm/Multilayer.scala
```diff
@@ -0,0 +1,79 @@
...
+          with Multilayer {
+
+    override def apply(graph: MultilayerGraphPerspective): graph.MultilayerGraph =
+      second(first(graph).clearMessages())
+
+    override def tabularise(graph: MultilayerGraphPerspective): Table =
+      second.tabularise(graph)
+  }
+
+  case class ChainedMultilayer2(
```
miratepuffin, line 26: I am not saying I have a better one, but we need to come up with a better name for this 😄

core/src/main/scala/com/raphtory/api/graphview/Deployment.scala
```diff
@@ -0,0 +1,17 @@
...
+
+/** Public interface for the Raphtory deployment
+  *
+  * @see [[DeployedTemporalGraph]]
+  */
+trait Deployment {
+
+  /** The configuration used by the deployment
+    */
+  val conf: Config
```
miratepuffin, line 13: Cheers for adding this!

core/src/main/scala/com/raphtory/api/graphview/RaphtoryGraph.scala
```diff
@@ -0,0 +1,36 @@
...
+import com.raphtory.client.QuerySender
+import com.raphtory.components.querymanager.Query
+
+/** Reduced GraphView with fixed timeline
+  *
+  * This [[GraphView]] is returned by the [[DottedGraph]] operations and does not support further timeline manipulation.
+  *
+  * @see [[GraphView]], [[MultilayerRaphtoryGraph]], [[DottedGraph]]
+  */
+class RaphtoryGraph private[api] (
```
miratepuffin, line 12: This is an odd way round to have the private no?

miratepuffin (approved):
As discussed we need to finish the docs (which are looking *chef's kiss*) but all good to go!

PR #357, opened 2022-04-11
DEV-837 Clean up Pulsar topics for algorithms

rachchan:


core/src/main/scala/com/raphtory/config/PulsarController.scala
```diff
@@ -144,8 +144,16 @@ class PulsarController(conf: Config) {
...
+  def deleteTopic(topic: String, force: Boolean = false) =
+    try {
+      pulsarAdmin.topics().delete(topic, force)
+      logger.debug(s"deleted topic $topic")
+    }
+    catch {
+      case error: Exception =>
+        logger.warn(s"failed to delete $topic")
+        error.printStackTrace()
+    }
```
ljeub-pometry, line 156: Remove this since it cannot be used safely ever?

ljeub-pometry (changes requested):

ljeub-pometry:
Also, need to fix the PulsarOutputTest which is creating a consumer outside of the normal internal stuff that needs to unsubscribe

core/src/main/scala/com/raphtory/client/GraphDeployment.scala
```diff
@@ -53,16 +53,6 @@ private[raphtory] class GraphDeployment[T: ClassTag: TypeTag](
   private val spoutTopic: String   = conf.getString("raphtory.spout.topic")
   private val prometheusPort: Int  = conf.getInt("raphtory.prometheus.server")
 
-  private val zookeeperAddress: String = conf.getString("raphtory.zookeeper.address")
-
-  private val partitionIdManager =
```
miratepuffin, line 58: Has this been moved into the subclass or just deleted?

miratepuffin: Ok nvm I found it

miratepuffin (approved):

ljeub-pometry (approved):

PR #364, opened 2022-04-13
Update root graph object for the queries

ricopinazo:
The documentation for the Raphtory object was not updated in the past and is not very clear in some bits, I am trying to also fix that.

### DIFF
--- a/core/src/main/scala/com/raphtory/deployment/Raphtory.scala
+++ b/core/src/main/scala/com/raphtory/deployment/Raphtory.scala
@@ -114,7 +114,7 @@ import scala.reflect.runtime.universe._
   * import com.raphtory.GraphState
   * import com.raphtory.output.FileOutputFormat
   *
-  * val builder = ???
+  * val builder = new YourGraphBuilder()
   * val customConfig = Map(("raphtory.pulsar.endpoint", "localhost:1234"))
   * val graph = Raphtory.stream(ResourceSpout("resource"), builder, customConfig)
   * graph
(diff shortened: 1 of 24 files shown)


core/src/main/scala/com/raphtory/deployment/Raphtory.scala
```diff
@@ -45,38 +106,64 @@ import scala.reflect.runtime.universe._
...
   *
   * ```{code-block} scala
-  *
   * import com.raphtory.deployment.Raphtory
-  * import com.raphtory.lotrtest.LOTRGraphBuilder
   * import com.raphtory.components.spout.instance.ResourceSpout
   * import com.raphtory.GraphState
   * import com.raphtory.output.FileOutputFormat
   *
+  * val builder = ???
```
miratepuffin, line 117: Could change this to something like val builder = new YourGraphBuilder()

core/src/main/scala/com/raphtory/deployment/Raphtory.scala
```diff
@@ -45,38 +106,64 @@ import scala.reflect.runtime.universe._
...
+
+  def deployedGraph(customConfig: Map[String, Any] = Map()): TemporalGraph = {
+    val conf             = confBuilder(customConfig)
+    val pulsarController = new PulsarController(conf)
+    val componentFactory = new ComponentFactory(conf, pulsarController)
+    val querySender      = new QuerySender(componentFactory, scheduler, pulsarController)
+    new TemporalGraph(Query(), querySender, conf)
+  }
+
+  @deprecated("Use Raphtory.stream(...) instead", "Raphtory 0.5.0")
```
miratepuffin, line 166: We can removed these once #361 is merged in.

miratepuffin (changes requested):
Just need to remove the deprecated functions and update the examples and can 🚢

miratepuffin (approved):

PR #367, opened 2022-04-14
Clean up generic types and documentation

ljeub-pometry:
- `Edge` and `ExplodedEdge` now have an abstract type member rather than taking type parameters
  * this is going to make things like writing generic edge filters much neater as we do no longer have to worry about type parameters
  * implemented intermediate trait (not user-facing) to turn abstract type into type parameter for passing along `IDType` from vertex
- cleaned up the documentation


miratepuffin (approved):
This all makes sense. I think we may need to make a graph of all these types though, even just for internal documentation, to keep track of the hierarchy. Probably need to clean up a couple of these folders as well before the final release.

PR #369, opened 2022-04-14
Feature/dev 844 new api docs

ricopinazo:


ricopinazo:
I have added a subsection explaining how to use `Raphtory.deployedGraph()` and the legacy client in queries. It can be left here or we can put this on a new page. In the latter case, I am not aware of the details of the deployment, but the subsection about the client would be something like this.

miratepuffin (approved):

PR #373, opened 2022-04-22
Add communication layer and set Akka for local deployments

ricopinazo:
The main change to look at is the addition of the `TopicRepository` class. This class rely on `Connector` objects that might have different underlying implementations to register components and to create endpoints to send messages to. There are implementations currently for Pulsar and Akka.

In the `Raphtory` object, for distributed deployments, we use a `TopicRepository` based only in a `PulsarConnector`:

```scala
new TopicRepository(new PulsarConnector(config), config)
```

For local deployments, we use a `TopicRepository` that uses an `AkkaConnector` to manage control messages. In order to do this, we override set a `PusarConnector` as default connector in the `TopicRepository` and then we override the connectors used for the control topics:

```scala
new TopicRepository(pulsarConnector, config) {
  override def jobOperationsConnector: Connector = akkaConnector
  override def jobStatusConnector: Connector     = akkaConnector
  override def queryPrepConnector: Connector     = akkaConnector
  override def endedQueriesConnector: Connector  = akkaConnector
  override def watermarkConnector: Connector     = akkaConnector
  override def rechecksConnector: Connector      = akkaConnector
  override def queryTrackConnector: Connector    = akkaConnector
  override def submissionsConnector: Connector   = akkaConnector
}
```

Another important changes:
- `ZookeeperIDManager` now extends `IDManager` trait, which has a local version: `LocalIDManager`
- There is an `EndPoint` trait that has a different implementation regarding the connector that creates it
- There is a `Topic` trait with four different implementations: `ExclusiveTopic`, `WorkPullTopic`, `BroadcastTopic`, `ShardingTopic`. The connector implementations need to handle the first three, whereas the last one is implemented based on the `ExclusiveTopic`.

Main concerns:
- There `application.conf` file has grown a lot because there are parameters to configure every topic as they are not attached to an specific component in the definition.
- I created a wrapper `MonixScheduler` extending a trait `Scheduler` with the intention of creating an Akka version `AkkaScheduler`. Now that we are not using the Akka scheduler as the global scheduler, the abstraction doesn't make any sense. We can rollback those changes or we can let them as they are to switch the implementation of the scheduler in the future.
- In the Akka implementation for the `EndPoint`, the `sendAsync` method is effectively async when the actor ref is built. However, if there are delays because a component needs to be deployed and registered as an actor, the first call to `sendAsync` might be blocking. This can be solved using higher abstractions provided by Akka, but might be a good balance for now.


core/src/main/resources/application.conf
```diff
@@ -22,6 +22,188 @@ raphtory {
       size                  = ${?RAPHTORY_RETENTION_SIZE}
 
     }
+    topics {
```
ricopinazo, line 25: The topics are defined by and `id` and a `subTopic`. We could define them by a `group` and `id` and a `subTopic` instead. This way we would be able to manage the configuration per topic group to simplify `application.conf`. It would be similar to the previous sorting but decoupled from the components.

core/src/main/scala/com/raphtory/components/querymanager/QueryManager.scala
```diff
@@ -14,32 +14,32 @@ import org.apache.pulsar.client.api.Schema
...
-    cancelableConsumer match {
-      case Some(value) =>
-        value.unsubscribe()
-        value.close()
-      case None        =>
-    }
+    listener.close()
     currentQueries.foreach(_._2.stop())
-    watermarkGlobal.close()
+    // watermarkGlobal.close() TODO: remove?
```
miratepuffin, line 42: Lets turn it off for the time being to simplify the akka/pulsar questions and then can turn back on/fix it if people require it.

core/src/main/scala/com/raphtory/config/LocalIDManager.scala
```diff
@@ -0,0 +1,16 @@
+package com.raphtory.config
+
+/** @DoNotDocument */
+class LocalIDManager extends IDManager {
+  private var nextId = 0
```
miratepuffin, line 5: Lets change this to an AtomicInt just to be safe

core/src/main/scala/com/raphtory/storage/pojograph/messaging/VertexMessageHandler.scala
```diff
@@ -20,12 +21,11 @@ import scala.concurrent.Future
 /** @DoNotDocument */
 class VertexMessageHandler(
     config: Config,
-    producers: Map[Int, Producer[Array[Byte]]],
+    producers: Option[Map[Int, EndPoint[QueryManagement]]],
```
miratepuffin, line 24: rename to endPoints?

miratepuffin (changes requested):

miratepuffin (approved):

PR #379, opened 2022-04-25
DEV663 Updated Python client to support py4j and submitted raphtory algorithms

Haaroon:
This is an updated version of the Python Client.
Summary: client can now initiate raphtory scala algorithms from python via py4j, pypi has been migrated 

Changelog Raphtory
- Added a Py4j server to Raphtory which starts when the raphtory instance deploys a graph 

Changelog Python
- Renamed client to pulsar_client
- Added JVM gateway with support via py4j 
- Added helper functions to default import some classes
- Moved to use new deployedgraph api
- Moved pypi github action to this repository with secret keys
- Added new workflow that builds from sub-directory
- fixed a bug where dataframe errored if columns were uneven 
- Updated docs
- Updated lotr example to work with python


core/src/main/scala/com/raphtory/algorithms/api/Table.scala
```diff
@@ -40,4 +40,5 @@ abstract class Table {
   def filter(f: Row => Boolean): Table
   def explode(f: Row => List[Row]): Table
   def writeTo(outputFormat: OutputFormat, jobName: String = ""): QueryProgressTracker
```
ricopinazo, line 42: We can remove the default value for jobName so that the two parameters version is also accessible from Python

Haaroon: will do! then i will have to change all examples so they don't break.

build.sbt
```diff
@@ -144,6 +146,7 @@ lazy val dependencies = new {
   val pulsarCommon     = "org.apache.pulsar"           % "pulsar-common"                  % pulsarVersion
   val pulsarCrypto     = "org.apache.pulsar"           % "pulsar-client-messagecrypto-bc" % pulsarVersion
   val pulsarOriginal   = "org.apache.pulsar"           % "pulsar-client-original"         % pulsarVersion
+  val py4j             = "net.sf.py4j"                 % "py4j"                           % py4jVersion excludeAll (excludeLog4j, excludeSlf4j)
```
miratepuffin, line 149: My man remembers the exclusion 🤩

core/src/main/scala/com/raphtory/algorithms/api/GenericTable.scala
```diff
@@ -17,11 +17,15 @@ class GenericTable(val query: Query, private val querySender: QuerySender) exten
     addFunction(Explode(closurefunc))
   }
 
-  override def writeTo(outputFormat: OutputFormat, jobName: String = ""): QueryProgressTracker = {
+  override def writeTo(outputFormat: OutputFormat, jobName: String): QueryProgressTracker = {
     val query = addFunction(WriteTo(outputFormat)).query
     querySender.submit(query, jobName)
   }
 
+   override def writeTo(outputFormat: OutputFormat): QueryProgressTracker = {
```
miratepuffin, line 25: Ahhh is this because python can't deal with the default arguments?

examples/raphtory-example-lotr/python/LOTR_demo.ipynb
```diff
@@ -16,9 +16,10 @@
    },
    "source": [
     "In this demo we will\n",
-    "* Create a Python Raphtory client\n",
+    "* Setup the Python Raphtory client \n",
```
miratepuffin, line 19: Can you clear out all the data from this so it doesn't bloat the line count.

python/raphtoryclient/client.py
```diff
@@ -198,14 +290,14 @@ def getDataframe(self, reader, delimiter=',', max_messages=sys.maxsize, col_name
                 if wait_counter == 6:
                     break
         print("Converting to columns...")
-        if len(messages) != 0:
-            num_cols = len(messages[0])
-            if len(col_names):
-                for i in range(len(col_names), num_cols, 1):  col_names.append("result_" + str(i - len(col_names)))
-            else:
-                for i in range(0, num_cols, 1):  col_names.append("result_" + str(i))
+        # if len(messages) != 0:
```
miratepuffin, line 293: If not needed, can this be removed?

miratepuffin (changes requested):

core/src/main/scala/com/raphtory/algorithms/api/GenericTable.scala
Haaroon: yes, py4j uses java bindings of the scala code which do not have default args

PR #382, opened 2022-04-26
Ignore out-of-order messages as these are duplicates

ljeub-pometry:
Pulsar message listeners seem to have an internal timeout which causes duplicate processing of messages. This uses the last messageID to check that the new message is later in the topic than previously seen messages and ignores any out-of-order messages during processing.


miratepuffin (changes requested):
Just need to make a slight loggin adjustment.

PR #385, opened 2022-05-03
Nice default job name in case of calling graph.execute with an algorithm

ljeub-pometry:


ricopinazo:
It definitely looks better this way. On top of this, two ideas in case they are helpful:
1. In the case they call `execute` several times, maybe we can append the name to the older one. For instance: `graph.execute(CustomAlgo).execute(NodeList)` might produce `"CustomAlgo-NodeList"` instead of just `"NodeList"`. This approach doesn't scale very well though.
2. Another option that scales well: if they call execute and there is already a name, we can remove it to avoid confusion.

Just to give you some ideas, I don't know which option is better. I think your solution is perfectly fine the way it is.

miratepuffin (approved):

PR #387, opened 2022-05-03
0.5.0 Streamline tutorial and graph.stop fixes

miratepuffin:
This is turning into a bit of a mega branch, but this is what has currently been done:

- [x] Documentation for introducing raphtory now fully in line with development/0.5.0. 
- [x] Raphtory can now be run without Pulsar and the documentation is updated on explaining how.
- [x] Changed the order of all docs to fix the flow.
- [x] Tested Examples to make sure they work as a fat jar.
- [x] Removed old Raphtory Client and all references to it. 
- [x] Removed Prometheus stack trace from being printed when multiple raphtory instances are running (needs a proper fix).
- [x] ~Some sort of issue with unsubscribing to pulsar topics~ This has been solved now by catching the exception and just closing the connection
image

- [x] Added `TemporalGraphConnection` class which is returned from the `deployedGraph` function - this has a disconnect function which allows us to release all the objects created during the call to `deployedGraph`.

- [x] graph still holding on to a multitude of resources (ports, threads, memory) when .stop is called - this is causing the tests to fail when running in their entirety.  Fixed by each graph deployment having its own connector which is closed at the end of the deployments lifetime.
image

- [x] Multiple running graphs are not able to have a Prometheus instance as the port is currently hogged by the first one to grab it. Fixed by having a global Prometheus server.
- [x] Lots of public loggers and other objects to make private (these are now all showing up in the scala docs) - Project refactored to hide all privatable variables. Also had a tidy of old classes/imports/funcs.


docs/source/Analysis/queries.md
```diff
@@ -77,73 +62,58 @@ graph
...
-we can set up a walk along a prefiltered year of our data in steps of one day
-creating a window of one week in each step as follows:
+In this example, starting from January 1 2020 we move forward one day at a time, looking back over only the last data of data. At each of these stopping points (`perspectives`) we execute the algorithm and write the results to the file. If we set up a spout from a streaming source which continues to ingest data, Raphtory will continue to create a new `windowed perspective` every day as the new information arrives.
+
+As can be seen in the example, the process to create perspectives has two steps. The first of these is setting the times you are interested in, which can be a singular point (using `at()`) or, alternatively, a sequence of points with a given increment. For the latter, four different methods are available:
+
+| Function            | Effect                                            |
+|---------------------|---------------------------------------------------|
+| `depart(time,increment)`   | Set the start time and increment           |
+| `climb(time,increment)`    | Set the start time and increment           |
```
ricopinazo, line 86: In `climb(time, increment)`, `time` refers to the end of the sequence

docs/source/Analysis/queries.md
```diff
@@ -77,73 +62,58 @@ graph
...
+| `range(start,end,increment`| Set the start time, end time and increment |
+| `walk(increment)`          | Set only the increment                     |
+
+**Note:** If no start and end time are provided, Raphtory will default to the minimum and maximum times in the data (or min and max of the time range if a `silce()` etc. has been applied).
```
ricopinazo, line 90: There is a typo here: `slice()` instead of `silce()`

docs/source/Ingestion/sprouter.md
```diff
@@ -1,17 +1,17 @@
...
 
-If you have downloaded the [Examples](https://github.com/Raphtory/Examples.git) folder from the installation guide previously, then the LOTR example is already set up. If not, please return there and complete this step first.  
+If you have the LOTR example already set up from the installation guide previously ([raphtory-example-lotr](https://github.com/Raphtory/Raphtory/tree/master/examples/raphtory-example-lotr)) then please continue. If not, please return there and complete this step first.  
 
-For this tutorial section we will continue to use the `raphtory-example-lotr` project and a dataset that tells us when two characters have some type of interaction in the Lord of the Rings trilogy books. The `csv` file (comma-separated values) in the examples folder can be found [here](https://github.com/Raphtory/Examples/blob/0.5.0/raphtory-example-lotr/resources/lotr.csv). Each line contains two characters that appear in the same sentence, along with which sentence they appeared in, indicated by a number (sentence count). In the example, the first line of the file is `Gandalf,Elrond,33` which tells us that Gandalf and Elrond appear together in sentence 33.
```
ricopinazo, line 14: This link is to the project Raphtory/Examples, is that ok?

docs/source/Analysis/queries.md
```diff
@@ -202,26 +172,22 @@ Running this algorithm returns the following data:
...
 val graph = Raphtory.deployedGraph()
 ```
 
 From this point, you can keep working with your graph as we have done so far.
 
-Additionally, you still have access to the `RaphtoryClient` class provided in previous releases of Raphtory:
+Additionally, you still have access to the `RaphtoryClient` class provided in previous releases of Raphtory. This is, however, deprecated and will be removed in later versions:
 
 ```scala
 val client = Raphtory.createClient()
```
ricopinazo, line 193: Are we going to remove this finally?

docs/source/Analysis/queries.md
```diff
@@ -202,26 +172,22 @@ Running this algorithm returns the following data:
...
+This data tells us that at a given time, person X and Gandalf are N number of hops away. In this instance, at time 32670, Samwise was at minimum 1 hop away from Gandalf, whereas Odo was 2 hops away.
 
 ## Using Raphtory as a client
 
-Finally, if you have a graph deployed somewhere else and want to submit new queries to it,
-you can do this via the `deployedGraph(customConfig)` method in `Raphtory` object,
-providing the appropriate configuration to locate it.
-If the graph is deployed in the same machine using the default configuration,
-you can even omit the configuration parameter:
+Finally, if you have a graph deployed somewhere else and want to submit new queries to it you can do this via the `deployedGraph(customConfig)` method in the `Raphtory` object. The `customConfig` here is to provide the appropriate configuration to locate the graph (i.e. the akka/pulsar address). If the graph is deployed in the same machine using the default Raphtory configuration you can omit this configuration parameter:
```
ricopinazo, line 183: > If the graph is deployed in the same machine using the default Raphtory configuration you can omit this configuration parameter

This is not true right now

docs/source/Analysis/queries.md
```diff
@@ -77,73 +62,61 @@ graph
...
+**Note:** If no start and end time are provided, Raphtory will default to the minimum and maximum times in the data (or min and max of the time range if a `slice()` etc. has been applied).
+
+The second step is to specify which direction we are looking in at each time point, for which we have three options. We can look to the `past`, to the `future`, or set a `window`. In the third case, we can align the time point as: 
+
+* The `Start` of the window - looking into the future the set amount of time. This is the default.
+* The `End` of the window - looking into the past the set amount of time. 
+* The `Middle` of the window - including data from half the time in both the past and future, providing a smoothing effect.
+
+
+**Note:** You can refer to the {scaladoc}`com.raphtory.algorithms.api.DottedGraph`} documentation for further details.
```
ricopinazo, line 99: typo here, extra bracket at the end:

> {scaladoc}`com.raphtory.algorithms.api.DottedGraph`}

docs/source/Ingestion/sprouter.md
miratepuffin: Good catch I will update it.

docs/source/Analysis/queries.md
miratepuffin: Yeah probably can do, I don't believe anyone is using it

docs/source/Analysis/queries.md
miratepuffin: Yeah this needs to be moved into the deployment section along with the update to streaming/batchload

core/src/main/scala/com/raphtory/config/MonixScheduler.scala
```diff
@@ -1,39 +1,64 @@
...
+      }
+      catch {
+        case e: RejectedExecutionException =>
+          logger.error(s"Scheduler rejected scheduling of tast $task")
+          None
+      }
+    else None
+
+  def shutdown(): Unit = {
+    schedulerStarted = false
```
Alnaimi-, line 60: Should this come after the shutdown in case of error?

core/src/main/scala/com/raphtory/deployment/Raphtory.scala
```diff
@@ -129,6 +112,7 @@ object Raphtory {
   def createGraphBuilder[T: ClassTag](
       builder: GraphBuilder[T]
   ): Unit = {
+    val scheduler        = new MonixScheduler()
```
Alnaimi-, line 115: Should this be a singleton object? Which means unique per JVM. Currently, new will create a new scheduler per component.

docs/source/Deployment/pulsarlocal.md
```diff
@@ -1,4 +1,63 @@
-# Pulsar
+# Raphtory Streaming and Apache Pulsar
+
+## Installing Apache Pulsar
+As of 0.5.0 Raphtory has moved away from Akka and now uses [Apache Pulsar](https://pulsar.apache.org) as its message broker to support communication between its distributed components. Using Pulsar means that all data sent within Raphtory is fully backed up and communication can easily scale to millions of messages a second. Pulsar also provides connectors to every conceivable source of data which can be used to pull data into Raphtory when building graphs.
+
+For any running instance of Raphtory, even local, we need to be able to connect to a Pulsar cluster. Fortunately, Pulsar comes with a standalone mode which packages everything together to be run once installed.  Pulsar can be installed in a variety of ways, which can be seen below. We have, however, create a [install script](../Deployment/pulsarlocal.md) which will automatically manage this for you, if you prefer.
```
Alnaimi-, line 6: Does that still apply?

Alnaimi-: If not, we should add a bit that says Pulsar is only required for multi-machine (distributed) deployments.

docs/source/Ingestion/sprouter.md
```diff
@@ -29,39 +29,39 @@ Gollum,Bilbo,308
...
-There are many data sources that may be used to feed graphs in Raphtory, for this example we will make use of the simplest of these, the `ResourceSpout`. This takes a file which is inside of the `resources` directory of an SBT project and pushes it into Pulsar for our graph builders to parse. We have already put the lotr.csv file into this folder (hence why it run in the last tutorial), so you shouldn't have to set anything if using it. If you are want to swap this file out for your own data, simply put your file into the resources directory and change the file name in the code.
+### File Spout
+There are many data sources that may be used to feed graphs in Raphtory, for this example we will make use of the `FileSpout`. This takes a file on your machine and pushes it into Pulsar for our graph builders to parse. We automatically download the lotr.csv file from the [Raphtory data repository](https://github.com/Raphtory/Data) (hence why it run in the last tutorial), so you shouldn't have to set anything if using it. If you are want to swap this file out for your own data, simply remove this download and change the `FileSpout` path to point to where your files are.
```
Alnaimi-, line 61: "If you are want" slight typo in that sentence

core/src/main/scala/com/raphtory/deployment/Raphtory.scala
miratepuffin: I am experimenting with this at the moment. If we have a separate scheduler per deployment we can kill all the threads at a top-level.

core/src/main/scala/com/raphtory/config/MonixScheduler.scala
miratepuffin: Maybe, the reason I put it first is that even if something goes wrong with the scheduler shutdown this will stop any new tasks from executing.

Alnaimi-: Is it possible to make the managing thread single threaded? So that other tasks can't be spawned while shutdown is being called?

miratepuffin: I don't believe so - lets discuss tomorrow though

miratepuffin: I have wrapped everything in locks 🔒

core/src/main/scala/com/raphtory/algorithms/api/TemporalGraphConnection.scala
```diff
@@ -0,0 +1,20 @@
+package com.raphtory.algorithms.api
+
```
Haaroon, line 2: Add explanation in scala header style

core/src/main/scala/com/raphtory/algorithms/api/TemporalGraphConnection.scala
```diff
@@ -0,0 +1,20 @@
...
+class TemporalGraphConnection(
+    query: Query,
+    private val querySender: QuerySender,
+    private val conf: Config,
+    private val scheduler: MonixScheduler,
+    private val topics: TopicRepository
+) extends TemporalGraph(query, querySender, conf) {
+
+  def disconnect(): Unit =
```
Haaroon, line 17: add explanation for this fn

core/src/main/scala/com/raphtory/config/MonixScheduler.scala
```diff
@@ -1,39 +1,106 @@
...
-  override def execute[T](component: Component[T]): Unit = scheduler.execute(() => component.run())
+  def execute[T](component: Component[T]): Unit = {
+    lock.lock()
+    if (schedulerRunning)
+      try scheduler.execute(() => component.run())
+      catch {
+        case e: RejectedExecutionException =>
+          logger.error(s"Scheduler rejected scheduling of Component $component")
+      }
+      finally lock.unlock()
```
Haaroon, line 51: no need for finally and else, just change to lock.unlock() at the end?

core/src/main/scala/com/raphtory/config/MonixScheduler.scala
```diff
@@ -1,39 +1,106 @@
...
+    lock.lock()
+    if (schedulerRunning) {
+      Task
+        .parSequenceUnordered(tasks)
+        .runAsync {
+          case Right(_)                   => onSuccess
+          case Left(exception: Exception) => errorHandler(exception)
+        }
+      lock.unlock()
+    }
```
Haaroon, line 88: remove duplicate lock.unlock() outside as the final statement

core/src/main/scala/com/raphtory/storage/pojograph/PojoGraphLens.scala
```diff
@@ -33,20 +35,24 @@ final case class PojoGraphLens(
...
+    private val errorHandler: (Throwable) => Unit,
+    private val scheduler: MonixScheduler
 ) extends GraphLens(jobId, start, end)
         with LensInterface {
+  private val logger: Logger = Logger(LoggerFactory.getLogger(this.getClass))
+
   private val voteCount         = new AtomicInteger(0)
   private val vertexCount       = new AtomicInteger(0)
-  var t1                        = System.currentTimeMillis()
+  private var t1                = System.currentTimeMillis()
```
Haaroon, line 46: unused var t1 is just overwritten and not used anywhere?

docs/source/Ingestion/sprouter.md
```diff
@@ -1,17 +1,17 @@
...
-- `Spouts` help with reading the data file and output tuples.
+- `Spouts` connect to the outside world, reading the data files and outputting individual tuples.
 - `Graph builders`, as the name suggests, convert these tuples into updates, building the graph.
 
-Once these classes are defined, they can be passed to the `stream()` or `batchLoad()` methods on the `Raphtory` object, which will use both components to build the temporal graph.  
+Once these classes are defined, they can be passed to the `stream()` or `batchLoad()` methods on the {scaladoc}`com.raphtory.deployment.Raphtory` object, which will use both components to build the {scaladoc}`com.raphtory.algorithms.TemporalGraph`. The difference here is that `stream()` will launch the full pipeline on top of [Apache Pulsar](https://pulsar.apache.org) (which you will see later in the tutorial) and assume new data can continuously arrive. `batchLoad()` on the other hand will compress the Spout and Graph Builder functions together, running as fast as possible, but only on a static dataset which does not change. For these initial examples we will only run `batchLoad()` as the data is static and we can set it going out of the box!
```
Haaroon, line 10: make steam and batch load two separate lines so its easier to read,

docs/source/Ingestion/sprouter.md
```diff
@@ -1,17 +1,17 @@
...
 
-If you have downloaded the [Examples](https://github.com/Raphtory/Examples.git) folder from the installation guide previously, then the LOTR example is already set up. If not, please return there and complete this step first.  
+If you have the LOTR example already set up from the installation guide previously ([raphtory-example-lotr](https://github.com/Raphtory/Raphtory/tree/master/examples/raphtory-example-lotr)) then please continue. If not, please return there and complete this step first.
```
Haaroon, line 12: add you shall not pass.

Haaroon (changes requested):

PR #408, opened 2022-05-30
Streamlining distributed service config

miratepuffin:
Through conversations with @felixcdr who is testing the distributed services of Raphtory several optimisations were proposed. This PR fixes two of these:

## Remove raphtory.deploy.distributed property

This property tells the `ConfigHandler` to add a random salt to the raphtory.deploy.id, in order to avoid conflicts between local deployments and their pulsar channels. However, this option will cause distributed services to not find each other in Pulsar due to the different hashes, while not throwing any useful errors. 

It seems the need for this salt comes exclusively for Raphtory components that are not created through the RaphtoryService or Raphtory client methods. If the code is restructured to explicitly request salted id or not, depending on the initial method calling the creation of the component, this configuration property can be safely removed.

Currently you need to always set the right value for the property, while it is not giving any flexibility.

### Solution

The `Raphtory.getDefaultConfig()` and `Raphtory.confBuilder()` functions now  take a `distributed:Boolean` argument which allows the components calling the function to specify the context within which they are running. As part of this change, I refactored the Prometheus Telemetrics to not use any conf - this is because I can't see any use in having custom name spaces and it was causing nightmares when having a global conf. 

## Replace batchIngestion() from the Service API with a different argument to start the service

Currently this property is set by the class extending the generic RaphtoryService, and if true it will cause Raphtory to instantiate the combo BatchLoadingSpoutBuilderPartitionManager with the `partitionmanager` argument. 

If a service implementation sets this property to true, there is no sense in  starting the service on `spout` or `builder` modes (not intuitive).

The intent however is not clear, it would be more explicit to define an additional option, `batchpartitionmanager` (good suggestions are welcome) that always implements the BLSBPM voltron, whereas `partitionmanager` will only create the partitionmanager. It is clearer this way, and will give more flexibility on how to deploy these services.

### Solution
The `batchIngestion()` function has been removed from the RaphtoryService. This has been replaced with an additional possible argument which specifies that the user wanted a `batchpartitionmanager` -- see below:
image

### DIFF
--- a/core/src/main/resources/application.conf
+++ b/core/src/main/resources/application.conf
@@ -221,14 +221,6 @@ raphtory {
     metrics {
       port = 9999
     }
-    namespaces {
-      spout         = "spout"
-      reader        = "reader"
-      writer        = "writer"
-      query         = "query"
-      storage       = "storage"
-      builder       = "graph_builder"
-    }
   }
 
   python {
--- a/core/src/main/scala/com/raphtory/config/ConfigHandler.scala
+++ b/core/src/main/scala/com/raphtory/config/ConfigHandler.scala
@@ -15,17 +15,15 @@ private[raphtory] class ConfigHandler {
   private lazy val defaults       = createConf()
   private lazy val logger: Logger = Logger(LoggerFactory.getLogger(this.getClass))
 
-  private lazy val deployedDistributed =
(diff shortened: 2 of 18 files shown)


miratepuffin:
Approved by @felixcdr ✅

PR #413, opened 2022-06-03
Dev 943 / Sink API

ricopinazo:


core/src/main/scala/com/raphtory/sinks/StreamSinkConnector.scala
```diff
@@ -0,0 +1,20 @@
+package com.raphtory.sinks
+
+import java.io.Writer
+
+abstract class StreamSinkConnector(itemDelimiter: String) extends SinkConnector {
+
+  private var justClosedItem = false
```
miratepuffin, line 7: Took me a while to work out what was happening with this 😄

core/src/main/scala/com/raphtory/sinks/SinkConnector.scala
```diff
@@ -0,0 +1,27 @@
+package com.raphtory.sinks
+
+/** An abstract sink of entities to be written out
+  *
+  * An item is a piece of information meaningful by itself and isolated from the rest of entities.
```
miratepuffin, line 5: typo

miratepuffin (approved):
Couple of Typo's but I think we are good to go!

PR #421, opened 2022-06-08
Add scaladocs for algorithm API

ljeub-pometry:


ljeub-pometry:
All changes from here are included in #423

PR #422, opened 2022-06-08
Join Json formats and add scaladocs for output classes

ricopinazo:


miratepuffin (dismissed):
A couple of typo's/sentence restructuring that we can do, but I can sort this as part of the scala docs pass over. All looks great otherwise!

PR #424, opened 2022-06-08
Copying all methods to the ConcreteGraphPerspective removes spurious error highlights

ljeub-pometry:
…error highlights


ljeub-pometry:
This is a work-around for an IntelliJ analyser bug that causes spurious type mismatch highlights: https://youtrack.jetbrains.com/issue/SCL-20292/Combining-dependent-types-and-self-type-annotations-causes-spurious-Type-Mismatch-highlights

The duplicate method definitions should be removed again once that bug is fixed!

PR #427, opened 2022-06-09
Fixing the Python Client to work with the latest version

Haaroon:
This updates the python client and associated notebooks to work with the latest version of raphtory 

Add active parameter to the python configuration
Set raphtory.python.active to false by default
simplified python lotr lab to point to the jupyter notebook instead
fixed conn info parameter in client.py
added warning to docs for the new python client variable
updated algorithms tutorial to use the new file output formats
extended FileOutputRunner to run the connected components algorithm
removed unneeded conf.py
removed excess demo file
changed tests to use the cc object


examples/raphtory-example-lotr/src/main/scala/com/raphtory/examples/lotrTopic/FileOutputRunner.scala
```diff
@@ -34,4 +35,13 @@ object FileOutputRunner extends App {
     .writeTo(output)
 
   queryHandler.waitForJob()
+
+  val queryHandlerCC = graph
+    .at(32674)
+    .past()
+    .execute(ConnectedComponents())
```
miratepuffin, line 42: This need to stay as Degree Separation but I can change back in my tutorial check through

miratepuffin (approved):

PR #428, opened 2022-06-10
Dev 981 cats

fabianmurariu:


core/src/main/scala/com/raphtory/internals/management/Scheduler.scala
```diff
@@ -0,0 +1,44 @@
...
+    val (_, cancelable) = IO
+      .parSequenceN(threads)(tasks.to(LazyList))
+      .onError(t => IO(errorHandler(t)))
+      .flatMap(_ => IO(onSuccess))
+      .unsafeToFutureCancelable()
+
+    cancelable
+  }
+
+  def shutdown(): Unit = {}
```
miratepuffin, line 43: Is this just here as a stub at the moment to be filled in? We had the old one to attempt to kill off everything that the scheduler had started, but I imagine this may not be necessary with the planned changes. Could possibly remove as is only called in the GraphDeployment and TemporalGraphConnection - with the latter solely existing to do this function call so the class could probably go as well.

core/src/main/scala/com/raphtory/internals/storage/pojograph/PojoBasedPartition.scala
```diff
@@ -161,7 +162,7 @@ class PojoBasedPartition(partition: Int, conf: Config)
       dstId: Long,
       properties: Properties,
       edgeType: Option[Type]
-  ): Option[GraphUpdateEffect] = {
+  ): Option[RemoteEdgeSync] = {
```
miratepuffin, line 165: Other ones to be changed below as well?

core/src/main/scala/com/raphtory/internals/storage/pojograph/entities/internal/PojoVertex.scala
```diff
@@ -10,34 +10,47 @@ import scala.collection.mutable
 class PojoVertex(msgTime: Long, val vertexId: Long, initialValue: Boolean)
         extends PojoEntity(msgTime, initialValue) {
 
-  var incomingEdges: mutable.Map[Long, PojoEdge] =
+  private val intoE: mutable.Map[Long, PojoEdge] =
```
miratepuffin, line 13: Just a preference for shorter var names? If we are changing, would prefer `inE` as into somewhat sounds like it's something actually inside of an edge.

core/src/test/scala/com/raphtory/lotrtest/Runner.scala
```diff
@@ -0,0 +1 @@
+
```
miratepuffin, line 1: This file just won't die 😄 can you delete please.

miratepuffin (commented):
All looking good so far, the only question is on the Shutdown func for the scheduler. I think you may also be using a different scalafmt to us so we are gonna have to sync on that.

core/src/main/scala/com/raphtory/internals/management/Scheduler.scala
fabianmurariu: yes, I'll take it out

fabianmurariu:
> All looking good so far, the only question is on the Shutdown func for the scheduler. I think you may also be using a different scalafmt to us so we are gonna have to sync on that.

Scheduler has no shutdown function, nor does it need one

core/src/main/scala/com/raphtory/api/input/GraphAlteration.scala
```diff
@@ -1,46 +1,8 @@
 package com.raphtory.api.input
 
-/** Properties are characteristic attributes like name, etc. assigned to Vertices and Edges by the
-  * [[Graph Builder]].
-  * @see [[GraphBuilder]]
-  */
+sealed private[raphtory] trait GraphAlteration
```
miratepuffin, line 3: (Once the merge conflicts are sorted) I think we can now move this class internal? Everything inside is private.

core/src/main/scala/com/raphtory/internals/storage/pojograph/PojoBasedPartition.scala
miratepuffin: A couple more I can point out, unless this is still half way though.

miratepuffin (commented):
Everything looking good, just a comment on the Graph Alterations no longer needing to be in API

Alnaimi- (approved):

PR #429, opened 2022-06-10
Clean up scaladoc links and structure

ljeub-pometry:
- Cleans up the package structure and fixes all links in the scaladocs
- Marks all classes in internals as `private[raphtory]`

Unfortunately, even with `-private` set, it seems that documentation for private classes is not generated, only for private methods in public classes.


miratepuffin (approved):
All looking good, I will just quickly remove the remaining @DoNotDocument

miratepuffin (approved):

PR #450, opened 2022-06-17
Fix examples after refactor

ljeub-pometry:
- `Type` was moved during refactoring but examples were not updated


fabianmurariu (approved):
LGTM

miratepuffin (approved):

PR #452, opened 2022-06-17
Fix PrintSink output being mixed with logs

ricopinazo:


core/src/main/scala/com/raphtory/sinks/PrintSink.scala
```diff
@@ -42,8 +42,28 @@ case class PrintSink(format: Format = CsvFormat()) extends FormatAgnosticSink(fo
...
+      override def closeItem(): Unit = {
+        currentItem.append(itemDelimiter)
+        justClosedItem = true
+      }
+
+      override def close(): Unit = {
+        if (justClosedItem)
+          System.out.print(currentItem)
+        System.out.println()
+      }
```
ljeub-pometry, line 67: I don't think this is useful for debugging. It would be nice to see the output where it actually happens. Also, in this way if we have a stuck perspective, we do not see any output at all?

ljeub-pometry (changes requested):
I think delaying the output until everything is finished is worse than the broken last line...

PR #461, opened 2022-06-23
Fix vertex messaging synchronisation between query executors

ljeub-pometry:
- QueryExecutors now check that they have received all vertex messages before sending GraphFunctionComplete
- QueryHandler checks message counts and fails if they don't match
- moved all remaining tests to MUnit

### DIFF
--- a/core/src/main/resources/application.conf
+++ b/core/src/main/resources/application.conf
@@ -214,6 +214,38 @@ raphtory {
           size                  = ${?RAPHTORY_PULSAR_TOPICS_JOB_OPERATIONS_RETENTION_SIZE}
         }
       }
+      vertex.messages.sync {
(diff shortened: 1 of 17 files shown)


miratepuffin (approved):
Everything seems fine, my only question would be, does it make sense for the Query Handler to still receive send/receive counts when we know they are handled. Most of `processGraphFunctionComplete` seems like it can just be removed. Could have the Query Executors return an error message if they haven't received any new messages for a certain period of time etc. instead?

ljeub-pometry:
I think sanity-checking the message counts is still a good idea however, it would be sufficient to do this only once at the end. We can change this when we further refactor the execution to handle iteration without global state directly between query executors.

miratepuffin (approved):

PR #465, opened 2022-06-28
Initial companies house stream spout and graph builder

rachchan:


miratepuffin (approved):
Great stuff! Looking forward to the algorithms running on top.

PR #487, opened 2022-07-05
TypeDB connector and spout

rachchan:


miratepuffin (approved):
I think we need to generalise the spout to give it a specific query. But we can return to this later.

PR #488, opened 2022-07-06
Feature/counter accumulator

narnolddd:
Counter accumulator implemented for counting vertices/edges per categorical property e.g. number of nodes in each connected component etc. Needs some documentation and comments are welcome on any way of making more efficient or generalisable.

- [x] Documentation
- [ ] Tidy up/refactor of graph state implementation class

### DIFF
--- a/core/src/main/scala/com/raphtory/api/analysis/graphstate/Counter.scala
+++ b/core/src/main/scala/com/raphtory/api/analysis/graphstate/Counter.scala
@@ -2,14 +2,31 @@ package com.raphtory.api.analysis.graphstate
 
 import scala.collection.mutable
 
+/**
+  * Public interface for the Counter API within [[GraphState]].
+  *
+  *  A Counter maintains the distribution of categorical groupings such as community, connected component or node type.
+  *  This enables measures such as the size/relative size of the largest group or largest k groups, filtering to include
+  *  only the largest/largest k groups etc.
+  *
+  *  @tparam T Value type of the category being grouped (usually long if referring to a group or component id)
+  *
+  * @see
+  * [[GraphState]], [[Histogram]]
+  */
+
 abstract class Counter[T] {
 
+  /** Total of the counts */
   def totalCount : Int
 
+  /** Getter function for the map of counts */
   def getCounts : mutable.Map[T,Int]
 
+  /** Returns the id and count of the largest group */
   def largest : (T,Int)
 
+  /** Returns the id and count of the k largest groups */
   def largest(k: Int): List[(T,Int)]
 
 }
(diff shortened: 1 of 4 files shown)


core/src/main/scala/com/raphtory/api/analysis/graphstate/Counter.scala
```diff
@@ -0,0 +1,32 @@
...
+  * [[GraphState]], [[Histogram]]
+  */
+
+abstract class Counter[T] {
+
+  /** Total of the counts */
+  def totalCount : Int
+
+  /** Getter function for the map of counts */
+  def getCounts : mutable.Map[T,Int]
```
fabianmurariu, line 24: I would make this return a MapView to encapsulate the mutable state otherwise this can be updated externally bypassing the synchronized

core/src/main/scala/com/raphtory/api/analysis/graphstate/GraphStateImplementation.scala
```diff
@@ -43,6 +43,60 @@ private object AccumulatorImplementation {
...
 
+private class CounterImplementation[T] () extends Counter[T] {
+  var totalCount : Int = 0
+  var counts: mutable.Map[T, Int] = mutable.Map[T,Int]()
+
+  override def getCounts: mutable.Map[T, Int] = counts
+  override def largest : (T,Int) = counts.maxBy(_._2)
+  override def largest(k: Int): List[(T,Int)] = counts.toList.sortBy(_._2).takeRight(k.min(counts.size))
+
+  def +=(newValue: (T, Int)): Unit = this.synchronized{
```
fabianmurariu, line 54: are you 100% sure this will be used across threads? if yes then perhaps using a [**ConcurrentHashMap.compute**](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#compute-K-java.util.function.BiFunction-) would workout better

core/src/main/scala/com/raphtory/api/analysis/graphstate/GraphStateImplementation.scala
```diff
@@ -43,6 +43,60 @@ private object AccumulatorImplementation {
...
+    }
+  }
+
+}
+
+private object CounterImplementation {
+  def apply[T]() = new CounterImplementation[T]()
+}
+
+private class CounterAccumulatorImplementation[T](retainState:Boolean) extends AccumulatorImplementation[(T, Int),Counter[T]] {
```
fabianmurariu, line 68: Can you please add some unit tests to this class?
covering

- [ ] merge
- [ ] reset
- [ ] +=

core/src/main/scala/com/raphtory/api/analysis/graphstate/GraphStateImplementation.scala
narnolddd: thanks! I'm not totally sure but I'm just following the pattern of the other accumulators there which require it

PR #496, opened 2022-07-12
Dev 1047 python pending

fabianmurariu:
# The purpose of this PR is to add the ability to submit python algorithms and graph loaders via a command line tool (PyRaphtory)
 - [x] Docker image with python support
 - [x] self packaging and self installing python project with poetry
## Local mode
- [x] graph builders
- [x] algorithms 
## Distributed mode
- [x] graph builders
- [x] algorithms

## Note:
1. python notebooks are excluded at this point
2. only a small subset of the Scala API is replicated in python
3. connected components is found in sample.py


bin/docker/raphtory/DockerfileV2
```diff
@@ -0,0 +1,38 @@
+ARG VERSION
+FROM hseeberger/scala-sbt:eclipse-temurin-11.0.14.1_1.6.2_2.13.8 as sbt-builder
+COPY core core/
```
jamesalford, line 3: In the distributed build using the previous docker file, we pass raphtory core code in as a jar. This would be useful as a development tool however we need to make the same changes for python to the other dockerfile while leaving the scala code to get copied in by jar

bin/docker/raphtory/DockerfileV2
```diff
@@ -0,0 +1,38 @@
...
+COPY examples examples/
+COPY project project/
+COPY deploy deploy/
+COPY build.sbt build.sbt
+COPY .scalafmt.conf .scalafmt.conf
+RUN sbt clean core/assembly
+
+FROM eclipse-temurin:11.0.15_10-jdk as python-build
```
jamesalford, line 11: should be an arg for version for easy changes/testing of new versions

fabianmurariu: Yes, I had some trouble setting that up

PR #510, opened 2022-07-28
Feature/DEV-1089 Degree, TwoHopPaths algos and Explode function into Python

Haaroon:
- Added support for Degree
- Renamed sample to connected components 
- Added out_neighbours and message_vertex support to python.scala and vertex.py,
- Wrapped explode select into fn, 
- Added support for the TwoHopPaths algorithm in Python
- Added support for explode function into python


Haaroon:
closed since outdated

PR #523, opened 2022-08-01
Wrapper-based interface for exposing scala classes in python

ljeub-pometry:
This is work-in-progress, so far this only implements a wrapper-based GraphBuilder


ljeub-pometry:
Functional: 
 - calling scala algorithms from python works well
 - building algorithms in python works
 - translation layer is slow and requires more optimisations

core/src/main/scala/com/raphtory/internals/management/PythonEncoder.scala
```diff
@@ -181,10 +197,116 @@ object PythonEncoder {
...
+    }
+
+  def find_class(name: String): Any = {
+    val runtimeMirror = universe.runtimeMirror(getClass.getClassLoader)
+    val module        = runtimeMirror.staticModule(name)
+    val obj           = runtimeMirror.reflectModule(module)
+    obj.instance
+  }
+
+  def make_varargs(obj: Iterable[Any]): Array[Object] =
```
Alnaimi-, line 236: Need some comments or documentary of the method and what is doing

core/src/main/scala/com/raphtory/internals/management/PythonEncoder.scala
```diff
@@ -181,10 +197,116 @@ object PythonEncoder {
...
+            case Some(v) => v.toMap
+            case None    => Map.empty[Int, java.lang.reflect.Method]
+          }
+
+          name -> methods.map { m =>
+            val hasVarArgs  = m.isVarArgs
+            val paramsNames = m.getParameters.map(p => camel_to_snake(p.getName))
+
+            val n = m.getParameterCount
+            if (
```
Alnaimi-, line 269: Needs an explanation of why we are checking if method has non default arguments. Also a point on inplicits.

core/src/main/scala/com/raphtory/internals/management/PythonEncoder.scala
ljeub-pometry: done

core/src/main/scala/com/raphtory/internals/management/PythonEncoder.scala
ljeub-pometry: I've added some comments to explain what this is doing. It may be worth revisiting this method to see if we can use scala reflection instead to simplify things. Implicit arguments do not appear to be special from the java side.

ljeub-pometry:
Closed in favour of #533

PR #534, opened 2022-08-23
3-node temporal motifs and edge state

narnolddd:
Implementing three node temporal motifs algorithm and as part of this implementing edge state, since the only alternative for this algorithm without doing this would be having a vertex state which is a map from a vertex's neighbours to values -- yuck!! Only partway through the process but wanted to share so that we can discuss and someone can say if there are any big red flags 😄


miratepuffin (approved):
Reviewed and fixed in person

PR #545, opened 2022-08-30
Feature/dev 1107 port nft

Haaroon:
porting nft over with new changes

paused until bug fix for case classes


python/pyraphtory/pyraphtory/proxy.py
```diff
@@ -0,0 +1,252 @@
...
+        return self
+
+    def __init_subclass__(cls, **kwargs):
+        """automatically register wrappers that have a '_classname' defined"""
+        super().__init_subclass__(**kwargs)
+        if cls._classname is not None:
+            register(cls)
+
+
+class MethodProxyDescriptor(object):
```
Alnaimi-, line 97: Should all of these methods be part of the commit / PR?

PR #546, opened 2022-08-30
Implementation of KCore

nrs1729:


core/src/main/scala/com/raphtory/algorithms/generic/Coreness.scala
```diff
@@ -0,0 +1,39 @@
...
+
+class Coreness(start: Int, end: Int) extends Generic {
+  // start and end determine the range of k values that KCore runs to determine the coreness
+  final val CORENESS = "CORENESS"
+  final val EFFDEGREE = "effectiveDegree" // todo get dynamically from KCORE?
+
+  // todo add checks
+  override def apply(graph: GraphPerspective): graph.Graph = {
+    var g = graph.step { vertex =>
+      println("hello world")
```
Alnaimi-, line 16: Please remove.

core/src/main/scala/com/raphtory/algorithms/generic/Coreness.scala
```diff
@@ -0,0 +1,39 @@
...
+    }
+    for(k <- start to end) {
+      g = KCore(k, resetStates = false).apply(g).step { vertex =>
+        if (vertex.getStateOrElse[Int](EFFDEGREE, 0) >= k) { // if still in the core graph for this value of k, increment the coreness //todo error handle better
+          vertex.setState(CORENESS, k)
+        }
+      }
+    }
+    g
+    //graph.asInstanceOf[graph.Graph] // TODO type mismatch with just graph: found "graph.type", required: "graph.Graph"
```
Alnaimi-, line 27: Can this be removed?

core/src/main/scala/com/raphtory/algorithms/generic/Coreness.scala
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  override def tabularise(graph: GraphPerspective): Table =
+    graph
+      .select(vertex =>{println("tab")
```
Alnaimi-, line 32: Remove println("tab")

core/src/main/scala/com/raphtory/algorithms/generic/Coreness.scala
```diff
@@ -0,0 +1,39 @@
...
+        Row(vertex.name, vertex.getStateOrElse[Int](CORENESS, -1))}
```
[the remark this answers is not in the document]
nrs1729: Thanks - I couldn't figure out how to do this with getStateOrElse (it needs an Int for the 2nd parameter and will run any code put there unconditionally), think it might need a different function? I've left it as -1 for now but could also just remove the OrElse? (it shouldn't need it)

ljeub-pometry (approved):
This looks good as an implementation with the current features. Once we have resettable filters, we should simplify the KCore algorithm.

PR #548, opened 2022-08-31
Fix distributed deployment definition

ricopinazo:


core/src/main/resources/log4j2.xml
```diff
@@ -31,10 +31,10 @@
         </Logger>
         <Logger name="org.apache" level="${env:RAPHTORY_APACHE_LOG:-off}">
         </Logger>
-        <Logger name="com.typesafe.akka" level="${env:RAPHTORY_AKKA_LOG:-WARN}">
-        </Logger>
-        <Logger name="akka" level="${env:RAPHTORY_AKKA_LOG:-ERROR}">
-        </Logger>
+<!--        <Logger name="com.typesafe.akka" level="${env:RAPHTORY_AKKA_LOG:-INFO}">-->
```
miratepuffin, line 34: Can you turn these back on.

core/src/main/scala/com/raphtory/internals/communication/connectors/AkkaConnector.scala
```diff
@@ -194,6 +191,7 @@ object AkkaConnector {
       akka.remote.artery.canonical.hostname="$hostname"
       akka.remote.artery.canonical.port=$localCanonicalPort
       akka.remote.artery.bind.port=$localBindPort
+      akka.remote.artery.bind.hostname="0.0.0.0"
```
miratepuffin, line 194: need to remove so we can still set it

examples/lotr/src/main/scala/com/raphtory/examples/lotr/TutorialRunner.scala
```diff
@@ -24,7 +24,7 @@ object TutorialRunner extends App {
   val url  = "https://raw.githubusercontent.com/Raphtory/Data/main/lotr.csv"
   FileUtils.curlFile(path, url)
 
-  val graph   = Raphtory.newGraph()
+  val graph   = Raphtory.connect("raphtory").newGraph()
```
miratepuffin, line 27: can just set this back  (or leave) everyone messing with this file at the moment

miratepuffin (changes requested):
Some very small bits to change before merging

miratepuffin:
Important changes for this have been taken int #576

Issue #513, opened 2022-07-29
MapView to encapsulate mutable state in counters


miratepuffin:
I would make this return a MapView to encapsulate the mutable state otherwise this can be updated externally bypassing the synchronized

_Originally posted by @fabianmurariu in https://github.com/Raphtory/Raphtory/pull/488#discussion_r917821217_

PR #551, opened 2022-09-01
map view for Counters 

narnolddd:
so as not to bypass the synchronized, resolving #513


miratepuffin (approved):

miratepuffin:
Great stuff should close #513

PR #552, opened 2022-09-01
Concrete table output

ljeub-pometry:
Sent raw row output to the local machine and collect it

Basic idea works but need to think about the interface


core/src/main/scala/com/raphtory/api/analysis/table/TableOutputTracker.scala
```diff
@@ -0,0 +1,135 @@
...
+      }
+
+  override def next(): TableOutput =
+    if (hasNext) {
+      val out = nextResult.get
+      nextResult = None
+      out
+    }
+    else
+      throw new NoSuchElementException("All perspectives already processed")
```
ljeub-pometry, line 134: Change message for case where output is not done but no more results are available due to timeout

miratepuffin (approved):

PR #554, opened 2022-09-02
proposing a shortest distance path algo

lejohnyjohn6:


core/src/main/scala/com/raphtory/algorithms/generic/distance/ShortestPathDistance.scala
```diff
@@ -0,0 +1,101 @@
...
+    graph
+      .setGlobalState({
+        graphState =>
+          graphState.newMin[T]("minDistance")
+      })
+      .step ( { (vertex, graphState) =>
+        if (vertex.name() == src_name) {
+          vertex.setState(DISTANCE, 0)
+          vertex.getOutEdges().foreach { edge => edge.send(edge.weight[T]())}
+          graphState("minDistance") += MAX[T]
```
ljeub-pometry, line 59: This line is not needed, the min accumulator is already initialised like this by default

ljeub-pometry (changes requested):
The algorithm itself looks good, can you add a test to make sure we don't inadvertently break this in the future?

PR #558, opened 2022-09-05
Removing Graph Builders as user facing class

miratepuffin:


core/src/main/scala/com/raphtory/Raphtory.scala
```diff
@@ -36,7 +38,7 @@ import scala.collection.mutable.ArrayBuffer
   * graph.deployment.stop()
   * }}}
   *
-  * @see [[api.input.GraphBuilder GraphBuilder]]
+  * @see [[GraphBuilder GraphBuilder]]
```
miratepuffin, line 41: Change to graph

core/src/main/scala/com/raphtory/Raphtory.scala
```diff
@@ -7,12 +7,14 @@ import com.raphtory.internals.context.LocalContext
...
 import com.raphtory.internals.context.RemoteContext
 import com.raphtory.internals.context.LocalContext.createName
+import com.raphtory.internals.graph.GraphBuilder
 import com.raphtory.internals.management._
 import com.raphtory.internals.management.id.IDManager
 import com.raphtory.internals.management.id.LocalIDManager
 import com.raphtory.internals.management.id.ZooKeeperCounter
 import com.raphtory.internals.management.id.ZookeeperLimitedPool
 import com.typesafe.config.Config
+import net.openhft.hashing.LongHashFunction
```
miratepuffin, line 17: Remove

miratepuffin (commented):
Make concrete add/remove functions inside graph trait which calls handleUpdate
handle update to be implemented in graph builder and query sending etc.
add abstract graph index

Haaroon:
i have tested the python variant of this and it works, the sample.py works fine, it creates graphs and obtains results

Haaroon (approved):

PR #569, opened 2022-09-08
dynamic class loading

ljeub-pometry:
- Test dynamic class loading (breaks with closed-over case classes)


core/src/main/scala/com/raphtory/api/analysis/table/TableOutputTracker.scala
```diff
@@ -98,26 +100,34 @@ case class TableOutputTracker(tracker: QueryProgressTracker, topics: TopicReposi
   def isJobDone: Boolean =
     outputDone
 
-  private def waitForNextResult(): Any =
+  private def waitForNextResult(): Any = {
+    logger.debug("waiting for result")
```
fabianmurariu, line 104: do we need these?

core/src/main/scala/com/raphtory/api/analysis/table/TableOutputTracker.scala
```diff
@@ -98,26 +100,34 @@ case class TableOutputTracker(tracker: QueryProgressTracker, topics: TopicReposi
...
+      logger.debug("hasNext false as output complete")
       false
+    }
     else
       waitForNextResult() match {
         case Some(res) =>
           res match {
             case EndOutput      =>
               outputDone = true
+              logger.debug("hasNext false as output complete after waiting for result")
```
fabianmurariu, line 126: are the debug statements really required?

core/src/main/scala/com/raphtory/internals/serialisers/DynamicClassLoader.scala
```diff
@@ -7,25 +7,32 @@ import org.slf4j.LoggerFactory
...
-        if (clzBytes != null) {
-          val clz = defineClass(name, clzBytes, 0, clzBytes.length)
-          logger.debug(s"Successfully injected class $clz")
-          clz
-        }
-        else throw t
+        loaded.computeIfAbsent(
+                name,
+                { _ =>
+                  val clzBytes = storage.get(name)
```
fabianmurariu, line 27: we should pull this out in a function

core/src/main/scala/com/raphtory/internals/serialisers/DynamicClassLoader.scala
```diff
@@ -45,9 +52,12 @@ object DynamicClassLoader {
     classStorage.computeIfAbsent(name, _ => clzBytes)
     classLoader.findClass(name)
   }
+  val loader = new DynamicClassLoader(this.getClass.getClassLoader, classStorage)
 
   def lookupClass(name: String): Option[Array[Byte]] = Option(classStorage.get(name))
 
+  private val ref = new AtomicReference[Option[DynamicClassLoader]](None)
```
fabianmurariu, line 59: we don't need this anymore

core/src/main/scala/com/raphtory/internals/serialisers/DynamicClassLoader.scala
```diff
@@ -45,9 +52,12 @@ object DynamicClassLoader {
...
+
   def apply(classLoader: ClassLoader): DynamicClassLoader =
```
fabianmurariu, line 61: classLoader is not used anymore

core/src/main/scala/com/raphtory/internals/serialisers/DynamicLoaderSerializer.scala
```diff
@@ -63,12 +63,13 @@ class DynamicLoaderSerializer(default: Serializer[DynamicLoader]) extends Serial
     // read how many classes are setup for Dynamic loading
     try {
       val n       = input.readInt()
+      val loader  = DynamicClassLoader(kryo.getClassLoader)
```
fabianmurariu, line 66: we can revert this

core/src/main/java/com/raphtory/internals/serialisers/DependencyFinder.java
```diff
@@ -0,0 +1,184 @@
...
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+import java.util.BitSet;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+
+// https://stackoverflow.com/questions/19862866/find-java-class-dependencies-at-runtime
```
fabianmurariu, line 13: This code is quite annoying but I just cannot come up with an alternative right now.

https://web.archive.org/web/20151231173739/http://asm.ow2.org/doc/tutorial-asm-2.0.html
this is an old example that shows how to use ASM visitors to lift all the dependencies of a class, I think it's worth having a look

core/src/main/scala/com/raphtory/internals/serialisers/DynamicLoaderSerializer.scala
```diff
@@ -63,12 +63,13 @@ class DynamicLoaderSerializer(default: Serializer[DynamicLoader]) extends Serial
...
       val classes = (0 until n).map { _ =>
         val name   = input.readString()
         val length = input.readInt()
         val bytes  = input.readBytes(length)
-        DynamicClassLoader.injectClass(name, bytes, DynamicClassLoader(kryo.getClassLoader))
-      }.toSet
+        DynamicClassLoader.injectClass(name, bytes, loader)
+      }.toList
```
fabianmurariu, line 72: I think we need a distinct here

ljeub-pometry: Adding distinct here is way too late. This is being handled when the DynamicLoader is resolved before sending it. I added a check in the write method such that it fails if we ever forget to do that.

examples/lotr/src/test/scala/DynamicClassLoaderTest.scala
```diff
@@ -0,0 +1,161 @@
...
+  private val compiledAlgo: Generic = tb.compile(tb.parse(minimalTestCode))().asInstanceOf[Generic]
+
+  val remoteGraphResource =
+    Resource.make(IO(remote().newGraph()))(graph => IO(graph.destroy())).evalMap(g => IO(g.load(source())))
+
+  val remoteProcess: Resource[IO, Process] = Resource.make {
+    IO {
+      println("starting remote process")
+      Process(
+              Seq(
```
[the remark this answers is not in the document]
ljeub-pometry: I managed to do this only using java so there is no risk of triggering compilation.

core/src/main/scala/com/raphtory/internals/serialisers/DynamicLoaderSerializer.scala
ljeub-pometry: This is not needed any more at all as there is only one DynamicClassLoader so it can all be handled in the object.

core/src/main/scala/com/raphtory/internals/serialisers/DynamicClassLoader.scala
ljeub-pometry: I have simplified this now

core/src/main/scala/com/raphtory/internals/serialisers/DynamicClassLoader.scala
ljeub-pometry: It is gone

core/src/main/scala/com/raphtory/internals/serialisers/DynamicClassLoader.scala
ljeub-pometry: It is done. It required currying :)

core/src/main/scala/com/raphtory/api/analysis/table/TableOutputTracker.scala
ljeub-pometry: No debug statements are ever required...

As we currently are having some hanging issues when things go wrong, these are quite useful for knowing where we are...

core/src/main/java/com/raphtory/internals/serialisers/DependencyFinder.java
ljeub-pometry: Well, I had a look at the alternative and it misses key things. I could probably re-implement this parser in scala but not sure it is a good use of my time as so far it seems to work pretty well...

miratepuffin (approved):

PR #576, opened 2022-09-13
Dev-1142 / gRPC Middleware

ricopinazo:


core/src/main/scala/com/raphtory/internals/context/RemoteContext.scala
```diff
@@ -31,8 +33,9 @@ class RemoteContext(deploymentID: String) extends RaphtoryContext {
     for {
       _               <- Py4JServer.fromEntryPoint[IO](this, config)
       _               <- Prometheus[IO](prometheusPort)
-      topicRepo       <- DistributedTopicRepository[IO](AkkaConnector.ClientMode, config)
```
ricopinazo, line 34: The TopicRepository on the client side is only used for local communication now

core/src/main/scala/com/raphtory/internals/management/QuerySender.scala
```diff
@@ -39,7 +39,8 @@ private[raphtory] class QuerySender(
   val totalPartitions: Int      = partitionServers * partitionsPerServer
   private lazy val writers      = topics.graphUpdates(graphID).endPoint
   private lazy val queryManager = topics.blockingIngestion.endPoint
-  private lazy val submissions  = topics.submissions.endPoint
+  private lazy val graphSetup   = topics.graphSetup.endPoint
```
ricopinazo, line 42: We build the endPoint for the graphSetup topic just once

miratepuffin (approved):
Have been testing this with Pedro and all seems good!

PR #577, opened 2022-09-13
Feature/dev 1190 Pyraphtory downloads a jar

Haaroon:
- Pyraphtory will now download a raphtory jar from github, upon first run, if a jar is not present. 
- it will also check the current jars in the folder, if it cant find a raphtory one it will download it
- it will also check the versions of the raphtory ones in the folder, if the versions dont match it will delete the incompatible version and download a new one
- unified python versions, so now it checks the .toml file for the version number going forward


python/pyraphtory/pyraphtory/context.py
```diff
@@ -101,6 +106,11 @@ def __init__(self, logging: bool = False):
         jars = ":".join([str(jar) for jar in jar_location.glob(env_jar_glob_lookup)])
         java_args = os.environ.get("PYRAPTHORY_JVM_ARGS", "")
 
+        # if jars is empty, then download it
+        if not jars:
+            jars = fileutils.download_raphtory(VERSION,
+                                               str(Path(inspect.getfile(self.__class__)).parent.parent / 'lib'))
+
```
ljeub-pometry, line 113: - Should we prompt the user before downloading the jar?
- Also, if the PYRAPHTORY_JAR_LOCATION was set, we probably should be raising an error if the jar is not   found rather than trying to download it?
- It does not seem like we check the version of the jar? Seems like we should check that the version is consistent by default? Also, does this mean that we would keep using the old jar if the python version is updated?

python/pyraphtory/pyraphtory/context.py
```diff
@@ -6,14 +6,18 @@
 from threading import Thread
 
 import pandas as pd
+import sys
 from abc import abstractmethod
 from pathlib import Path
 from subprocess import PIPE, Popen
+from threading import Thread
```
ljeub-pometry, line 13: Duplicate

python/pyraphtory/pyraphtory/context.py
```diff
@@ -6,14 +6,18 @@
 from threading import Thread
 
 import pandas as pd
+import sys
```
ljeub-pometry, line 9: Duplicate

python/pyraphtory/pyraphtory/context.py
```diff
@@ -6,14 +6,18 @@
...
 from pathlib import Path
 from subprocess import PIPE, Popen
+from threading import Thread
 from weakref import finalize
 from typing import IO, AnyStr
+import pyraphtory.fileutils as fileutils
 
 from py4j.java_gateway import JavaGateway, GatewayParameters
 from py4j.protocol import Py4JJavaError
+from typing import IO, AnyStr
```
ljeub-pometry, line 20: Duplicate

python/pyraphtory/pyraphtory/context.py
ljeub-pometry: Also noticed there is a bug on line 79-80 (this one is my fault I guess). These two lines should not be included in the `if logging` bit (i.e., the `logging` toggle should only disable the writing of the output, we still need the iterator to actually return the value to find the connection details. Can you fix this or should I do it?

ljeub-pometry (dismissed):
- This looks largely fine and the download code looks good. 
- There are a bunch of duplicate imports which should be removed
- I have some questions about version management for the jar (we currently don't seem to check versions at all)

python/pyraphtory/pyraphtory/fileutils.py
```diff
@@ -0,0 +1,77 @@
...
+import shutil
+import os.path
+import hashlib
+
+_URL = 'url'
+_SHA256 = 'sha256'
+
+RAPHTORY_JARS_HASH = {
+    '0.2.0a0': {
+        _URL: 'https://github.com/Raphtory/Raphtory/releases/download/v0.2.0a0/core-0.2.0a0.jar',
```
fabianmurariu, line 12: Have a look at https://stackoverflow.com/questions/17341122/link-and-execute-external-javascript-file-hosted-on-github

it shows a better way to link github files via a CDN

python/pyraphtory/pyraphtory/context.py
```diff
@@ -23,6 +27,7 @@
 from pyraphtory import interop
 import sys
 
+VERSION = '0.2.0a0'
```
fabianmurariu, line 30: versioning should be external? like from an env variable or from a file?

Haaroon: can you recommend anything? I have a version file in the base package for raphtory, but that is inaccessible in python. I can do the same, where we have a version file in the python root dir?

Haaroon: okay now the python script gets the version from the toml file. note that toml is static and therefore it cannot load its own version .

Haaroon (commented):
doen changes

Haaroon:
i have tested this by hand, 
- with a normal raphtory jar thats correct
- wth a raphtory jar that has an incorrect version number
- with no raphtory jar

miratepuffin (approved):

PR #595, opened 2022-09-18
Removing client zookeeper dependency

miratepuffin:


examples/lotr/src/main/scala/com/raphtory/examples/lotr/TutorialRunner.scala
```diff
@@ -55,7 +55,7 @@ object TutorialRunner extends App {
 
   //Connected Components
   graph
-    .at(32674)
+    .range("1/1/2020", "1/1/2021", "1 month 1 week and 1 day")
```
ricopinazo, line 58: Is this working? the range of those dates in milliseconds is probably outside lotr timeline

core/src/main/scala/com/raphtory/internals/components/cluster/OrchestratorComponent.scala
```diff
@@ -44,6 +44,28 @@ abstract class OrchestratorComponent(conf: Config) extends Component[ClusterMana
       }
     }
 
+  protected def establishPartition(
+      component: String,
+      graphID: String,
+      clientID: String,
+      IDManager: IDManager,
```
ricopinazo, line 51: I guess that this works, but maybe we could start using lowercase for acronyms

core/src/main/scala/com/raphtory/internals/components/cluster/OrchestratorComponent.scala
```diff
@@ -44,6 +44,28 @@ abstract class OrchestratorComponent(conf: Config) extends Component[ClusterMana
...
+      func: (String, String, Config, IDManager) => Unit
```
ricopinazo, line 52: Does `func` need to be abstract. Not sure, but I don't see any scenario where we might want to call something different from `deployPartitionService`

PR #601, opened 2022-09-23
Python docstrings

ljeub-pometry:
- Scala docstrings are available from python objects 
- Some automatic translation of scala to python syntax
- Running pytest in CI (more tests to be added)


ljeub-pometry:
Method resolution order for python wrappers is currently broken, do not merge yet

ljeub-pometry:
Ok, seems to work now

python/pyraphtory/poetry.lock
```diff
@@ -0,0 +1,401 @@
+[[package]]
```
miratepuffin, line 1: This can be removed can't it?

python/pyraphtory/tests/test_pyraphtory.py
```diff
@@ -2,4 +2,4 @@
 
 
 def test_version():
```
miratepuffin, line 4: We have a version file now, can be read into here?

miratepuffin (changes requested):

python/pyraphtory/poetry.lock
ljeub-pometry: Poetry docs suggest to commit it so everyone ends up with the same packages, not sure it is that helpful.

python/pyraphtory/tests/test_pyraphtory.py
ljeub-pometry: That is a good idea, let me try

PR #606, opened 2022-09-27
Bug/duplicate querries

ljeub-pometry:
Fixes bug in Table.get() which was causing duplicate queries


test/src/test/scala/test/raphtory/DynamicClassLoaderTest.scala
```diff
@@ -211,6 +211,7 @@ class DynamicClassLoaderTest extends CatsEffectSuite {
 
   remoteGraphInline.test("test inline graphbuilder definition") { g =>
     val res = g.execute(EdgeList()).get().toList
+    println(res)
```
miratepuffin, line 214: Can you remove this and I can merge :)

PR #615, opened 2022-09-30
Fix TableOutputTracker having potentially inconsistent number of partitions

ljeub-pometry:

Send number of partitions in output tracking messages to avoid hang in case where local and remote config are different

Remote connection hangs

No

integration tests

Better way to sync number of partitions?


miratepuffin (approved):
Watched it get fixed live 🍿

PR #617, opened 2022-10-03
Fix flaky DynamicClassLoader test due to standalone process not ready

ljeub-pometry:

Fixes the startup sequence in the DynamicClassLoader test such that the Standalone instance is guaranteed to be running before trying to connect to it.

DynamicGraphLoader test was flaky

No

Automatic tests

`Raphtory.connect()` currently always succeeds, even if there is no remote instance, making a retry approach impossible.

### DIFF
--- a/core/src/main/scala/com/raphtory/internals/components/cluster/ClusterManager.scala
+++ b/core/src/main/scala/com/raphtory/internals/components/cluster/ClusterManager.scala
@@ -23,7 +23,7 @@ class ClusterManager(
   private lazy val cluster = topics.clusterComms.endPoint
 
   override private[raphtory] def run(): Unit =
-    logger.info(s"Starting HeadNode")
+    println(s"HeadNode started")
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/com/raphtory/internals/components/cluster/ClusterManager.scala
```diff
@@ -23,7 +23,7 @@ class ClusterManager(
   private lazy val cluster = topics.clusterComms.endPoint
 
   override private[raphtory] def run(): Unit =
-    logger.info(s"Starting HeadNode")
+    println(s"HeadNode started")
```
miratepuffin, line 26: Any reason this was changed from logger to println?

miratepuffin: Ok nevermind, I see below

miratepuffin (approved):
Can change the println back to a logger once we fix the connection issue

PR #618, opened 2022-10-04
fix vertex messaging

ljeub-pometry:

Fixes vertex messaging from python after the arrow flight changes

Vertex messaging in python was failing due to missing implicit arguments

The user facing Python API is unchanged

python sample script runs again without errors

Actual implicit resolution to make arrow flight work still needs to be implemented

### DIFF
--- a/core/src/main/scala/com/raphtory/api/analysis/visitor/Vertex.scala
+++ b/core/src/main/scala/com/raphtory/api/analysis/visitor/Vertex.scala
@@ -3,6 +3,7 @@ package com.raphtory.api.analysis.visitor
 import com.raphtory.api.analysis.visitor
 import EdgeDirection.Direction
 import PropertyMergeStrategy.PropertyMerge
+import com.raphtory.internals.communication.SchemaProviderInstances
 import com.raphtory.internals.components.querymanager.SchemaProvider
 
 import scala.collection.View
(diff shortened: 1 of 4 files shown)


Alnaimi- (dismissed):

core/src/main/scala/com/raphtory/api/analysis/visitor/Vertex.scala
```diff
@@ -58,31 +59,31 @@ trait Vertex extends EntityVisitor {
...
-  def messageVertex[T: ClassTag](vertexId: IDType, data: T)(implicit provider: SchemaProvider[T]): Unit
+  def messageVertex[T](vertexId: IDType, data: T)(implicit provider: SchemaProvider[T]): Unit
 
   /** Send the same message data to all out-neighbours of this vertex
     * @param message message data to sent
     */
-  def messageOutNeighbours[T:ClassTag](message: T)(implicit provider: SchemaProvider[T]): Unit =
-    outNeighbours.foreach(messageVertex(_, message))
+  def messageOutNeighbours[T](message: T)(implicit provider: SchemaProvider[T]): Unit =
+    outNeighbours.foreach(messageVertex(_, message)(provider))
```
fabianmurariu, line 74: you don't need to explicitly pass (provider) here

iamsmkr (approved):

miratepuffin (approved):
Great stuff! Will merge in now

PR #619, opened 2022-10-04
fix messy module versions

ljeub-pometry:

Consistent version numbers for examples and other packages

Each package was specifying its own version number which were inconsistent

No

Changing version number in version and running publishLocal now produces sane results

No


miratepuffin (approved):
This PR is going to make @jamesalford very happy :D

PR #622, opened 2022-10-05
Merged in the arrow-core package.

jatindersangha:
Releasing the Apache Arrow storage model for Raphtory.

Without the core arrow code we cannot move away from the Pojo storage model currently in use.

No is fully internal 

It has a small amount of its own tests, but is being utilised in #621 via a local build.

This is not used at all at the moment, will be wired in as part of Raphtory 0.3.0


fabianmurariu (approved):
LGTM

PR #624, opened 2022-10-06
DEV 1213 / RaphtoryService

ricopinazo:
This adds a new top-level RaphtoryService interface that can be used seamlessly between local and remote deployments to interact with the service from the client side. In the remote scenario, it uses gRPC in order to work. This eliminates the need for a gRPC middleware between client and server, which was only a quick fix to solve the issues of Akka in remote mode.

This allows us to start progressively reimplementing our internal services as functional interfaces based on gRPC.

No

As cluster mode might be potentially affected by these changes, I manually tested a cluster using the TutorialRunner and it worked.

No


miratepuffin (approved):
Notes from call:

- A lot of the case classes/messages can be reviewed for removal and/or swapped into proto.
- Once output/query tracker is merged and sorted we can remove akka on the client side
- Need to add a health check for the server - is a useful features but also to remove the println check in the dynamic loader test.
- Need to build a singleton local context which has the same interfaces as the remote context (bringing local and distributed deployments even closer together) and review how everything is closed.

PR #625, opened 2022-10-07
DEV-1214 and DEV-1234 / Ingestion service and healthcheck

ricopinazo:
This PR reimplements the Ingestion service udner a functional interface and adds a basic implementation of a ServiceRepository with service discovery capabilities (based on Zookeeper in cluster mode)

Additionally, because the changes are minimal, it also includes the health check service referred by DEV-1234

We need the functional interface for the ingestion service to continue the port of the platform from Akka to gRPC. The basic implementation of the ServiceRepository is needed for the ingestion service to be found by the head node in cluster mode.

The addition of the health check allows us to know externally if the raphtory service started using the utility grpc-health-probe. Which can be useful as well for Docker and Kubernetes deployments.

No

Yes. I manually tested the cluster deployment as there are not currently tests for that.

No


core/src/main/scala/com/raphtory/internals/components/ingestion/IngestionExecutor.scala
```diff
@@ -14,55 +15,63 @@ import com.raphtory.internals.components.querymanager.NonBlocking
...
-    executeSpout()
-  }
-
-  override def handleMessage(msg: Any): Unit = {} //No messages received by this component
-
-  private def executeSpout(): Unit = {
+  def release(): F[Unit] =
+    for {
+//      _ <- Async[F].delay(close()) -> Needed if this class extends FlushToFlight
+      _ <- Async[F].delay(writers.values.foreach(_.close()))
```
fabianmurariu, line 50: are any of the calls wrapped by `delay` blocking?
if so we should use `blocking`

core/src/main/scala/com/raphtory/internals/components/querymanager/QueryManager.scala
```diff
@@ -93,8 +93,8 @@ private[raphtory] class QueryManager(
 
   override def handleMessage(msg: QueryManagement): Unit =
     msg match {
-      case ingestData: IngestData       =>
-        ingestion sendAsync ingestData
+//      case ingestData: IngestData       => TODO: turn back on this?
```
fabianmurariu, line 96: do we turn this back on?

core/src/main/scala/com/raphtory/internals/components/repositories/LocalServiceRepository.scala
```diff
@@ -0,0 +1,30 @@
...
+class LocalServiceRepository[F[_]: Async](topics: TopicRepository, services: Ref[F, Map[String, Any]])
+        extends ServiceRepository[F](topics) {
+
+  override def registered[T](instance: T, descriptor: ServiceDescriptor[F, T]): Resource[F, Unit] =
+    Resource.make(services.update(services => services + (descriptor.name -> instance)))(_ =>
+      services.update(services => services - descriptor.name) // TODO: this doesn't work for multiple instances
+    )
+
+  override def getService[T](descriptor: ServiceDescriptor[F, T]): Resource[F, T] =
+    Resource.eval(services.get.map(serviceList => serviceList(descriptor.name).asInstanceOf[T])) // TODO: this can fail
```
fabianmurariu, line 20: no `TODO` returning a wrong instance of a service is a total failure and will be propagated via `F[_]`

fabianmurariu: By that I mean it works as expected

fabianmurariu (approved):
LGTM 
BUT!
My biggest issue with this is that we added quite a bit of code without any unit tests.
I'd like us to getting used to adding smaller tests
For example the Ingestion executor is a prime example of simple logic that could be tested.
At the same time the Graph management is also testable

core/src/main/scala/com/raphtory/internals/components/ingestion/IngestionExecutor.scala
ricopinazo: The `close` implementation of the `AkkaConnector` is indeed empty. I'll bear that in mind though when reimplementing the writer interface

PR #627, opened 2022-10-12
Dev 1235 context

iamsmkr:
- Merge local/remote contexts into RaphtoryContext
- Implemented ConfiguBuilder as one stop shop for configurations

- Simple interface for Contexts
- Single source of truth for configurations

No. Yes, will need documentation changes.

Manual testing

`RaphtoryContext.getGraph(graphID: String): DeployedTemporalGraph` needs to request graphs from the head node instead.


core/src/main/scala/com/raphtory/internals/context/RaphtoryContext.scala
```diff
@@ -1,48 +1,45 @@
...
+    val qs = for {
+      _          <- Prometheus[IO](config.getInt("raphtory.prometheus.metrics.port"))
+      service    <- serviceResource
+      topicRepo  <- LocalTopicRepository[IO](config, None)
+      querySender = new QuerySender(graphID, service, new Scheduler(), topicRepo, config, createName)
+      _          <- Resource.eval(IO(querySender.establishGraph()))
+    } yield querySender
+
+    val (querySender, shutdown) = qs.allocated.unsafeRunSync()
+    new DeployedTemporalGraph(Query(graphID = graphID), querySender, config, local, shutdown)
```
ricopinazo, line 31: I think we are not using the `local` parameter anymore. Maybe we can remove it.

core/src/main/scala/com/raphtory/package.scala
```diff
@@ -0,0 +1,34 @@
...
+import com.raphtory.internals.management.id.LocalIDManager
+import com.raphtory.internals.management.id.ZooKeeperCounter
+import com.raphtory.internals.management.id.ZookeeperLimitedPool
+import com.typesafe.config.Config
+
+package object raphtory {
+
+  private[raphtory] def createName: String = Nomen.est().adjective().color().animal().get()
+
+  private[raphtory] val defaultConf: Config = ConfigBuilder().build().getConfig
```
ricopinazo, line 16: We could use this method in all the places we have `ConfigBuilder().build().getConfig`

core/src/main/scala/com/raphtory/api/analysis/graphview/DeployedTemporalGraph.scala
```diff
@@ -30,14 +30,12 @@ class DeployedTemporalGraph private[raphtory] (
...
-    if (!local)
-      querySender.destroyGraph(force)
+    querySender.destroyGraph(force)
     shutdown.unsafeRunSync()
   }
 
   override def close(): Unit = {
-    if (!local)
-      querySender.disconnect()
+    querySender.disconnect()
```
ricopinazo, line 38: We need to remember to make this `close` idempotent

core/src/main/scala/com/raphtory/package.scala
iamsmkr: - Internally, yes! But there seems no way to constraints this. 
- From client projects `ConfigBuilder` is available to users to be able create fresh config instance.

core/src/main/scala/com/raphtory/internals/context/RaphtoryContext.scala
iamsmkr: @miratepuffin Could you please confirm?

core/src/main/scala/com/raphtory/api/analysis/graphview/DeployedTemporalGraph.scala
iamsmkr: Probably will add a comment for this PR. Will certainly consider this as part of next PR. Adding this to: **Are there any further changes required?**

iamsmkr: I just checked the `OrchestratorComponent`, it logs **WARN** if graph is not found when destroying and disconnecting graphs.

```
logger.warn(s"Graph '$graphID' requested for destruction by '$clientID', but did not exist")
logger.warn(s"'$clientID' disconnected from Graph '$graphID', but did not exist")
```

Which makes these APIs idempotent!

core/src/main/scala/com/raphtory/internals/context/RaphtoryContext.scala
miratepuffin: yeah I believe that is the case, the only place I think it was being used was the different behaviour for local and remote destroy/closure which obviously all changes with this.
image

iamsmkr:
Created a new PR instead https://github.com/Raphtory/Raphtory/pull/630

PR #629, opened 2022-10-18
JSON source

rachchan:
JSON sources and networkX json source

so that user can easily ingest and build a graph in Raphtory from network X json data or other json data

no

Written new unit tests

nope


miratepuffin (changes requested):

core/src/main/scala/com/raphtory/api/input/sources/JSONSource.scala
```diff
@@ -0,0 +1,70 @@
...
+ * @param edgeKey : state the key of the edge type/relationship in your JSON data (default = "Person to Person")
+ * @param timeKey : state the key of the timestamp in your JSON data (default = 1)
+ */
+
+class JSONSource(override val spout: Spout[String], sourceKey: String = "source", sourceType: String = "Name", targetKey: String = "target", targetType: String = "Name", edgeKey: Option[String] = None, timeKey: Option[String] = None) extends Source {
+  override type MessageType = String
+
+  private var source: String = _
+  private var target: String = _
+  private var _type: Option[String] = None
```
Alnaimi-, line 32: Why the _ notation? The default Scala guideline is to escape with back-ticks: `type`

project/Dependencies.scala
```diff
@@ -91,6 +94,7 @@ object Dependencies {
 
   lazy val twitterChill   = "com.twitter"         %% "chill"     % chillVersion excludeAll (excludeLog4j, excludeSlf4j)
   lazy val typesafeConfig = "com.typesafe"         % "config"    % typesafeConfigVersion excludeAll (excludeLog4j, excludeSlf4j)
+  lazy val ujson = "com.lihaoyi"        %% "upickle"  % ujsonVersion excludeAll (excludeLog4j, excludeSlf4j)
```
Alnaimi-, line 97: Keeping alphabetical order 😍

core/src/main/scala/com/raphtory/api/input/sources/JSONSource.scala
rachchan: sorry MY BAD!!

miratepuffin (approved):

PR #630, opened 2022-10-18
Newgraphapis

iamsmkr:
New graph API with managed resources and simplified impl of local/remote contexts 

Managing resources was an issue and the contexts code was tricky

Yes, requires document changes

integration tests

No


core/src/main/scala/com/raphtory/RaphtoryApp.scala
```diff
@@ -0,0 +1,38 @@
...
+
+import cats.effect._
+import cats.effect.unsafe.implicits.global
+import com.raphtory.internals.components.RaphtoryServiceBuilder
+import com.raphtory.internals.context.RaphtoryContext
+import com.raphtory.internals.management.GraphConfig.ConfigBuilder
+
+trait RaphtoryApp {
+
+  sealed trait RaphtoryContextType
```
ricopinazo, line 11: Instead of using this to differentiate the two cases, we could have an object RaphtoryApp containing two traits: Local and Remote. This way we would write MyApp extends RaphtoryApp.Local. Only for the remote one, we could provide methods to override the address. This might be address in a future PR though.

core/src/main/scala/com/raphtory/api/analysis/graphview/DeployedTemporalGraph.scala
```diff
@@ -3,41 +3,35 @@ package com.raphtory.api.analysis.graphview
...
   def config: Config = conf
 
   def destroy(force: Boolean = false): Unit = {
-    if (!local)
-      querySender.destroyGraph(force)
+    querySender.destroyGraph(force)
     shutdown.unsafeRunSync()
   }
 
   override def close(): Unit = {
```
ricopinazo, line 33: We shouldn't expose a `close` method to the user anymore. We need to ensure that we are calling `querySender.disconnect()` at the end of `runWithgraph`. As for the `shutdown` hook, this is set to `IO.unit` in all the cases now, so we can remote it and stop using it inside `destroy`

core/src/main/scala/com/raphtory/RaphtoryApp.scala
iamsmkr: I didn't like ContextTypes anyway and this approach seems like a natural fit for our use case since we are anyway allowing only one context per app. Already implemented in this PR itself.

core/src/main/scala/com/raphtory/api/analysis/graphview/DeployedTemporalGraph.scala
iamsmkr: Fixed

connectors/twitter/src/main/scala/com/raphtory/twitter/spout/LiveTwitterSpout.scala
```diff
@@ -51,7 +51,7 @@ class LiveTwitterSpoutInstance() extends SpoutInstance[Tweet] {
 
 class LiveTwitterAddSpout(tweetQueue: ConcurrentLinkedQueue[Tweet]) {
   private val logger: Logger         = Logger(LoggerFactory.getLogger(this.getClass))
-  private val raphtoryConfig: Config = Raphtory.getDefaultConfig()
+  private val raphtoryConfig: Config = ConfigBuilder().build().getConfig
```
fabianmurariu, line 54: Why don't we have a `getDefaultConfig` on ConfigBuilder

core/src/main/scala/com/raphtory/internals/context/RaphtoryContext.scala
```diff
@@ -1,48 +1,100 @@
...
+import com.raphtory.internals.context.GraphException._
 
-import scala.collection.mutable
+class RaphtoryContext(serviceAsResource: Resource[IO, RaphtoryService[IO]], config: Config) {
+  protected val logger: Logger = Logger(LoggerFactory.getLogger(this.getClass))
 
-abstract class RaphtoryContext {
+  // With this API user is trying to connect to an existing graph with following expectations:
+  // - if the graph is not available with the provided graph id, throw an exception otherwise fetch the graph to work with
+  def runWithGraph[T](graphID: String, destroy: Boolean = false)(f: DeployedTemporalGraph => T): T = {
```
fabianmurariu, line 24: these two functions have a lot in common, I'd prefer an idempotent operation callig `runWithGraph` but behave like `runWithNewGraphOrCreate` instead of failing if the graph is not there we create it
we can add params if we absolutely must fail

core/src/main/scala/com/raphtory/service/ClusterManagerService.scala
```diff
@@ -4,19 +4,14 @@ import cats.effect.ExitCode
...
-import com.raphtory.internals.management.arrow.ZKHostAddressProvider
+import com.raphtory.internals.management.GraphConfig.ConfigBuilder
 
 object ClusterManagerService extends IOApp {
 
   def run(args: List[String]): IO[ExitCode] =
     (for {
-      config  <- Resource.pure(Raphtory.getDefaultConfig())
+      config  <- Resource.pure(ConfigBuilder().build().getConfig)
       service <- RaphtoryServiceBuilder.manager[IO](config)
```
fabianmurariu, line 15: you don't need this, just make a `val` at the start of the `run` function and pass it down to `manager[IO](config)`

core/src/main/scala/com/raphtory/service/Standalone.scala
```diff
@@ -4,35 +4,14 @@ import cats.effect.ExitCode
...
-import higherkindness.mu.rpc.server.GrpcServer
-import org.apache.arrow.memory.RootAllocator
+import com.raphtory.internals.management.GraphConfig.ConfigBuilder
 
 object Standalone extends IOApp {
 
   def run(args: List[String]): IO[ExitCode] =
     (for {
-      config  <- Resource.pure(Raphtory.getDefaultConfig())
+      config  <- Resource.pure(ConfigBuilder().build().getConfig)
```
fabianmurariu, line 14: you don't need this, just make a val at the start of the run function and pass it down to manager[IO](config)

core/src/test/scala/com/raphtory/api/analysis/table/TestTableOutput.scala
```diff
@@ -1,37 +1,48 @@
...
-          } yield g
-  )
-  override def munitFixtures                          = List(suiteGraph)
+  def run[R](f: DeployedTemporalGraph => R): R =
+    RaphtoryServiceBuilder
+      .standalone[IO](defaultConf)
+      .use { standalone =>
+        IO {
+          val ctx = new RaphtoryContext(Resource.eval(IO(standalone)), defaultConf)
+          ctx.runWithNewGraph() { graph =>
```
fabianmurariu, line 22: can we have a runWithNewGraphIO version that doesn't call unsafeRunSync to be used inside IO contexts

core/src/main/scala/com/raphtory/internals/context/RaphtoryContext.scala
iamsmkr: We have already visited this with @miratepuffin. We don't want creating a new graph as default behaviour.

fabianmurariu (dismissed):
LGTM
I would have preferred we keep the API for testing so we don't change as many tests

core/src/test/scala/com/raphtory/algorithms/SquareCountTest.scala
```diff
@@ -5,40 +5,42 @@ import com.raphtory.TestQuery
 import com.raphtory.algorithms.generic.motif.SquareCount
 import com.raphtory.api.input.Source
 import com.raphtory.api.input.sources.CSVEdgeListSource
-import com.raphtory.spouts.ResourceSpout
 
 class SquareCountTest extends BaseCorrectnessTest {
-  test("test PQ square") {
+  withGraph.test("test PQ square") { graph =>
```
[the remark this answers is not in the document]
iamsmkr: We were creating new graph already!

miratepuffin (approved):

miratepuffin:
Reviewed with @iamsmkr in person :)

PR #636, opened 2022-10-24
Fix for Disparity Filter

aliyassin4:


miratepuffin:
@aliyassin4 Closing this as we will soon have a much nicer API to work with. Will ping you when it is live so we can get it implemented!

PR #640, opened 2022-10-26
Feature/io context

ljeub-pometry:
- Added a RaphtoryIOContext object which takes care of creating Context resources for use by tests and RaphtoryApp
- Cleaned up the tests using the new context resources

- Code for creating `RaphtoryContext` was duplicated in many places

At the moment this class is internal though we may consider making it user-facing

The tests still pass

Maybe make these methods part of the public API?


fabianmurariu (approved):
LGTM

miratepuffin (approved):

PR #641, opened 2022-10-26
Feature/dev 1276 Python JRE Installer

Haaroon:

- A JRE Installer that will install java, ivy and pyraphtory from two commands
- Removal of the Python raphtory downloader
- Bundling lightweight raphtory jars with pyraphtory
- changes to the build script 

- to improve the user experience when installing raphtory, now you don't need to download java. This two commands will do everything from python

- yes, two new commands are needed for install, need to document 

- You can test this by running the following code

```
# install to clean docker that has no java but only python
docker run --name ubuntu --rm -i -t python:3.9 bash
pip install -vv -i https://test.pypi.org/simple/ pyraphtory-jvm==2.0.a5
pip install -i https://test.pypi.org/simple/ pyraphtory==0.2.0a5

## Test
pip install ipython

## run, this should make a pyraphtory graph in your container
> ipython
from pathlib import Path
from pyraphtory.context import PyRaphtory
from pyraphtory.vertex import Vertex
from pyraphtory.spouts import FileSpout
from pyraphtory.builder import *
ctx = PyRaphtory.local()
graph = ctx.new_graph()
```

- no


Haaroon:
TODO

- Write setup docs on how to build this 
- Ensure make file builds the latest pyraphtory_jvm files correctly
- Update releases and versions to the latest commit
- Update docs so it now uses these

python/pyraphtory_jvm/pyraphtory_jvm/ivy.xml
```diff
@@ -0,0 +1,223 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
fabianmurariu, line 1: why isn't this generated every time?

python/pyraphtory_jvm/setup.py
```diff
@@ -0,0 +1,59 @@
...
+        print("Build_wheel")
+        import pyraphtory_jvm
+        pyraphtory_jvm.jre.check_dl_java_ivy(download_dir=os.getcwd()+'/pyraphtory_jvm/data/')
+        build.run(self)
+
+class PostInstallCommand(install):
+    """Post-installation for installation mode."""
+    def run(self):
+        install.run(self)
+        # PUT YOUR POST-INSTALL SCRIPT HERE or CALL A FUNCTION
```
fabianmurariu, line 17: are these comments important?

fabianmurariu (dismissed):
Assuming this works LGTM

python/pyraphtory_jvm/pyraphtory_jvm/ivy.xml
Haaroon: this is copied from the sbt process into this folder, what do u mean by generated?

python/pyraphtory_jvm/setup.py
Haaroon: nope

Haaroon:
thunderbirds are go

python/pyraphtory/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "pyraphtory"
-version = "0.2.0a1"
+version = "0.2.0a5"
```
miratepuffin, line 3: Can these versions be standardised in next PR?

python/pyraphtory_jvm/pyraphtory_jvm/ivy.xml
miratepuffin, line 1: I believe @fabianmurariu means why do we need this here if we create it fresh.

miratepuffin: Each build*

python/pyraphtory/pyproject.toml
Haaroon: yeah sure

python/pyraphtory_jvm/pyraphtory_jvm/ivy.xml
Haaroon: the ivy file needs to be here, because we use it to upload the sdist to pypi, and it uses the ivy file to build raphtory later, in theory i can delete it as long as always remember to push it when we do a publish

Haaroon: i'll add this as a ticket instead of doing it now, so it reminds me to document the build process

Alnaimi-: Make a ticket for it and tag me please

miratepuffin (approved):

PR #643, opened 2022-10-27
ref progress tracker

iamsmkr:
Refactor QueryProgressTracker and TableOutputTracker to remove dependency on Akka.

Since both the components are client-side components required by QuerySender to be able to send messages received, the messages don't really need to go through Akka.

No

Integration Tests

No

### DIFF
--- a/core/src/main/resources/protocol.proto
+++ b/core/src/main/resources/protocol.proto
@@ -22,7 +22,7 @@ message OptionalId {
 }
 
 // Requests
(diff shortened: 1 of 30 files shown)


ricopinazo:
This is looking good. Now we can stop creating the TopicRepository in RaphtoryContext.scala. I would include that in this PR so this makes complete sense.

miratepuffin (approved):

PR #651, opened 2022-11-01
Fix bug preventing some exceptions from ingestion from bubbling up

ricopinazo:
We now allocate the ingestion executor before answering to the client, so any exceptions thrown whilst creating the spout/builder bubble up in the same gRPC call. Additionally, we log any errors appearing when executing the source. 

Raphtory was getting stuck if the spout/builder had a problem at creation. This solves the issue and prevent the errors from being ignored.

This causes `graph.load()` to throw exceptions, but there is no need to document that for the time being

Integration tests pass.

No

### DIFF
--- a/core/src/main/scala/com/raphtory/internals/components/OrchestratorService.scala
+++ b/core/src/main/scala/com/raphtory/internals/components/OrchestratorService.scala
@@ -10,8 +10,12 @@ import com.raphtory.internals.components.OrchestratorService.GraphList
 import com.raphtory.protocol.GraphInfo
 import com.raphtory.protocol.Status
 import com.raphtory.protocol.success
+import com.typesafe.scalalogging.Logger
+import org.slf4j.LoggerFactory
 
 abstract class OrchestratorService[F[_]: Concurrent, T](graphs: GraphList[F, T]) {
+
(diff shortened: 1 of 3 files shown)


core/src/main/scala/com/raphtory/internals/components/OrchestratorService.scala
```diff
@@ -33,9 +36,11 @@ abstract class OrchestratorService[F[_]: Concurrent, T](graphs: GraphList[F, T])
   final protected def attachExecutionToGraph(graphId: String, execution: Graph[F, T] => F[Unit]): F[Unit] =
     for {
       graph <- graphs.get.map(graphs => graphs(graphId))
-      _     <- graph.supervisor.supervise(execution(graph))
+      _     <- graph.supervisor.supervise(execution(graph).onError { case e: Throwable => logError(e) })
```
fabianmurariu, line 39: thisi is usually ```case NonFatal(e) =>```

fabianmurariu: is there no cleanup to do here?

ricopinazo: Just added the `NonFatal` bit

miratepuffin (approved):

PR #659, opened 2022-11-03
Improve python repr output

ljeub-pometry:
Makes repr output for wrappers more readable in case of Maps/Tuples
Old output was hard to parse, new output can be used directly with eval in python to get dicts/tuples as expected

Output looks better
No


Haaroon (approved):

Haaroon:
why are tests failing?

Haaroon (approved):

PR #660, opened 2022-11-03
make the CSV format use quoted json for printing objects

ljeub-pometry:
Uses the Jackson json writer to print objects 

Java `toString` is useless with arrays and other types and breaks civ convention as object strings may contain commas


ljeub-pometry:
IMG_1684

miratepuffin (approved):

PR #672, opened 2022-11-09
submit Assortativity

Peijie-Zhong:


Haaroon:
Hi, if you can please add a test for this and then we can approve.

Haaroon (changes requested):
Needs a test

Peijie-Zhong:
> 

Hi Haaroon, I've updated a test case. Could you please check if it's OK? Thanks

miratepuffin (approved):

PR #684, opened 2022-11-15
Fix test by rounding results

Haaroon:

- pyraphtory test fails due to rounding differences between host machine and ci cd machine
- Dynamically removing warnings during CI CD build to speed up process (cuts build time by 1 minute)
- Parallelised SBT tests which halves the testing time (From 4 mins to 2 mins)
- Fixed notebooks which has code cells changed to text by mistake
- Added missing dependencies for new pandoc

fix the tests

no

locally made new tests 

no

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -267,7 +267,7 @@ lazy val assemblySettings = assembly / assemblyMergeStrategy := {
   case _                                                                                  => MergeStrategy.first
 }
 
-Test / parallelExecution := false
+Test / parallelExecution := true
 
 Global / concurrentRestrictions := Seq(
         Tags.limit(Tags.Test, 1)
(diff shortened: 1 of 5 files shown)


Haaroon:
the build will fail because the raphtory tests currently fail and wont be fixed until the api is updated, all the bits i have touched build fine

build.sbt
```diff
@@ -269,6 +269,8 @@ lazy val assemblySettings = assembly / assemblyMergeStrategy := {
 
 Test / parallelExecution := true
 
+core / Compile / logLevel := Level.Error
```
Alnaimi-, line 272: I don't think we should do this. Because then we will miss out on deprecation logs. Instead we have a ticket in current sprint to address all warn messages.

Haaroon: this currently saves 1 minute of building time with it present, maybe i can just add it during the ci cd somehow

Haaroon: okay i figured it out, i can append this to the file before the build

Alnaimi- (approved):
This looks fine with me, warnings should be checked locally by the developer

PR #696, opened 2022-11-17
PyRaphtory Ingestion Tutorial

miratepuffin:
In this PR I am redoing the introductory sections of the tutorial as Jupiter notebooks. This allows the user to download and execute the tutorial themselves.

Misc Changes include:

- Moving the sources to be a top level (core) package
- General Tidy up of random classes
- Deleted the LOTR pulsar runner
- Deleted some old pulsar docker scripts
- added a `check_raphtory_logging_env` function to the py4jgateway which turns java logging off unless the user specifically adds it. This needs a proper fix, but does the job for the time being.
- Changed all python prints to logging
- Pointed the output from the java version check to /dev/null unless there is an error
- Added a `disableReflectWarning` function which does what it says on the tin.
- Added a `make python-build-quick` function which only rebuilds pyRaphtory

Changes to the Python API:
- Renamed the Properties to defined them as either explicitly mutable or immutable.
- Overwritten the `add_vertex` and `add_edge` functions to:
     - Allow the src/dst id to be strings (removing the need for a user to call assign_id)
     - Pythonised the properties as a list and type as a string for `add_vertex` and `add_edge`
- Added a check in the `to_df` function which checks if there is window labels in the data and if not doesn't add the column in the dataframe - this removes all the `Nones` which was quite confusing for new users.
- Added the CsvFormat to python.
- Created a new pyraphtory package for sources
    - Moved the pyraphtory source from the builders package to the sources package
    - Added the rest of the properties to input.py
- Renamed the builders pyraphtory package to input, brining it inline with scala

Actual Doc changes include:
- Created an Ingestion tutorial which goes through all the different ways to add data into a graph 
- Updated the original tutorial to remove references to java logs and windows as no longer there
- Added a second LOTR dataset to explain merging of sources.
- Removed references to kube in docs.
- Added a test on the CI to check if the documentation notebooks run correctly

Documentation is not hands on enough to get users actually running things.
This is a change to the docs
The notebooks run and the docs compile.
- [ ] Expand the second dataset
- [ ] Need to fix the context closing as it currently stops other graphs being created
- [ ] Need to remove the concept of Immutable properties and swap them to metadata
- [ ] Need to change Properties to a Dict to remove the concept of types
- [ ] Errors with Jpype seem to crash the Kernel 
- [ ] Can we get rid of the Row object?
- [ ] If we filter vertices (and their edges) should this change the history of the remaining nodes
- [ ] We need to document the environment variables for the CLI
- [ ] Need to expand the CLI past just standalone
- [ ] Need to add get_name and get_graph_names functions to graph and cox
- [ ] Need to add some more tests to pyraphtory looking at properties and types
- [ ] Need to check over NFT examples with Jpype
- [ ] Exceptions in .load randomly pop up in later cells
- [ ] Need to investigate a better way of connecting to databases


docs/source/conf.py
```diff
@@ -58,8 +58,8 @@
 ]
 
 # Uncomment to turn of rebuilding of scala and algorithm docs (use when writing other docs to speed up compile)
-# build_scaladocs = False
-# build_algodocs = False
+build_scaladocs = False
```
miratepuffin, line 61: Need to change this back

python/pyraphtory/tests/test_pyraphtory.py
```diff
@@ -78,11 +78,11 @@ def test_run_page_rank_algo(self):
     def test_check_vertex_properties_with_history(self):
         graph = self.ctx.new_graph()
         graph.add_vertex(1, 1, Properties(
-            ImmutableProperty("name", "TEST"),
-            StringProperty("strProp", "TESTstr"),
+            ImmutableString("name", "TEST"),
```
Haaroon, line 81: To do review it to be a normal string in the future, with an immutable boolena flag?

core/src/main/scala/com/raphtory/spouts/WebSocketSpout.scala
```diff
@@ -20,20 +20,7 @@ import java.io.InputStreamReader
   * @param auth The authorization code in Base64 format to access the url e.g. "Basic XXXXXXXXX"
   * @param contentType The content type of the data e.g. application/json
   *
-  * @example
```
Haaroon, line 23: why did u remove the example, its used in the docs to show how it works

Haaroon (commented):
I have reviewed most of the files, largely seems okay

docs/source/index.rst
```diff
@@ -53,7 +52,6 @@ Alternatively, hit next and lets get Raphtory :doc:`installed <Install/python/py
    :caption: Deploying Raphtory
 
    Deployment/baremetalsingle.md
```
Haaroon, line 54: Move examples to begin before the deployment. 

Put Python API last.

core/src/main/scala/com/raphtory/spouts/WebSocketSpout.scala
miratepuffin: The code was old

Thrameos:
I am a bit concerned with the item that errors in JPype crash kernel.  Is this an error in JPype or uncaught execeptions generated when using JPype.  I try hard to make JPype bulletproof so if you find a way to crash it please send an issue report.

miratepuffin:
Hi @Thrameos, that is very much a short hand for an error I had where we are not catching some exceptions within Raphtory and this was causing my Notebook kernel to die. This is possibly due to the amount of output being printed amongst a host of other things, I don't believe it is an issue with JPype itself. When I dig into it I will be sure to let you know what the problem was!

Thrameos:
Okay so two ways to catch it.   Either use a specific Java type of which ``java.lang.Throwable`` is the best because it is the root of all java exceptions, or use the JPype base class for exceptions ``JException``.   Both should give you the same behavior.

miratepuffin:
Thanks @Thrameos! Ok to tag you in the PR when we are working on it?

Thrameos:
Sure, you can tag me for any issues you have while working with jpype.  Though as I scan every issue on github that contains "jpype" in the context at least once a day, I tend to pop up regardless of whether I am wanted or not.  Helps to have a a unique project name.

https://github.com/search?o=desc&q=jpype&s=updated&type=Issues

PR #699, opened 2022-11-22
Ingestion is not blocking when it needs to

fabianmurariu:
Fix ingestion blocking by only finalizing when the ingestion is done

ensure the unblocking message only happens at the end of the stream

No

Unit tests
No


core/src/main/scala/com/raphtory/internals/components/ingestion/IngestionExecutor.scala
```diff
@@ -85,7 +85,7 @@ private[raphtory] class IngestionExecutor[F[_], T](
                                     queryManager.sendAsync(NonBlocking(sourceID = source.sourceID, graphID = graphID))
```
Alnaimi-, line 85: How eager is this?

iamsmkr:
We can probably close this PR as this problem is already solved in our branch which will be merged soon!

PR #704, opened 2022-11-23
Feature/JPype

ljeub-pometry:

Changes the Scala interop to use JPype instead of Py4j when running locally. 

Advantages: 
 - improved performance due to avoiding inter-process communication
 - debuggable algorithm execution (needs some hacks to work with threads spawned from the jvm)

Py4j is slow 

Changes are transparent to the user, everything should work as before.

Sample script runs, existing python tests pass, manually tested remote connection to standalone instance still works

Figure out if it is possible to fix the overload bug in JPype

### DIFF
--- a/python/pyraphtory/pyraphtory/debug.py
+++ b/python/pyraphtory/pyraphtory/debug.py
@@ -1,19 +1,22 @@
-try:
-    import pydevd
-    pydevd.connected = True
-    pydevd.settrace(suspend=False)
-    def enable_pydev_debug():
-        """
-        Use this function to ensure the pydevd debugger (used by e.g. Pycharm)
-        works for python code called from a thread created by the jvm.
+import sys
 
-        """
+
+def enable_pydev_debug():
+    pass
+
+
+if hasattr(sys, "gettrace") and sys.gettrace() is not None:
+    try:
+        import pydevd
         pydevd.connected = True
         pydevd.settrace(suspend=False)
-except Exception as e:
-    def enable_pydev_debug():
+        def enable_pydev_debug():
+            """
+            Use this function to ensure the pydevd debugger (used by e.g. Pycharm)
+            works for python code called from a thread created by the jvm.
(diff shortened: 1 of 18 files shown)


python/pyraphtory/pyraphtory/debug.py
```diff
@@ -0,0 +1,19 @@
+try:
```
Haaroon, line 1: do you still need this?

ljeub-pometry: Something like this is necessary to get the debugger to work for the callbacks. The problem is that they run in threads started from the jvm which are otherwise ignored.

miratepuffin (approved):

miratepuffin:
Looks good to me, need to rejig the install instructions etc/dependencies, but I can do that

Thrameos:
Any issues with converse to JPype?

ljeub-pometry:
> Any issues with converse to JPype?

Overall we are quite happy with how this went and certainly appreciate the roughly 10-fold performance improvement compared to the old Py4J-based implementation. The main issue we have run into has to do with overload resolution. I managed to reproduce the issue with some simple java classes and will open an issue on the JPype GitHub page. The other issue we ran into was debugging python code called from a new java thread as per this issue: https://github.com/jpype-project/jpype/issues/1018 (we are currently using the workaround mentioned there).

As more of a feature request, we have some situations where it is much more convenient to run python within the jvm rather than the other way around. We are currently using Pemja in that situation but supporting to different interface libraries is a bit of a headache so if that use case could be supported in JPype, that would simplify our situation significantly...

Thrameos:
There is an experimental project epypj which allows full embedding of python in java (like it exposes all of the python CAPI to Java).  But it is on hold for lack of users willing to help make the testbench.

I will look into your issue.  Most like it just requires a cast but it may be an odd ball case we didnt cover.

PR #705, opened 2022-11-23
fixed workflow

Haaroon:

workflow was missing a step

failed build

usign `act` locally


ljeub-pometry (approved):
This does look like it should fix it and it doesn't look like it triggered the error. Ok to merge!

PR #706, opened 2022-11-23
disable build

Haaroon:

we dont need build. we dont need it to make tags and release artifacts so disabled

makes ci fail

here on this ci


ljeub-pometry (approved):
Ok to merge as a quick fix but we need to figure out what the real solution is.

PR #717, opened 2022-12-01
DEV-1236 / Akka clean-up

ricopinazo:
- All the old classes for the old Component/TopicRepository abstraction were removed.
- The watermark updating was removed from the writers.
- The message clase classes were sorted, removing unnecessary ones and moving some client-facing ones to Protobuf.

We need this to finish the 

No

Unit tests locally and on the pipeline, integration tests only locally

There are still some client-facing case classes that we might want to move to Protobuf to improve interoperability in the future, but it isn't worth it for the time being and we can tackle it once we are sure we need it.


miratepuffin (approved):
Will we need any of the pulsar connector functions in the future?

PR #720, opened 2022-12-03
Feature/analysis tutorial rework

miratepuffin:
#### Actual Changes to the docs
- Created new notebook about submitting queries over time
- Created new notebook for 6 degrees of separation algorithm creation
- Removed broken or out of date items from the TOC, to be added back in when they are fixed
- Removed loads of old images that aren't used any more in the docs

#### Changes to the query API
- Changed windows to be end aligned by default
- Removed the offset for walk - can add this back, but trying to simplify things at the moment
- Drastically simplified the Perspective generation now that we have bounded times
- Fixed a bug with the to_df function which I introduced which caused it to return after 1 perspective.
- Fixed the windowing functions and introduced the alignments

The analysis tutorials were based on scala and out of date
It does have many but these were fixed alongside the documentation
Wrote new unit tests for changes to the Perspective controller + all docs are now tested via nbmake

- The docs can always be better :) 
- The dynamic class loader test takes a VERY long time - need to investigate this
- Need to expand the perspective controller tests
- Need to add a new tutorial on how to deal with timestamps
- Need to reenable multithreading for analysis - currently has a race condition
- Anyway for the col names to be automated in to_df - currently has the option to set them to whatever as long as you have the right number of values.
- Need to add a section on sinks.


Haaroon (approved):
🚒

PR #731, opened 2022-12-07
remove builder from pyraphtory to fix tests

Haaroon:


ljeub-pometry (approved):
Just a post-refactor fix for classes that were moved. Good to merge!

ljeub-pometry:
Although, seems the version change for the package is causing issues?

PR #750, opened 2022-12-09
Source should unblock and graph should not be destroyed when ingestion fails

ljeub-pometry:

Improved error handling when load fails

failing load could lead to a hang in python as the graph remained blocked forever

No

Hang is resolved


ricopinazo (approved):
Looks good to me

miratepuffin (approved):

miratepuffin (approved):

PR #754, opened 2022-12-12
Feature/dev 1384 simplify workflow

Haaroon:

- rename a folder so they run on test
- simplify the test.yml so the make file can run all these commands more easily 
- separated tests so they all run in parallel reduced test time from 11m7s to 6m9s


Haaroon (commented):
none of these files changes just renamed the folder, only two files changed

miratepuffin (approved):

PR #755, opened 2022-12-13
Feature/lazy docs

ljeub-pometry:

All docstrings and annotations are processed lazily to reduce the wrapper creation overhead

Import times and warmup for wrappers was slow

no

docs still build and imports are much faster

more cleanup of the documentation and including docstrings for classes

### DIFF
--- a/python/pyraphtory/pyraphtory/interop.py
+++ b/python/pyraphtory/pyraphtory/interop.py
@@ -366,7 +366,7 @@ class ScalaProxyBase(object):
             method = method_array[0]
             try:
                 exec(_codegen.build_method(name, method, _jpype), globals(), output)
-                output[name].__doc__ = _codegen.LazyDocstr(method.docs())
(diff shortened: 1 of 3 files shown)


python/pyraphtory/pyraphtory/interop.py
```diff
@@ -473,8 +480,8 @@ class InstanceOnlyMethod(object):
     def __init__(self, method):
         self.__func__ = method
         self.__name__ = method.__name__
-        self.__signature__ = inspect.signature(method)
-        self.__doc__ = f"Instance only method {self.__name__}{self.__signature__}"
+        # self.__signature__ = inspect.signature(method)
+        # self.__doc__ = f"Instance only method {self.__name__}{self.__signature__}"
```
miratepuffin, line 484: Are these needed?

miratepuffin (approved):
I believe we just need to delete these comments and then we can merge

PR #758, opened 2022-12-15
Feature/graph state output

narnolddd:

Added documentation to `GraphState` algo
Made graph-wide algorithms actually store the required state to be used in conjunction with GraphStateOutput

So that graph-wide algorithms (e.g. `GlobalTriangleCount`, `NodeEdgeCount`) can be used in conjunction with the `GraphStateOutput` class

Yes, `GraphState` is a new algorithm so added documentation based on its `NodeList` counterpart.

Tests for existing graph-wide algorithms are still passing.

Some algorithms have a more complex `tabularise` function, could do with spending time to make these extend `GraphState` too (e.g. `ThreeNodeMotifs`)


miratepuffin (approved):
Just need to removed some commended out code

PR #760, opened 2022-12-16
Dev 1377/pyraphtory jvmtoml

ljeub-pometry:

Reworked the build system for pyraphtory such that pip install works from GitHub and builds core jars to put into source distribution and jre and dependencies are included in wheel. Also changed it such that JAVA_HOME is ignored unless the special environment variable PYRAPHTORY_USE_SYSTEM_JAVA is set to something truthy

Make it easier to install/develop pyraphtory

Only for developers

The tests pass after refactoring. Also confirmed that the generated sdist can be installed and results in a working pyraphtory on a barebones linux vm without sbt or java installed.

ivy dependency conflict resolution could use some more cleanup. The current solution is a hack.


Makefile
```diff
@@ -2,84 +2,89 @@ SHELL:=/bin/bash -euxo pipefail
...
+.PHONY: sbt-thin-build
+sbt-thin-build: version clean sbt-build
+
+
+.PHONY: python-build
+python-build: version
+	python -m pip install .
+
+
+.PHONY: python-build-editable
```
miratepuffin, line 55: 🙌🏻 🙌🏻 🙌🏻

build.sbt
```diff
@@ -128,7 +128,8 @@ lazy val arrowCore =
                     chronicleMap,
                     fastUtil,
                     commonsLang,
-                    junitInterface
+                    junitInterface,
+                    netty classifier osDetectorClassifier.value,
```
miratepuffin, line 132: Oh nice! Does this fix the windows issue by chance?

examples/lotr/src/main/python/LOTR-PyRaphtory.ipynb
```diff
@@ -2,53 +2,118 @@
...
      "name": "#%%\n"
     }
-   }
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Requirement already satisfied: pyvis in /Users/lucasjeub/opt/miniconda3/lib/python3.10/site-packages (0.3.0)\n",
```
miratepuffin, line 53: Might want to clear this just as its got your name in it.

miratepuffin (approved):

miratepuffin (approved):

Haaroon (changes requested):
if i understand correctly this merges the pyraphtory_jvm into pyraphtory, if so then 

- the docs need to be changed to point to the new way to install pyraphtory for users. 
- binder environment needs to be changed otherwise it will fail 

otherwise it looks good, i just built this on my mac

examples/lotr/src/main/python/LOTR-PyRaphtory.ipynb
ljeub-pometry: done

build.sbt
ljeub-pometry: Not sure, it is something Fabian did to try to reduce the number of package version conflicts

ljeub-pometry:
Fixed binder environment and removed mention of pyraphtory_jvm from the install instructions. We need to add more docs on advanced install (install from source, distributed option, etc)

miratepuffin:
Will merge this for the time being and we can make a ticket for the advanced docs

Haaroon (approved):

PR #762, opened 2022-12-21
Add SqlEdgeSource and SqlVertexSource

ricopinazo:
- Adds two new sources: SqlEdgeSource and SqlVertexSource.
- Moves the sending of updates logic from the GraphBuilder and Graph traits to the IngestionExecutor itself.

- The two new sources allow the ingestion from SQL databases
- Moving the sending of updates to the IngestionExecutor was necessary because the new sources don't use the Spout/GraphBuilder pattern anymore, they extend the Source trait directly, so this class has been generalized to return an fs2.Stream of chunks of updates that the IngestionExecutor later send to the writers.

Yes, this adds two new sources. Both were appended a scaladoc and a new section on the ingestion notebook with an example using both was added.

This PR passes all the existing tests. Additionally, a suite of unit tests for both sources was added under SqlSourceTest test class and a python test was added as well just to check this works in Python. The performance was tested to confirm that there is no regression with 10 iterations of the TwitterTest comparing this branch and master (commit 21362a1).

No


core/src/main/scala/com/raphtory/internals/graph/GraphBuilder.scala
```diff
@@ -7,33 +7,18 @@ import cats.effect.std.Dispatcher
...
 
-  private val partitioner       = Partitioner(defaultConf)
-  private val totalSourceErrors = TelemetryReporter.totalSourceErrors.labels(s"$sourceId", graphId) // TODO
-
-  def buildGraphFromT(chunk: Chunk[T], globalIndex: Ref[F, Long]): F[Unit] =
+  def parseUpdates(chunk: Chunk[T], globalIndex: Ref[F, Long]): F[Seq[GraphUpdate]] =
     for {
       b <- F.delay(new mutable.ArrayBuffer[GraphUpdate](chunk.size))
-      cb = UnsafeGraphCallback(partitions.size, sourceId, -1, graphId, b)
+      cb = UnsafeGraphCallback(-1, b)
```
miratepuffin, line 21: put index=-1 just to know what it is

python/tests/test_pyraphtory.py
```diff
@@ -161,3 +164,10 @@ def test_mutable_integer(self):
         i = pyraphtory.input.MutableInteger("test", 2)
         self.assertEqual(i.value(), 2)
         self.assertEqual(i.key(), "test")
+
```
miratepuffin, line 167: Can we change this to a local sqlite instance which we only have to download once

miratepuffin (changes requested):

miratepuffin (approved):

PR #769, opened 2023-01-06
set chunksize to be a conf variable

narnolddd:

Proposed to make chunkSize a conf variable

Makes it easier to vary the threading chunk size as needed for different algorithm workloads on the fly

No, just an extra environment/conf variable

Ran it with two different chunk sizes given as an environment variable and they changed accordingly

N/A


miratepuffin (approved):

Haaroon:
change from draft so this can be merged

PR #773, opened 2023-01-09
Remove sbt versions and bump docs to 1.7.0 sbt

jamesalford:
Remove sbt versions from example files 

So examples use the same sbt version as core project

No

sbt assembly and make python-build

No

### DIFF
--- a/docs/project/build.properties
+++ b/docs/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.6.2
+sbt.version=1.7.0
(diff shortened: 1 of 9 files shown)


docs/project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.6.2
+sbt.version=1.7.0
```
miratepuffin, line 1: Should this one not be in core?

jamesalford: removed and doc make adjusted

miratepuffin (approved):

miratepuffin (approved):

PR #774, opened 2023-01-10
Bug/python properties

ljeub-pometry:

- fixes property value constructors from python

no

added tests to make sure all properties can be constructed correctly

no


miratepuffin (approved):
Have checked locally, looks good to me!

PR #792, opened 2023-01-31
DEV-1451 / Move source index to global scope to avoid collisions

ricopinazo:
This PR moves the index used to order timestamps from the source level to the IngestionService, so it is shared between sources.

This was causing a bug, because if there are two edges between the same nodes and for the same timestamp, there is a chance for them to have the same index. Therefore, when adding them, they are considered the same and the number of events for the vertex is incremented only by one.

This bug was not showing up because we were using `ConnectedComponentes` algorithm in MultiSpoutDeploymentTest, which works even if the number of vertex events changes. I changed the test to use `GraphState` algorithm, which doesn't mask incorrect ingestion.

This bug can be reproduced as follows:
```bash
git checkout 33359c8b # most recent commit for master right now
git checkout bug/DEV-1451-multi-spout -- **/MultiSpoutDeploymentTest.scala
sbt "testOnly *MultiSpoutDeploymentTest"
```

No

Regular tests, which were changed to detect more bugs for multi spout deployments

No


iamsmkr (approved):
LGTM

PR #793, opened 2023-02-01
DEV-1453 / Fix disruptor taking different PojoEdge for the same edge

ricopinazo:
We now synchronize the adding of the edge events for both vertices

This is needed so both vertices take the same PojoEdge for storage. As this function is called in parallel and there is a chance of getting the PojoEdge from one event for the source vertex and a different PojoEdge from another event for the target vertex.

No

regular tests and performance test on TwitterTest to confirm that there is no performance downgrade

No


iamsmkr (changes requested):
We need separate events. Revert this!

ricopinazo:
The performance for TwitterTest hasn't degraded. It is safe to merge this PR
Screenshot 2023-02-01 at 12 48 19

PR #794, opened 2023-02-01
build binder on commit to master

Haaroon:

change 1 line

binder too slow otherwise

no

in ci, cant test until pushed

no


ljeub-pometry (changes requested):
I think we need this to run for both tags and master, otherwise this will be slow for actual releases

PR #796, opened 2023-02-02
DEV-1452 / Fill isolated vertices with empty lists in EdgeList

ricopinazo:
`EdgeList` allows getting a list of properties for every edge given their names. It does so by creating a state in every vertex for each property and filling them with the properties of all the edges. Now `EdgeList` fills these states with empty lists if the vertex has no edges. The same change was applied to `TemporalEdgeList`, whose implementation is very similar.

We need the changes to prevent `EdgeList` from trying to use the state from the vertex instead of from the edges in the cases where there are no edges and the state in the vertex already had a value.

No

The error can be reproduced by running this python code in master (33359c8b):

```python
from pyraphtory.context import PyRaphtory
from pyraphtory.input import MutableInteger
ctx = PyRaphtory.local()
graph = ctx.new_graph()
graph.add_vertex(0, 1, [MutableInteger("test", 1)])
graph.add_vertex(0, 2, [MutableInteger("test", 2)])
graph.add_edge(0, 1, 2, [MutableInteger("test", 3)])
print(graph.execute(ctx.algorithms.generic.EdgeList("test")).to_df())
print(graph.execute(ctx.algorithms.temporal.TemporalEdgeList("test")).to_df())
```

With the changes in this PR, it doesn't fail anymore.

No


ljeub-pometry (changes requested):
- Would have been good to add the toy example as a test to show it is actually fixed. 
- Also, `Seq.empty` is generally better than `Seq()` as it doesn't create a new object. 
- I really don't like the zip/unzip logic. It is way more complicated than this needs to be and is the only reason we had this bug in the first place. A simple `foreach` over properties with an internal map over edges would do just fine.

ricopinazo:
You are right @ljeub-pometry. I changed the implementation as you proposed and looks way nicer and fixes the bug as well. I added a python test with the code in the PR description as well.

ljeub-pometry (approved):
Looks good and thanks for the test

PR #804, opened 2023-02-07
Try to fix python shutdown segfaults

ljeub-pometry:

Hacked into the python shutdown sequence to pass control to java

Python shutdown causes segfaults when trying to turn the jvm started by Jpype due to demon threads trying to access freed memory

Not normally unless you are doing something funky with `atexit`

Manually tested the different shutdown scenarios and it seems to work with exit codes preserved.


python/src/pyraphtory/interop/_interop.py
```diff
@@ -111,6 +111,7 @@ def check_raphtory_logging_env():
         pass
 
     jpype.startJVM(_config.java_args, classpath=_config.jars.split(":"))
+    _config._jpype_running = True
```
Alnaimi-, line 114: Cool

python/src/pyraphtory/_config.py
```diff
@@ -4,9 +4,67 @@
...
+        self.exception = exc
+        self._orig_exc_handler(exc_type, exc, *args)
+
+def exit_handler():
+    """Forwards exit code to java and calls java exit if jvm was started"""
+    try:
+        if _jpype_running:
+            System = JClass("java.lang.System")
+            if hooks.exit_code is not None:
+                System.exit(hooks.exit_code)
```
Alnaimi-, line 55: Crazy. I like it. Hopefully Python invokes these gracefully.

Alnaimi- (approved):

PR #818, opened 2023-04-24
Simplify windowing in graph wrapper structs

ricopinazo:


docbrown/src/db/path.rs
```diff
@@ -10,7 +11,7 @@ use std::ops::Range;
 use std::sync::Arc;
 
 #[derive(Copy, Clone)]
-pub(crate) enum Operations {
+pub enum Operations {
```
ljeub-pometry, line 14: why is this needed?

python/src/dynamic.rs
```diff
@@ -14,15 +15,13 @@ impl<G: GraphViewInternalOps + Send + Sync + 'static> DynamicGraphView for G {}
...
-        Self(Arc::new(graph))
-    }
+pub(crate) trait FromGraph {
+    fn into_dynamic(self) -> DynamicGraph;
 }
 
-impl From<Graph> for DynamicGraph {
-    fn from(value: Graph) -> Self {
-        Self(Arc::new(value))
+impl<G: GraphViewOps> FromGraph for G {
```
ljeub-pometry, line 22: This should be called IntoDynamic since it implements the into-style method?

miratepuffin (approved):
Looks a lot cleaner to me!

ljeub-pometry (approved):
This is good and simplifies things a lot. Just one curious visibility change that I don't think is necessary?

python/src/dynamic.rs
ricopinazo: Done

docbrown/src/db/path.rs
ricopinazo: After pairing for a while, the conclusion of this is that ideally `DynamicGraph::window` should return `DynamicGraph` instead of `WindowedGraph`. That way we could avoid doing translations that force us to make some things public, such as `Operations`. However, we tried it and it seems impossible or at least very hard with the approach we followed. It might be feasible some other way though, so we can try to improve this later.

PR #825, opened 2023-04-26
Refactor state api

fabianmurariu:
Graph.clone is not very cheap so VertexView and all structs that relied on cheap clone now move to Arc<G>
new API for running algorithms
deprecate Mul32 and Sum32

new API used for
page_rank
connected_components
triangle_count

removal of program.rs and moving to the new API for all the algorighms


miratepuffin:
All looking good as per our convo today!

PR #849, opened 2023-04-27
Feature/edge time index

ljeub-pometry:

Adds a time index for edges and further optimises use of the vertex time index.

Much improved performance on windowed operations.

There is more optimisation potential, especially for degrees.


ljeub-pometry:
Not ready to merge, it is using too much memory as is.

miratepuffin (approved):

PR #851, opened 2023-04-28
add notebooks for companies house

rachchan:


miratepuffin (approved):
LGTM - couple of extensions for us to do after merge
* add helper functions for earliest/latest_datetime() start_datetime/end_datetime on graph/vertex/edge
* g.edges.property(“resigned_on”).max() should be working (or something along these lines)
* All of the g.window() calls can now be datetimes - same as the rolling

PR #852, opened 2023-05-02
Enron example and API improvements

ricopinazo:
This PR adds a notebook `enron.ipynb` and a number of improvements that were necessary to make it work properly, among others:
- structs such as `PyGraphWindowSet`, `PyVertexWindowSet`, etc, are now python iterables instead of iterators, which means that now we need two structs for each of them, one for the utterable and one for the iterator returned by the iterator. This is a lot of boilerplate and we might want to simplify it somehow
- `add_vertex` and `add_edge` in python now support strings as timestamps
- the functions `window` and `at` now support strings as bounds, but only in Python. In rust, we might implement IntoTime for chrono `Datetime` in the future to allow for the same type of things without adding `Result` types as the output, but in Python we need to return `PyResult` anyways because we extract the time from a `&PyAny`.
- a trait LayerOps has been defined in a similar fashion to TimeOps but for layer operations. It has been implemented for all the structs `TimeOps` is, except `EdgeView`, since an edge references to only one layer by definition
- the layer function has been added as well to the python versions of this structs. As a result, it was necessary to create translations, for instance, for PyVertex from `VertexView<LayeredGraph<G>>`, in addition to `VertexView<WindowedGraph<G>>` and `VertexView<DynamicGraph>`. Instead, this has been generalized, so we translate from `VertexView<G: GraphViewOps + IntoDynamic>`, i.e. we create a `PyVertex` from a `VertexView` that contains a graph that can be translated into a `DynamicGraph`. And instead of implementing `IntoDynamic` for `GraphViewOps`, we have specialized the implementations for the four types we have implementing `GraphViewOps`, because that way, the implemetation of IntoDynamic for `DynamicGraph` is just returning itself, avoiding unnecessary wrapping.
- a `doc_strings` module has been added containing macros to define common doc strings. Has been placed in the rust crate with the aim of potentially using it for the rust API as well

Documentation for added functions such as `layer` and `default_layer` was added via the `doc` directive, but there might be some docs out of date.

unit and integration tests

Yes, docs should be reviewed


miratepuffin (approved):
Amazing stuff!

PR #861, opened 2023-05-09
Dev 1654 post merge

fabianmurariu:
Removal of things not needed


raphtory/src/db/view_api/internal.rs
```diff
@@ -227,30 +227,6 @@ pub trait GraphViewInternalOps {
...
-    /// # Returns
-    ///
-    /// * `Box<dyn Iterator<Item = EdgeRef> + Send>` - An iterator over all the edge references.
-    fn edge_refs_window(
-        &self,
-        t_start: i64,
-        t_end: i64,
-        layer: Option<usize>,
-    ) -> Box<dyn Iterator<Item = EdgeRef> + Send>;
-
```
ljeub-pometry, line 253: These methods should have been used by GraphViewOps `edges` for future optimisation (I am currently working on that...). I see that for some reason that is not the case. Can we make `GraphViewOps` use them instead of deleting them?

ljeub-pometry (changes requested):
`edge_refs` and `edge_refs_window` were supposed to be used by `GraphViewOps.edges()` so that can be optimised. For some reason that was not the case. Can we keep them around as I am currently making the changes to actually make use of them...

raphtory/src/db/view_api/internal.rs
fabianmurariu: I can leave them in, I don't think I'll use them, this is mean to be a small PR to move to GraphViewOps for the eval part

ljeub-pometry (approved):
LGTM

miratepuffin (approved):
LGTM - @ljeub-pometry you happy for these functions to hit the bin?

miratepuffin:
The man answers questions before I ask them 🧙🏻‍♂️

PR #864, opened 2023-05-09
add layer_name function to EdgeView and PyEdge

ricopinazo:

### DIFF
--- a/python/src/edge.rs
+++ b/python/src/edge.rs
@@ -286,6 +286,10 @@ impl PyEdge {
         self.edge.time()
     }
 
+    pub fn layer_name(&self) -> String {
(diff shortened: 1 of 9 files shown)


python/src/edge.rs
```diff
@@ -286,6 +286,10 @@ impl PyEdge {
         self.edge.time()
     }
 
+    pub fn layer_name(&self) -> String {
```
miratepuffin, line 289: Can you add a small doc string

miratepuffin (changes requested):

python/src/edge.rs
ricopinazo: Done

PR #880, opened 2023-05-11
Updated Companies House notebook and added new date time methods in python

rachchan:
Python methods for outputting earliest, start, latest and end time for vertex, edge and graph, all in date time format.

Having epoch timestamps being outputted in notebook was a bit janky, now can have date times which are much easier to understand in analysis

yes and yes

unit tests


miratepuffin (approved):

miratepuffin:
Need to add all of the property functions to the Edges

PR #897, opened 2023-05-17
Feature/vertex view ops in eval

ljeub-pometry:

Make EvalVertexView support VertexViewOps

Make all the api available for algorithms

### DIFF
--- a/raphtory/src/db/view_api/time.rs
+++ b/raphtory/src/db/view_api/time.rs
@@ -3,7 +3,7 @@ use crate::core::time::{Interval, IntervalSize, IntoTime};
 use chrono::DateTime;
 
 /// Trait defining time query operations
-pub trait TimeOps: Send {
+pub trait TimeOps {
     type WindowedViewType: TimeOps;
 
     /// Return the timestamp of the default start for perspectives of the view (if any).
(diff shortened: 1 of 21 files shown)


fabianmurariu (approved):
LGTM

raphtory/src/db/view_api/time.rs
```diff
@@ -115,16 +115,45 @@ impl<T: TimeOps + Clone + 'static> WindowSet<T> {
...
-            Box::new(self.clone().map(|view| view.end().unwrap() - 1))
+    /// Returns the centered time index of this window set
+    pub fn center_time_index(&self) -> CentredTimeIndex<T> {
+        CentredTimeIndex {
+            windowset: self.clone(),
         }
     }
+
+    /// Returns the end time index of this window set
+    pub fn end_time_index(&self) -> EndTimeIndex<T> {
```
ricopinazo, line 126: Instead of having CentredTimeIndex and EndTimeIndex, could we have only one TimeIndex struct with the center attribute inside, so we check that flag in the implementation of `next`? That way we can maintain the `time_index()` function as it was instead of having two versions of it, which maybe is a little confusing for users. Actually that simplifies the python implementation for `time_index`, that can just rely on this one instead.

PR #898, opened 2023-05-17
remove the need to declare a GraphViewInternalOps when proxying one

fabianmurariu:
Small change to the GraphViewInternalOps to derive one when we proxy


ljeub-pometry (approved):
Looks good and makes sense now that we have multiple of these wrappers.

PR #909, opened 2023-05-19
Stablecoin nb

iamsmkr:


raphtory/src/algorithms/mod.rs
```diff
@@ -40,3 +40,4 @@ pub mod triangle_count;
 pub mod triplet_count;
 pub mod three_node_local;
 pub mod three_node_motifs;
+pub mod motifs;
```
miratepuffin, line 43: I think this needs to be in the three_node_motifs file - so we don't have to rename it all when we add other types of motif counting

miratepuffin: Actually on this point, can we either have them as one file - or in a sub module together? We probably need to start organising the algorithm folder a bit better anyway.

miratepuffin (changes requested):

Issue #920, opened 2023-05-19
Raphtory-IO needs to be released to crates


miratepuffin:
Now that we have split Raphtory-core and Raphtory-IO to allow the WASM complication to work we need to publish raphtory-IO to crates and attach this to the release CI.

PR #971, opened 2023-05-23
make graphql server load all the graphs under graphs folder

ricopinazo:


raphtory-io/Cargo.toml
```diff
@@ -28,4 +28,4 @@ bzip2 = "0.4"
 flate2 = "1.0"
```
miratepuffin, line 29: I think for #920 we need to also set this version number to workspace?

miratepuffin (approved):
Looks great!

PR #972, opened 2023-05-24
Feature/graph ception

ljeub-pometry:

- Make it possible to have a graph as a property.
- Attempts to make the graph cheaper to clone.

### DIFF
--- a/raphtory/src/db/graph.rs
+++ b/raphtory/src/db/graph.rs
@@ -7,9 +7,9 @@
 //! # Examples
 //!
 //! ```rust
-//! use raphtory::db::graph::Graph;
+//! use raphtory::db::graph::InternalGraph;
 //! use raphtory::db::view_api::*;
-//! let graph = Graph::new(2);
+//! let graph = InternalGraph::new(2);
 //! graph.add_vertex(0, "Alice", &vec![]);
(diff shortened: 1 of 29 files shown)


raphtory/src/db/view_api/internal.rs
```diff
@@ -650,66 +651,85 @@ pub trait GraphViewInternalOps {
     ) -> Box<dyn Iterator<Item = LocalVertexRef> + Send>;
 }
 
+pub trait WrappedGraph {
```
fabianmurariu, line 654: What about `AsRef<G>`

raphtory/src/db/graph.rs
```diff
@@ -56,7 +59,173 @@ pub struct Graph {
...
+        write!(
+            f,
+            "Graph(num_vertices={}, num_edges={})",
+            self.num_vertices(),
+            self.num_edges()
+        )
+    }
+}
+
+impl<G: GraphViewOps> PartialEq<G> for InternalGraph {
```
fabianmurariu, line 197: this is heavy, really heavy!

raphtory/src/db/view_api/internal.rs
ljeub-pometry: Does not work because of unconstrained parameter when used with generic graph

PR #998, opened 2023-05-31
add initial plugin example

### LINKED ISSUE #914: Create an example/template GraphQL plugin project
We need to create an example project which shows users how they can create a new raphtory algorithm in rust with the pyraphtory bindings, register this for graphQL and run this totally separate from the Raphtory core code base.

### LINKED ISSUE #916: Point GraphQL at a plugin folder to load all user generated functions
Whilst we have swapped to dynamic graphQL in #912 we still need to point the service to a specific plugins folder which it should load everything from.

ricopinazo:
The main changes are:
- Now `rahptory-graphql` works as a library to define custom servers with custom algorithms
- An example was added to show how new algorithms can be defined an added to a user-defined server
- The graphql service has been improved. Now a graph and a windowed graph expose the same functions, and adding support for layers would be straightforward.

These changes are needed so the user is not constrained to the algorithms defined in the library when usin the graphql server

`raphtory-graphql` crate is yet to be documented, but there are tickets for it

There is an initial test added to `raphtory-graphql

### Issues
resolves #914
resolves #916

We should add a ticket to change `DynamicGraph` to be in `raphtory` crate instead of having it duplicated in different crates, probably under a development feature flag.
With these changes, we are loading new algorithms by using an static variable. We should create a ticket to do this properly, since the change doesn't seem trivial


miratepuffin (approved):
A couple of comments discussed on call - to be turned into tickets by @ricopinazo - otherwise looks good to me!

PR #1011, opened 2023-06-01
Fixing the ingestion of the stable coin data

### LINKED ISSUE #1001: Stablecoin loader is pointing at the test data instead of the full data set
running the stablecoin example notebook currently fails because the example data loader only downloads the test data which does not contain luna: 
image

We need to point the data loader at the actual SNAP url and manually download the test data only in the tests (also should add a Luna edge to the test data)

miratepuffin:
**General changes:**
- I added a flag to `fetch_file` to say if you wanted to save the file in tmp or not
- I added an unzip function 

**To the stable coin example:**
- I added a subset option to only load the smallest file (which should then work on a laptop)
- I folded the `fetch_file` function into the generic one
- Tided up the class and now correctly pull the data and reads only the CSVs

### Issues
fixed #1001


iamsmkr (approved):
LGTM

PR #1021, opened 2023-06-06
Add datetime to add vertex

### LINKED ISSUE #1013: Extract time needs to work with datetime
Currently when adding an edge or vertex you can do so with strings and ints - this should also accept datetime objects
https://github.com/Pometry/Raphtory/blob/57d845c552cfe8d35e420385ccbaa44ffa948aaa/py-raphtory/src/utils.rs#L131

```python
start_date = datetime(2020, 1, 1)

#current
start_str = start_date.strftime("%Y-%m-%d %H:%M:%S")
graph.add_vertex(timestamp=start_str,id=id)

#new
graph.add_vertex(timestamp= start_date,id=id)
```

Haaroon:
### Issues

resolves #1013

### DIFF
--- a/py-raphtory/src/graph.rs
+++ b/py-raphtory/src/graph.rs
@@ -65,7 +65,7 @@ impl PyGraph {
     /// Adds a new vertex with the given id and properties to the graph.
     ///
     /// Arguments:
-    ///    timestamp (int): The timestamp of the vertex.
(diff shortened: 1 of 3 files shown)


ricopinazo:
Check if you want what I pointed out in a comment. It's not important at all, we can merge this anyways

PR #1037, opened 2023-06-09
Added basic filtering for nodes in graphql

miratepuffin:
Added some basic filtering to nodes in graphQL - currently this is on name, type, in degree and out degree - needs a lot of expansion, but easy enough to do as more complex situations arise.

image

Currently if you have vertices which are of multiple types with different properties you have to request everything from all of them leading to many null fields. You also end up pulling the entire graph every time which is less than ideal.

Yes, will be documented as part of the graphQL schema.

Tested this against several example datasets - will need more testing as we expand the functionality of graphql

Yes, discussed in #1038


ricopinazo (approved):
LGTM

PR #1041, opened 2023-06-13
impl graph properties

iamsmkr:
Implement static and temporal graph properties

### DIFF
--- a/python/tests/test_graphdb.py
+++ b/python/tests/test_graphdb.py
@@ -33,6 +33,18 @@ def create_graph(num_shards):
     return g
 
 
+def test_graph_property():
+    g = create_graph(1)
+    props = {"prop 1": 1, "prop 2": "hi", "prop 3": True}
+    g.add_static_property(props)
+
+    sp = g.property_names(True)
+    sp.sort()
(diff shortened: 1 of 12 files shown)


python/tests/test_graphdb.py
```diff
@@ -33,6 +33,18 @@ def create_graph(num_shards):
     return g
 
 
+def test_graph_property():
```
miratepuffin, line 36: Can you implement everything from the test vertex properties function

raphtory/src/db/view_api/internal.rs
```diff
@@ -414,6 +414,8 @@ pub trait GraphViewInternalOps {
     /// Option<Prop> - The property value if it exists.
     fn static_vertex_prop(&self, v: LocalVertexRef, name: String) -> Option<Prop>;
 
+    fn static_prop(&self, name: String) -> Option<Prop>;
```
miratepuffin, line 417: Can you put in a couple of comments here for the APIs

miratepuffin (changes requested):

miratepuffin (changes requested):
Looking good - add the tests and we can merge

miratepuffin (approved):

PR #1044, opened 2023-06-14
Features/materialize subgraph

iamsmkr:
Implement graph materialize API which given a subgraph (and possibly any graph) creates a fresh graph to which we can add properties and even save to files


py-raphtory/src/edge.rs
```diff
@@ -196,6 +197,14 @@ impl PyEdge {
         self.edge.static_property(name).map(|prop| prop.into())
     }
 
+    pub fn static_properties(&self) -> HashMap<String, Prop> {
```
miratepuffin, line 200: Can you add some quick doc strings to these

miratepuffin (approved):
If you can just comment the new functions in the API classes that would be fab

Issue #1002, opened 2023-06-01
Port `DynamicGraph` over to `raphtory` crate


ricopinazo:
DynamicGraph is being used as well in `raphtory-graphql` crate. For now it was just temporary copied over, but it should be in `raphtory` crate instead. Several things need to be done on this:
- Create `development` feature flag that includes `DynamicGraph` to be used by our own crates
- There are many struct members which are public only because we needed them to be for `DynamicGraph` implementation to work. Now that it is in the same crate, we can turn them back to private.
- Double check if when developing custom algorithms we need to have now `raphtory` as a dependency and consider re-exporting `DynamicGraph` from the python crate if that is the only struct we need from `raphtory`

ljeub-pometry:
This is irrelevant now that DynamicGraph no longer exists and python wrappers are part of the core crate (see #1055)

Issue #965, opened 2023-05-22
Clean up pyraphtory package


miratepuffin:

ljeub-pometry:
Cleaned up in #1065

PR #1055, opened 2023-06-19
Remove dynamic graph

ljeub-pometry:
- No need for the DynamicGraph struct anymore, replaced by a simple type definition.
- `raphtory-io` and `py-raphtory` integrated into `raphtory` crate as features `io` and `python`

Cleaning up and simplifying the code, reduce duplication.

no

the tests

- Maybe this works for GraphQL/js as well?
- Revisit and simplify the python wrappers now that we can actually implement `IntoPy` for all our type :)


ljeub-pometry:
This has increased in scope quite a bit now, `py-raphtory` and `raphtory-io` have been integrated into the core raphtory crate as features `python` and `io` respectively. This means that `IntoDynamic` and other traits can be properly reused without errors due to being external to crate and means that we should be able to implement all the appropriate
pyo3 traits for our types (this is still to be cleaned up!)

miratepuffin:
I think this closes #1002 and opens up the possibility of #965 -- just to link them

miratepuffin (approved):

PR #1063, opened 2023-06-22
Bug/partial windows

ljeub-pometry:
Fix the semantics for `expanding` and `rolling` such that partial windows at the end are returned

`expanding` and `rolling` are currently ignoring some of the data

Yes, documentation is tweaked to reflect the change (it was already outdated before)

Changed the tests for the new semantics


ricopinazo (approved):
LGTM

PR #1042, opened 2023-06-13
Remove shards

fabianmurariu:
### What changes were proposed in this pull request?
Take it apart then put it back together again

### Why are the changes needed?
remove shards

### Does this PR introduce any user-facing change? If yes is this documented?
probably

### How was this patch tested?
unit  integration

### Issues

_If this resolves any issues, please link to them here, the format is a KEYWORD followed by @<ISSUE NUMBER>__
KEYWORDS available are `close`, `closes`, `closed`, `fix`, `fixes`, `fixed`, `resolve`, `resolves`, `resolved`.
Please delete this text before creating your PR 

### Are there any further changes required?
I hope this is the last rewrite

PR #1080, opened 2023-06-29
Fix state post remove shards

fabianmurariu:
fix issues with finalize on for the task evaluation

return iterators for OUT and &A and rewrite finalize from them

No

unit tests

### Issues

closes #1042


miratepuffin (approved):
LGTM - though I wasn't involved in the discussion - assume @ljeub-pometry is happy as well as you @fabianmurariu

PR #1083, opened 2023-06-30
stop automatically reading hidden files

ricopinazo:


ljeub-pometry (approved):
Good fix for now, probably should have more advanced filtering support in the future.

PR #1091, opened 2023-07-06
Search edges

fabianmurariu:
for graphql search by edges
fix issue where we reindex everything on every request

unit tests

very likely


ricopinazo (approved):
LGTM

I think instead of having `search_edges` we potentially want to have that as a filter in the `edges` function, but as I understand that we want to force setting `limit` and `offset` parameters it's probably fine for now. Also, it is not trivial to combine other filters with search and make `limit` and `offset` work properly

PR #1095, opened 2023-07-07
Feature/python graphql

miratepuffin:
Exposed the Graphql server as a python submodule.

Makes it a lot easier for people to run Raphtory as a GraphQL service.
Makes it easier to test your pipeline (one script and no longer have to save your graphs)

Yes - full documentation of GraphQL to follow

New tests added for graphQL

Probably several other helper functions could be added


fabianmurariu (approved):
LGTM

fabianmurariu (changes requested):
This looks well but we need a way to manage the lifecycle of the server preferably via 
```python 
with run_server() as server:
 # use server
 
```

PR #1104, opened 2023-07-11
update the python loader to include polars dataframe

### LINKED ISSUE #936: Implement a dataframe loader for python/rust
Often much of the data we are working with in python is first manipulated as a dataframe in e.g. pandas - [Polars](http://pola-rs.github.io) provides a dataframe API in python and rust which we can piggyback off for ingestion in both languages.

We can also look at exporting results back to this format RE #910

fabianmurariu:
load a graph in Python from polars dataframe

to allow interop with various python data sources

### Issues

resolves #936


Haaroon:
this works in windows without my hack?

miratepuffin (approved):
LGTM - we need to create a separate ticket for splitting the properties into temporal and static, but can ship this for now :)

Issue #931, opened 2023-05-19
State object for algorithm results


miratepuffin:
Unless they are global and only return one result for the whole network, algorithms currently return a map which maps `vertex.name()` to the result. 

As an improvement to this, we should instead return a state object which allows you to query via id/name/vertex etc. This state should also be able to contain the results from multiple algorithms which are chained together.

PR #1105, opened 2023-07-11
Algorithms now have their own result object with many features

Haaroon:

Implemented a custom algorithm result object by which all algorithms that return a hashmap will now return instead

- Adds new algorithm result object 
 - result object contains the base hashmap generated by the algorithm, get, get all, sort, groupby methods
 - Wrap around all algorithms that return a hashmap 
- Fix all algorithm tests
- Add pandas export support
- Add python support
- Add python tests
- Fixed algorithms to use default arguments 
- - Add docs

yes, all docs updated

added tons of tests

### Issues

issue #931


raphtory/src/python/graph/algorithm_result.rs
```diff
@@ -0,0 +1,274 @@
...
+py_algorithm_result!(AlgorithmResultStrVecI64Str, String, Vec<(i64, String)>);
+py_algorithm_result_ord_hash_eq!(AlgorithmResultStrVecI64Str, String, Vec<(i64, String)>);
+
+py_algorithm_result!(AlgorithmResultU64VecUsize, u64, Vec<usize>);
+py_algorithm_result_ord_hash_eq!(AlgorithmResultU64VecUsize, u64, Vec<usize>);
+
+py_algorithm_result!(AlgorithmResultStrF64, String, OrderedFloat<f64>);
+
+#[pymethods]
+impl AlgorithmResultStrF64 {
```
miratepuffin, line 155: Why not make a separate macro for this where the key can be anything

miratepuffin (approved):
LGTM, couple of small things, but then we can ship

raphtory/src/algorithms/algorithm_result.rs
```diff
@@ -0,0 +1,390 @@
...
+        H: Clone + Hash + Eq + Ord,
+        Y: Clone + Ord + Hash + Eq
+{
+    /// Groups the `AlgorithmResult` by its values.
+    ///
+    /// # Returns
+    ///
+    /// A `HashMap` where keys are unique values from the `AlgorithmResult` and values are vectors
+    /// containing keys of type `H` that share the same value.
+    pub fn group_by(
```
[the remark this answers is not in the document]
Haaroon: idk about whether group by can return an algo result
because then in the macro, the macro is returning itself
which it cannot do, because its building itself?

raphtory/src/python/graph/algorithm_result.rs
Haaroon: because we cant have generic types in pyo3

PR #1116, opened 2023-07-17
fix subgraph neighbours returning duplicates

ljeub-pometry:

the neighbours for subgraph are no longer duplicate

triangle count for subgraph and original graph now matches

more tests for subgraphs


fabianmurariu (approved):
LGTM

PR #1117, opened 2023-07-17
Feature/python index graph

miratepuffin:

Exposed the tantivity index for search edges and vertices to python 

Allows us to easily experiment with these features and test what is useful to users.

Yes and yes

Several tests added - many bugs found which need to be fixed as we explore this functionality more. 

Several fixes to be made on layer searching, fuzzy searching


ricopinazo (approved):
LGTM

PR #1122, opened 2023-07-17
Bug/compute state

### LINKED ISSUE #1121: The compute state does not play nice with subgraphs
The compute state currently assumes that all the vertices are always there and runs with the first graph.num_vertices()  vertices of the graph. If you have a subgraph, this means we run the computation with vertices that are not supposed to be in the subgraph and exclude vertices that are supposed to be in the subgraph leading to very wrong results.

fabianmurariu:
fix the eval code ignoring filtering graphs 

unit tests
### Issues

closes #1121


miratepuffin (approved):
@narnolddd tested and is all working

PR #1123, opened 2023-07-17
Feature kcore

narnolddd:
Add a k-core algorithm for getting vertices with degree >= k recursively
Improved triangle counting algorithm which runs 2-core first 

Handy algorithm in any case, but useful for optimising subgraph counts 

No

Unit tests


raphtory/src/algorithms/k_core.rs
```diff
@@ -0,0 +1,154 @@
...
+pub fn k_core_set<G>(graph: &G, k: usize, iter_count: usize, threads: Option<usize>) -> HashSet<VID>
+where
+    G: GraphViewOps,
+{
+    let ctx: Context<G, ComputeStateVec> = graph.into();
+
+    let step1 = ATask::new(move |vv| {
+        let deg = vv.degree();
+        let state: &mut KCoreState = vv.get_mut();
+        state.alive = if deg < k { false } else { true };
```
ljeub-pometry, line 51: this is just `deg >= k` (no need for the if statement)

raphtory/src/algorithms/k_core.rs
```diff
@@ -0,0 +1,154 @@
...
+                .filter(|b| *b)
+                .count()
+                >= k;
+            let state: &mut KCoreState = vv.get_mut();
+            if current != prev {
+                state.alive = current;
+                Step::Continue
+            } else {
+                Step::Done
+            }
```
ljeub-pometry, line 71: only need to do this if `prev == true`, this is doing a lot of unnecessary work (once a node is dead it is not coming back)

raphtory/src/algorithms/k_core.rs
```diff
@@ -0,0 +1,154 @@
...
+            },
+            threads,
+            iter_count,
+            None,
+            None,
+        )
+        .into_iter()
+        .filter(|(k, v)| *v)
+        .map(|(k, v)| k)
+        .collect()
```
ljeub-pometry, line 97: you don't need to make a temporary `HashMap` and filter on it again. Just do the filter in the closure directly and return the set.

raphtory/src/algorithms/k_core.rs
```diff
@@ -0,0 +1,154 @@
...
+/// * `g` - A reference to the graph
+/// * `k` - Value of k such that the returned vertices have degree > k (recursively)
+/// * `iter_count` - The number of iterations to run
+/// * `threads` - number of threads to run on
+///
+/// # Returns
+///
+/// A hash set of vertices in the k core
+///
+pub fn k_core_set<G>(graph: &G, k: usize, iter_count: usize, threads: Option<usize>) -> HashSet<VID>
```
ljeub-pometry, line 42: This probably should return `VertexSubgraph<G>`

ljeub-pometry (changes requested):
Very useful, just some minor improvements and low-hanging optimisations.

raphtory/src/algorithms/k_core.rs
narnolddd: nice, thanks!

raphtory/src/algorithms/k_core.rs
narnolddd: very nice spot, fixing now.

raphtory/src/algorithms/k_core.rs
narnolddd: Yea, I have added just now as a separate function just in case you would want just the set instead of the graph but I agree :)

raphtory/src/algorithms/k_core.rs
narnolddd: Ahh of course 🤦‍♀️ thanks got it!

raphtory/src/algorithms/k_core.rs
```diff
@@ -0,0 +1,154 @@
...
+#[derive(Clone, Debug)]
+struct KCoreState {
+    alive: bool,
+}
+
+impl KCoreState {
+    fn new() -> Self {
+        Self { alive: true }
+    }
+}
```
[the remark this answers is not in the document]
narnolddd: was just trying this but ended up in borrowing hell 😅 so may leave for now, but happy to come back to it if there are any issues

raphtory/src/algorithms/k_core.rs
```diff
@@ -48,24 +49,28 @@ where
...
-            let state: &mut KCoreState = vv.get_mut();
-            if current != prev {
-                state.alive = current;
-                Step::Continue
+            if prev == true {
+                let current = vv
+                    .neighbours()
+                    .into_iter()
+                    .map(|n| n.prev().alive)
+                    .filter(|b| *b)
```
ljeub-pometry, line 64: I missed this before, this is just `filter(|n| n.prev().alive)`, no need to map first

ljeub-pometry (approved):
Found one more simplification that I missed before but otherwise this is looking good!

PR #1124, opened 2023-07-18
Feature/layering in pandas

miratepuffin:
Enabled the user to set layers in the pandas edge loader - this can be as a column in the dataframe or as an explicit value

Moves towards making the pandas loader feature complete ~ in parity with manual loading

This will be documented alongside the full pandas docs.

In examples by @rachchan 

Need to add functionality for static properties and deletion columns.


fabianmurariu (approved):
LGTM

PR #1126, opened 2023-07-19
Adds a format checker to the workflows, which fails if the code has not been formatted, adds a dev folder with instructions to setup a nightly env + point the git hooks to the local folder

### LINKED ISSUE #1085: Add `cargo fmt --check` to CI

Haaroon:

Adds a format checker to the workflows, which fails if the code has not been formatted
- checks out nightly cargo in a workflow
- runs the format check
- add a new `dev` folder with a bootstrap.sh file that setups hooks and the dev environment 

prevent dirty formatted code

no

both by hand and in the workflow 

### Issues

resolves #1085 

no


Haaroon:
yeah sorry i wrote the wrong commit message 🙃

Haaroon:
this run should fail because i did not format the code.

Haaroon:
ready

dev/bootstrap.sh
```diff
@@ -0,0 +1,23 @@
...
+  echo "$(basename ${BASH_SOURCE[0]}): $@"
+}
+
+install_hooks() {
+  git config core.hooksPath \
+    || git config core.hooksPath ./dev/hooks
+}
+
+install_rust_nightly() {
+  rustup toolchain install nightly
```
fabianmurariu, line 17: why are we installing nightly?

dev/hooks/pre-commit
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# Run rustfmt on the whole repository using nightly toolchain
+cargo +nightly fmt --all -- --check
```
fabianmurariu, line 4: why nightly?

dev/bootstrap.sh
Haaroon: so that uses can run the formatter, otherwise it will fail to run in the hook.

dev/hooks/pre-commit
Haaroon: because it has the features for formatting imports which stable does not have

PR #1132, opened 2023-07-24
Feature/expanded edges

iamsmkr:
Implements expanded nodes feature wherein given a list of nodes, a list of all the edges that are 2 hops aways but are related to the existing nodes in the graph are returned.


miratepuffin (approved):
We need to modify this slightly once the new edge layering is merged tomorrow and add some tests to ensure the output it correct, but otherwise looks good to me.

PR #1149, opened 2023-08-03
Feature/multi edge support

### LINKED ISSUE #1138: Cannot add edge/vertex updates at the same time
Currently if I add two edge/vertex updates at exactly the same time Raphtory will eat the second update, not adding it into the model and won't let the user know this has happened. This means a simple check after ingesting some snapshots will lead to the user finding fewer edges in the graph than in the original data.

We need to find a manner to store this information - similar to the secondary time index we had in old Raphtory.

### LINKED ISSUE #1161: Calling add_edge in python with dict or list input deadlocks
```
from raphtory import Graph
g = Graph()
g.add_edge(1, 1, 3, {"test": {"mapped": 1}})
```
dead-locks on the last line and never returns

ljeub-pometry:

add secondary index (by default a running count of insertion events) to support multiple edge events with the same timestamp

avoid potential data loss for data sets with coarse timestamps

### Issues

fixes #1138 
fixes #1161

### DIFF
--- a/examples/rust/src/bin/btc/main.rs
+++ b/examples/rust/src/bin/btc/main.rs
@@ -97,7 +97,7 @@ fn main() {
                     [("amount".to_string(), Prop::U64(sent.amount_btc))],
                     None,
                 )
-                .unwrap()
+                .unwrap();
             })
             .expect("Failed to load graph from CSV data files");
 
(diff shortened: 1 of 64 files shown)


miratepuffin (changes requested):
Couple of notes from the call:

Make issue for count of exploded edges
Make issue for exposing windowing on secondary time index/exposing secondary time index to users
Make issue for finalising API for custom secondary time index
Quick look at add_edge returning the EdgeRef
Change _t functions to explode

ljeub-pometry:
Opened #1162, #1163, #1164 as per comment above.

Issue #1140, opened 2023-07-27
GraphQL server mutability


miratepuffin:
Currently the graphQL server once run has no way of changing the data/graphs it serves - we would like to be able to both:

- Add new vertex/edge updates via mutation queries 
- Add whole new graphs from either a path on disk or sent across the wire. 

If a graph previously existed with the same name it should be rejected unless a replace/force flag is set.

PR #1168, opened 2023-08-10
Make GraphQL server mutable

ljeub-pometry:

Add mutation to the GraphQLServer that allows reading new graphs from directory. Creates the foundation for implementing #1140 

No longer need to restart the server to load new data

yes, new method has docstring.

manually, we should add some actual tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2822,6 +2822,7 @@ dependencies = [
  "opentelemetry",
  "opentelemetry-jaeger",
  "ordered-float 3.7.0",
+ "parking_lot",
  "poem",
  "raphtory",
  "serde",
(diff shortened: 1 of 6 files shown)


ricopinazo (approved):
LGTM

miratepuffin:
- Need to test on a windows machine to see what the error is (seems to be something to do with generation of file paths)
- Going to test to see if we can create new graphs in WASM and get access to the graph bincode
    - If bincode is easy enough to get hold of, add some capability to the graphql server to directly receive saved raphtory graphs and add them into it's loaded graphs
- If the above point is NOT easy to do we have two options:
    - Enable the UI to save graphs to disk and point the local server at these files (wrapped in a function call implying that is being sent to the server) - this will allow UI devs to build/test functionality whilst the 'sending to server' functionality is worked on and can be swapped in as soon as it is working.
    - Serialise the whole graph and send it to the GraphQL server as JSON

PR #1172, opened 2023-08-11
add graph schema to graphql

ricopinazo:

This PR adds a new function to the GraphQL server to obtain the graph schema. An example query might look as follows:

```
{
  graph(name: "jira") {
    schema {
      nodes {
        typeName,
        properties {
          key,
          variants
        }
      }
    },
  }
}
```

Additionally, it changes the semantics of `PropertyHasFilter` so that `key` and `valueStr`/`valueNumber` need to refer to the same property and adds a function `layerNames` which returns of the layer names (this should be added to the schema itself in the future)

This changes ease using raphtory GraphQL server to build other systems, such as UIs

Yes, some docs for the new functionality was added, as well as some docs that were missing

There are tests missing for the new function, but it was tested manually

We should consider adding information for the edges as well in a new future


iamsmkr (approved):
LGTM

PR #1196, opened 2023-08-21
add schema for layers and edges

ricopinazo:

A graphql query to get the full schema of the graph would look like this now:
```
{
    graph(name: $graphName) {
        schema {
            nodes {
                typeName,
                properties {
                    key,
                    variants
                }
            },
            layers {
                name,
                edges {
                    srcType,
                    dstType,
                    properties {
                        key,
                        variants
                    }
                }
            }
        }
    }
}
```

We need this to improve integration of raphtory with other tools

Yes, docs were added to the new functions

It was manually tested

Yes, tests should be added to this at some point


raphtory-graphql/src/model/mod.rs
```diff
@@ -10,11 +10,11 @@ use crate::{
...
 };
-use async_graphql::Context;
+use async_graphql::{Context, EmptyMutation, EmptySubscription};
 use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine};
 use dynamic_graphql::{
-    App, Mutation, MutationFields, MutationRoot, ResolvedObject, ResolvedObjectFields, Result,
-    Upload,
+    dynamic::Schema, App, Mutation, MutationFields, MutationRoot, ResolvedObject,
+    ResolvedObjectFields, Result, Upload,
 };
```
ljeub-pometry, line 18: why did the imports change? doesn't seem like anything else changed here

raphtory-graphql/src/model/schema/mod.rs
```diff
@@ -1,3 +1,41 @@
...
+fn get_vertex_type<G: GraphViewOps>(vertex: VertexView<G>) -> String {
+    let prop = vertex.properties().get("type");
+    prop.map(|prop| prop.to_string())
+        .unwrap_or_else(|| "NONE".to_string())
+}
+
+type SchemaAggregate = HashMap<String, HashSet<String>>;
+
+fn merge_schemas(s1: SchemaAggregate, s2: SchemaAggregate) -> SchemaAggregate {
+    let mut merged_map = s1.clone();
```
ljeub-pometry, line 24: could you not just mark the input argument as mut? you are taking ownership already

ljeub-pometry (approved):
This looks good, though I think you may have introduced
some unused imports?

PR #1197, opened 2023-08-22
Pandas loading bars

miratepuffin:
We have added tqdm style loading bars to the Pandas ingestors 
This makes it clearer how long it is going to take to ingest your data - i.e. do I need to put on a cup of tea
Yes, but nothing that affects the API's
Manual testing as it is visuals

We need to work out how to get this working in jupyter


ljeub-pometry (approved):
Looks good

Issue #1038, opened 2023-06-09
Expand the filtering in Graphql


miratepuffin:
As from #1037 we have some basic filtering of nodes in graphQL - selecting in/out degree less than/greater than a given value and checking the value of names/node type.

This can be expanded in multiple ways - to name a few:

- The edges, and neighbours could have the same filters available. 
- The user could also be able to filter all of these types by properties.

PR #1203, opened 2023-08-28
feat(graphql): Extend property filtering to edges in graphql

Dullaz:

Add property filtering to EdgeFilter for GraphQL.

Allows users to filter edges using the PropertyHasFilter

Yes. Users can now use the PropertyHasFilter in the EdgeFilter in GraphQL. It shows up automatically in the GraphQL schema, but I'm not sure if there's somewhere else that needs to be updated.

Manually :P.  TBD

### Issues
Does not resolve, but contributes to:
https://github.com/Pometry/Raphtory/issues/1038

1. Note sure I like doing `matches_node_properties` and `matches_edge_properties` but I don't know enough about rust to do it with traits yet.

### DIFF
--- a/raphtory-graphql/src/model/filters/property.rs
+++ b/raphtory-graphql/src/model/filters/property.rs
@@ -1,8 +1,11 @@
-use crate::model::{filters::primitives::NumberFilter, graph::node::Node};
+use crate::model::{
+    filters::primitives::NumberFilter,
+    graph::{edge::Edge, node::Node},
+};
 use dynamic_graphql::InputObject;
 use raphtory::{core::Prop, db::api::view::VertexViewOps};
 
-#[derive(InputObject)]
+#[derive(InputObject, Clone)]
 pub(crate) struct PropertyHasFilter {
     key: Option<String>,
     value_str: Option<String>,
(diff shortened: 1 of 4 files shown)


Dullaz:
I think there's a nicer way of doing this with traits but I don't know what the overlap is between an Edge and Node, or what other models need to support property filtering.

Happy to update this with a bit of guidance :)

raphtory-graphql/src/model/filters/property.rs
```diff
@@ -24,6 +27,23 @@ impl PropertyHasFilter {
             None => node.vv.properties().values().any(valid_prop),
         };
     }
+
+    pub(crate) fn matches_edge_properties(&self, edge: &Edge) -> bool {
+        let valid_prop = |prop| valid_prop(prop, &self.value_str, &self.value_num);
+
+        return match &self.key {
+            Some(key) => {
+                if let Some(prop) = raphtory::prelude::EdgeViewOps::properties(&edge.ee).get(key) {
```
miratepuffin, line 36: @Dullaz if you just import raphtory::prelude::* at the top this should be a good bit cleaner and I can merge :)

miratepuffin (changes requested):

Dullaz:
@miratepuffin should be ready now, i changed the way i imported property_filter in edge_filter too

miratepuffin (approved):

PR #1207, opened 2023-08-29
Bug/windowed output

ljeub-pometry:

Fix the output for algorithms that use local state such that they don't return results for vertices that are not in the graph

### DIFF
--- a/raphtory/src/algorithms/motifs/three_node_temporal_motifs.rs
+++ b/raphtory/src/algorithms/motifs/three_node_temporal_motifs.rs
@@ -21,6 +21,7 @@ use crate::{
     },
 };
 
+use crate::core::entities::vertices::vertex_ref::VertexRef;
 use itertools::enumerate;
 use num_traits::Zero;
 use rustc_hash::FxHashSet;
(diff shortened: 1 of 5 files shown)


raphtory/src/algorithms/motifs/three_node_temporal_motifs.rs
```diff
@@ -387,8 +388,10 @@ pub fn triangle_motifs<G: GraphViewOps>(
         |_, _, _els, local| {
             let mut tri_motifs = HashMap::new();
             for (vref, mc) in enumerate(local) {
-                let v_gid = graph.vertex_name(vref.into());
-                tri_motifs.insert(v_gid.clone(), mc.triangle);
+                if graph.has_vertex_ref(VertexRef::Local(vref.into())) {
```
Alnaimi-, line 391: Is it possible to make the enumerate function not return the vref if it doesn't exist in the local graph? Maybe we need to different enumerate functions for local

miratepuffin: That is the Ticket @Haaroon has allocated to him

miratepuffin (approved):

PR #1211, opened 2023-08-31
Changing property type now returns an error instead of silently ignoring the value

### LINKED ISSUE #951: Properties which change type need to throw an error
Currently in python if you add properties to an edge/vertex multiple times with the same key but different value types, only values of the first type are stored, for example:

```python
g.add_vertex(1,1,{"prop1": 1})
g.add_vertex(2,1,{"prop1": "string 1"})
g.add_vertex(3,1,{"prop1": 3})

g.vertex(1).property_history("prop1")
```
```
[(1,1),(3,3)]
```

The resulting vertex will only have the integer values not the string in-between.

This is erroring in the core, but the exception is being swallowed - we either need to make it possible to have properties be multiple value types or at least throw the error in python so that the user knows this data is missing.

ljeub-pometry:

Added a test for all temporal property additions

### Issues
fixes #951

### DIFF
--- a/raphtory/src/core/entities/properties/tprop.rs
+++ b/raphtory/src/core/entities/properties/tprop.rs
@@ -5,6 +5,7 @@ use crate::{
             LayerIds,
         },
         storage::{locked_view::LockedView, timeindex::TimeIndexEntry},
+        utils::errors::GraphError,
         Prop,
     },
     db::graph::graph::Graph,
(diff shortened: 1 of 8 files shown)


raphtory/src/core/entities/properties/tprop.rs
```diff
@@ -52,74 +53,54 @@ impl TProp {
...
-                *self = TProp::from(t, prop);
-            }
-            TProp::Str(cell) => {
-                if let Prop::Str(a) = prop {
-                    cell.set(t, a.to_string());
+    pub(crate) fn set(&mut self, t: TimeIndexEntry, prop: Prop) -> Result<(), GraphError> {
+        if matches!(self, TProp::Empty) {
+            *self = TProp::from(t, prop);
+        } else {
+            // TODO: we should bubble up the result if we change the type of the property
```
miratepuffin, line 60: This comment can go now

miratepuffin (approved):

raphtory/src/core/entities/properties/tprop.rs
ljeub-pometry: Indeed...

miratepuffin (approved):

PR #1228, opened 2023-09-05
Improved property additions api for vertices and edges

### LINKED ISSUE #1217: Update constant property functions
Going to make some small API tweaks:
We need to rename add_static_properties to add_constant_properties for the graph (as we refer to them as constant everywhere else)
I think it makes sense for us to change:
```python
g = Graph()
g.add_vertex(1,1)
g.add_edge(1,1,2)
g.add_vertex_properties({"some vertex data":1)
g.add_edge_properties({"some edge data":2)
```
To:
```python
g = Graph()
v = g.add_vertex(1,1)
e = g.add_edge(1,1,2)
v.add_constant_properties({"some vertex data":1)
e.add_constant_properties({"some edge data":2)
```
We require the vertex/edge to exist before adding constant properties and we throw an exception if they don’t - this should make the requirement explicit
Should also add:
```
g = Graph()
v = g.add_vertex(1,1)
e = g.add_edge(1,1,2)
v.add_properties(timestamp=2, properties={"some vertex data":1)
e.add_properties(timestamp=3, properties="some edge data":2)
```

ljeub-pometry:

Constant properties for vertices and edges are now added from the vertex or edge view directly rather than from the graph. This makes the assumption that the vertex and edge exists before adding properties explicit. Also added convenience api on vertex and edge for adding more updates (this is actually more efficient than using the graph as we can skip the id translation)

Yes, basic docstrings are there but the examples also need to be updated

updated existing tests to use new apis

### Issues

resolves #1217 

enum dispatch for the graph view will make the python side better behaved. Right now `neighbours` and `edges` don't return mutable views which may be surprising. `vertex` and `edge` are manually overridden to make them work correctly.


miratepuffin (approved):
Looks so much cleaner!

Issue #1189, opened 2023-08-17
Number of transactions


BaCk7:
There should be a method to extract the number of transactions in the graph, without leveraging the history() functions

miratepuffin:
Solved via #1241

Issue #1010, opened 2023-06-01
Feature: Temporal edge count 


miratepuffin:
We have `graph.num_edges()` which returns the count of unique edges within the view - this issue would create a `graph.num_temporal_edges()` which would do the same thing but for exploded edges.

PR #1241, opened 2023-09-05
Feature/temporal edges

narnolddd:

Temporal edges function which counts number of exploded edges.

Is a fairly important metric to have for temporal edges

Made docstrings for both.

Small unit test in graph.

### Issues

Solves #1189 and #1010 I believe!

Could probably be implemented more efficiently using internals.


ljeub-pometry (approved):
This looks good as a first implementation, we can indeed optimise this a lot internally but the top-level api is fine and implementation looks correct.

PR #1259, opened 2023-09-12
Support for writing external pyo3-based algorithm extensions

ljeub-pometry:

- Make it possible to pass a python raphtory graph across crate boundary using a bincode roundtrip
- Add methods to get bincode without saving to file

Trying to use a python graph created by a different python extension built against raphtory fails with a type error without this change

adds `bincode` methods 

added a dummy extension example with tests

Maybe there is a better way to do this without requiring serialising and deserialising the graph?


miratepuffin (approved):
Wicked. We should probably also add a test for the custom graphql function and add it to the same project - can then delete the one in examples.

PR #1269, opened 2023-09-18
String deduplication

ljeub-pometry:

- Store all strings as immutable Arc<str> to make cloning of properties cheap
- Implement a per-graph string pool to avoid storing duplicate string values and instead only store pointers to the same underlying string

- memory optimisation

no, changes are completely transparent to the user-facing apis

existing tests still work and added a test to check the deduplication is effective

### DIFF
--- a/raphtory/src/core/mod.rs
+++ b/raphtory/src/core/mod.rs
@@ -27,7 +27,15 @@
 use crate::db::{api::view::GraphViewOps, graph::graph::Graph};
 use chrono::NaiveDateTime;
 use serde::{Deserialize, Serialize};
-use std::{cmp::Ordering, collections::HashMap, fmt, sync::Arc};
+use std::{
+    borrow::Borrow,
+    cmp::Ordering,
+    collections::HashMap,
+    fmt,
+    fmt::{Display, Formatter},
+    ops::Deref,
+    sync::Arc,
+};
 
 #[cfg(test)]
 extern crate core;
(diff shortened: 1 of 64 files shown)


raphtory/src/core/mod.rs
```diff
@@ -37,6 +45,65 @@ pub mod state;
...
+    }
+}
+
+impl From<ArcStr> for String {
+    fn from(value: ArcStr) -> Self {
+        value.to_string()
+    }
+}
+impl Deref for ArcStr {
+    type Target = Arc<str>;
```
fabianmurariu, line 70: Isn't this `str` ?

ljeub-pometry: My thinking was that it is wrapping an `Arc<str>` so you should be able to do anything that you can do with an Arc. Derek coercion should chain the calls to `deref()` for you so this should still act like it implements `Deref<Targe=str>` since that is implemented for `Arc<str>`.

fabianmurariu: fair

fabianmurariu (approved):
Apart from the serialisation memory issue, which comes with it's own set of problems this LGTM

Haaroon:
I too approve, with large files i have seen a reduction of 15G in ram, a previous file set (netflow and logs on day2) i saw 55G of raphtory ram usage, now with these changes the graph uses 40.7G of ram

Haaroon (approved):

PR #1271, opened 2023-09-19
Archive

iamsmkr:
Gql api to archive/unarchive graph

Same as above

Yes

Tests

### Issues

No


raphtory-graphql/src/model/mod.rs
```diff
@@ -143,6 +146,63 @@ impl Mut {
...
+
+            let static_props_without_name: Vec<(String, Prop)> = subgraph
+                .properties()
+                .into_iter()
+                .filter(|(a, b)| a != "name")
+                .collect_vec();
+
+            new_subgraph
+                .add_constant_properties(static_props_without_name)
+                .expect("Failed to add static properties");
```
miratepuffin, line 178: I don't think we want a load of expects in here. We shouldn't get an issue, but best to handle it

raphtory-graphql/src/model/mod.rs
```diff
@@ -185,17 +245,51 @@ impl Mut {
...
+            // graph_name == new_graph_name, means its a "save" and not "save as" action
+            if graph_name.ne(&new_graph_name) {
+                let static_props: Vec<(String, Prop)> = subgraph
+                    .properties()
+                    .into_iter()
+                    .filter(|(a, b)| a != "name" && a != "creationTime" && a != "uiProps")
+                    .collect_vec();
+                new_subgraph
+                    .add_constant_properties(static_props)
+                    .expect("Failed to add static properties");
```
miratepuffin, line 264: Same as above

miratepuffin (changes requested):
All looks good, but can we change out the expects and it would cause the whole server to crash?

miratepuffin (approved):

PR #1273, opened 2023-09-20
Motif improvements

narnolddd:

functionality for running multiple deltas on motif counts in a single run. also for choosing tie breaking mechanism from time + index and random reordering within block

before this, doing multiple deltas separately incurred a lot of wasted time in identifying static motifs over and over for each delta choice, this is hopefully a bit more optimised for that.

no, the original algorithm can be used before (it calls a special case of the new more general algo).

 single delta test still works.

Some parts seem not very optimal e.g. the sorting function is decided in the top level algorithm but then each different motif count has to do a match on the sorting type given (this matching then is done for every vertex), this was the only way I could get it working but feels like there should be a better way. Also the code feels a bit ugly. If anyone has some suggestions on short hand +/or optimisation, that would be amazing!


raphtory/src/algorithms/motifs/three_node_temporal_motifs.rs
```diff
@@ -127,54 +113,82 @@ pub fn star_motif_count<G: GraphViewOps>(
...
-    star_count.return_counts()
+
+    deltas
+        .into_iter()
+        .map(|delta| {
+            let mut star_count = init_star_count(evv.degree());
+            star_count.execute(&events, delta);
+            star_count.return_counts()
+        })
+        .collect::<Vec<[usize; 24]>>()
```
ljeub-pometry, line 124: You are redoing the count separately for all the deltas here and in other places but the results are nested as a motif for small delta is still there for large delta, right? Should be possible to use that to speed up the computation a lot as you can save many comparisons.

ljeub-pometry (approved):
This looks good but I think you are missing a trick which should speed up the multi-delta computation. The results should be nested as a motif for a small delta is still a motif for large delta. If you were to compute the results across all values of delta in one loop instead of restarting from scratch for each new delta value (and ensure that the deltas are sorted), you should be able to save on a lot of comparison operations.

PR #1319, opened 2023-10-04
Fix bug with tqdm not running in jupyter

miratepuffin:
updated to the new version of kdam(tqdm)
means the loading bars are now available within jupyter notebooks and that the standard out doesn't fill up/hang
No
With my eyes in vscode, jupyter and ipython
Need to upgrade this to 0.5.0 when it releases

### DIFF
--- a/raphtory/Cargo.toml
+++ b/raphtory/Cargo.toml
@@ -43,7 +43,7 @@ lock_api = { version = "0.4", features = ["arc_lock", "serde"] }
 dashmap = {version ="5", features = ["serde"] }
 serde_with = "3.3.0"
 enum_dispatch = "0.3"
-kdam = "0.4.1"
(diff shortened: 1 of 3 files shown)


raphtory/Cargo.toml
```diff
@@ -43,7 +43,7 @@ lock_api = { version = "0.4", features = ["arc_lock", "serde"] }
 dashmap = {version ="5", features = ["serde"] }
 serde_with = "3.3.0"
 enum_dispatch = "0.3"
-kdam = "0.4.1"
+kdam = { git = "https://github.com/clitic/kdam.git", branch = "pyo3", features = ["notebook"] }
```
Haaroon, line 46: You cannot publish a version onto cargo without a version number

miratepuffin: I have just asked the other dev to publish 0.5.0 of the lib so I can update this

PR #1322, opened 2023-10-06
Efficient version for global motif counts

narnolddd:
Array accumulators (thank you @fabianmurariu !) and global motif counting algo

Other algorithm was storing results per vertex in memory till right at the end, this is a massive memory hog unless you particularly care about the local results, especially if you're running with multiple time deltas

TBC

Small unit test

need to expose to python + add back in the array acc test


narnolddd:
Also just edited some python docs for balance that were referring to rust objects.

miratepuffin (approved):

PR #1325, opened 2023-10-06
Added closeness centrality

miratepuffin:

- Updated dijkstra's be able to run in any direction and without weights

Algorithms was missing
Yes and documented
All changes to dijkstras and closeness have new tests


Haaroon (changes requested):
requires docs

Issue #1305, opened 2023-09-28
Add the karate club graph generator


Haaroon:
Graph named “Zachary’s Karate Club” with 34 nodes and 78 edges

PR #1327, opened 2023-10-09
Vertex Names Returned in New Algorithm Object

Haaroon:
### Issues

### DIFF
--- a/raphtory/src/algorithms/algorithm_result_new.rs
+++ b/raphtory/src/algorithms/algorithm_result_new.rs
@@ -1,0 +1,597 @@
+use crate::{
+    core::entities::vertices::vertex_ref::VertexRef,
+    prelude::{Graph, GraphViewOps, VertexViewOps},
+};
+use itertools::Itertools;
+use num_traits::Float;
+use ordered_float::OrderedFloat;
+use std::{
+    borrow::Borrow,
+    collections::{hash_map::Iter, HashMap},
+    fmt,
+    fmt::Debug,
+    hash::Hash,
+    marker::PhantomData,
+};
+
+pub trait AsOrd<T: ?Sized + Ord> {
+    /// Converts reference of this type into reference of an ordered Type.
+    ///
+    /// This is the same as AsRef (with the additional constraint that the target type needs to be ordered).
+    ///
+    /// Importantly, unlike AsRef, this blanket-implements the trivial conversion from a type to itself!
+    fn as_ord(&self) -> &T;
+}
+
+impl<T: Ord> AsOrd<T> for T {
+    fn as_ord(&self) -> &T {
+        self
+    }
+}
+
+impl<T: Float> AsOrd<OrderedFloat<T>> for T {
+    fn as_ord(&self) -> &OrderedFloat<T> {
+        self.into()
+    }
+}
+
+impl<T: Float> AsOrd<(OrderedFloat<T>, OrderedFloat<T>)> for (T, T) {
+    fn as_ord(&self) -> &(OrderedFloat<T>, OrderedFloat<T>) {
+        // Safety: OrderedFloat is #[repr(transparent)] and has no invalid values, i.e. there is no physical difference between OrderedFloat and Float.
+        unsafe { &*(self as *const (T, T) as *const (OrderedFloat<T>, OrderedFloat<T>)) }
+    }
+}
+
+/// An 'AlgorithmRepr' struct that represents the string output in the terminal after running an algorithm.
+///
+/// It returns the algorithm name, number of vertices in the graph, and the result type.
+///
+pub struct AlgorithmRepr {
+    pub algo_name: String,
+    pub result_type: String,
+}
+
+/// A generic `AlgorithmResult` struct that represents the result of an algorithm computation.
+///
+/// The `AlgorithmResult` contains a hashmap, where keys (`H`) are cloneable, hashable, and comparable,
+/// and values (`Y`) are cloneable. The keys and values can be of any type that satisfies the specified
+/// trait bounds.
+///
+/// This `AlgorithmResult` is returned for all algorithms that return a HashMap
+///
+pub struct AlgorithmResult<G, V, O = V> {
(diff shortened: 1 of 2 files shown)


Haaroon:
resolves #1305

raphtory/src/algorithms/metrics/balance.rs
```diff
@@ -113,22 +113,26 @@ pub fn balance<G: GraphViewOps>(
...
-            |_, ess, _, _| ess.finalize(&min, |min| min),
-            threads,
-            1,
-            None,
-            None,
-        ),
-    )
+    let mut map: HashMap<usize, f64> = HashMap::new();
+    for (vertex_name, value) in runner_result.iter() {
+        if let Some(vertex) = graph.vertex(vertex_name.to_string()) {
```
miratepuffin, line 129: We need to change ess to return internal ID  so that we can remove this map

raphtory/src/algorithms/community_detection/connected_components.rs
```diff
@@ -273,35 +275,44 @@ mod cc_test {
...
-            assert_eq!(
-                actual,
-                Some((*smallest, edges.len())),
-                "actual: {:?}",
-                actual
-            );
         }
     }
+
+    // #[quickcheck]
```
miratepuffin, line 281: Let's look at this together on monday

raphtory/src/algorithms/metrics/balance.rs
```diff
@@ -169,32 +173,32 @@ mod sum_weight_test {
...
-            ("4".to_string(), -15.0),
-            ("5".to_string(), 0.0),
+            ("1".to_string(), Some(-32.0)),
+            ("2".to_string(), Some(-5.0)),
+            ("3".to_string(), Some(-3.0)),
+            ("4".to_string(), Some(-15.0)),
+            ("5".to_string(), Some(0.0)),
         ];
-        assert_eq!(res.sort_by_key(false), expected);
+        assert_eq!(res.sort_by_vertex_id(false), expected);
```
miratepuffin, line 202: Lets go over these names on monday as well

raphtory/src/algorithms/algorithm_result.rs
```diff
@@ -97,39 +103,65 @@ where
...
+            .collect_vec()
     }
 
-    /// Sorts the `AlgorithmResult` by its keys in ascending or descending order.
+    /// Returns a hashmap with vertex names and values
+    ///
+    /// Returns:
+    ///     a hashmap with vertex names and values
+    pub fn get_with_names(&self) -> HashMap<String, Option<V>> {
+        let mut as_map = HashMap::new();
```
miratepuffin, line 141: You can do the same as above, but with collect<Hashmap<>>::

raphtory/src/algorithms/algorithm_result.rs
```diff
@@ -55,32 +56,37 @@ pub struct AlgorithmRepr {
 ///
 /// This `AlgorithmResult` is returned for all algorithms that return a HashMap
 ///
-pub struct AlgorithmResult<K, V, O = V> {
+pub struct AlgorithmResult<G, V, O = V> {
```
miratepuffin, line 59: I believe there are a couple of things we need to fix in here:

- For all of the functions by default they should no longer return a string, instead they should return a vertex object. This doing a lot of extra work that we are 'not being asked for' and removes the actual planned functionality. 
- We should return to the old function names for the above, but add an `_names` variation which would map the vertex objects to their name/id
- We need to run some benchmarks on some larger graphs to make sure these changes aren't adding a load of overhead.

miratepuffin (changes requested):
We need to go through some of these return types and extra for loops we have added as I think they are adding quite a bit of extra work. Can pair on this on monday.

raphtory/src/algorithms/centrality/degree_centrality.rs
```diff
@@ -42,22 +42,25 @@ pub fn degree_centrality<G: GraphViewOps>(
...
+        vec![Job::new(step1)],
+        None,
+        |_, ess, _, _| ess.finalize(&min, |min| min),
+        threads,
+        1,
+        None,
+        None,
+    );
+    let mut map: HashMap<usize, f64> = HashMap::new();
+    for (vertex_name, value) in runner_result.iter() {
```
Haaroon, line 56: Turn into map and collect rather than for loop

raphtory/src/algorithms/metrics/balance.rs
Haaroon: rename to sort_by_vertex

raphtory/src/algorithms/community_detection/connected_components.rs
Haaroon: fixed

miratepuffin (changes requested):
Loolks clean AF!
Couple of small suggestions for fixes + missing docs.
When we return via get_all can we add some small tests to check that the vertex has the correct views applied to it - i.e. if I run an algorithm on a windowed graph or a layered graph - when I get the result out it should also be windowed+layered.
Need to do benchmarking as discussed.

raphtory/src/python/graph/algorithm_result.rs
```diff
@@ -34,50 +36,85 @@ macro_rules! py_algorithm_result {
...
+            )> {
+                self.0.sort_by_vertex(reverse)
             }
 
             /// Creates a dataframe from the result
             ///
             /// Returns:
             ///     A `pandas.DataFrame` containing the result
-            pub fn to_df(&self) -> PyResult<PyObject> {
+            fn to_df(&self) -> PyResult<PyObject> {
```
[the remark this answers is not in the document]
Haaroon: doesnt make a difference, it is still exposed

Haaroon:
> Loolks clean AF! Couple of small suggestions for fixes + missing docs. When we return via get_all can we add some small tests to check that the vertex has the correct views applied to it - i.e. if I run an algorithm on a windowed graph or a layered graph - when I get the result out it should also be windowed+layered. Need to do benchmarking as discussed.

added tests

PR #1343, opened 2023-10-19
Adding methods for server side pagination

rachchan:
Added methods to graphql to support server side pagination in the UI, including search count, edge count and node count, and altered edges method so that it can take a limit and offset from the UI.
 
Better UI performance, rather than loading all pages at once, only queries each page at a time

Works like before, just added more functionality to existing methods and new methods

Playground 

### Issues

### DIFF
--- a/raphtory-graphql/src/model/graph/graph.rs
+++ b/raphtory-graphql/src/model/graph/graph.rs
@@ -136,11 +136,7 @@ impl GqlGraph {
     }
 
     async fn search_count(&self, query: String) -> usize {
-        self.graph
-            .search(&query, 999999, 0)
-            .into_iter()
-            .flatten()
-            .count()
+        self.graph.search_count(&query).unwrap_or(0)
     }
 
     async fn node_count(&self, filter: Option<NodeFilter>) -> usize {
--- a/raphtory-graphql/src/model/graph/graph.rs
+++ b/raphtory-graphql/src/model/graph/graph.rs
@@ -135,8 +135,12 @@ impl GqlGraph {
             .collect()
     }
 
-    async fn search_count(&self, query: String) -> usize {
-        self.graph.search_count(&query).unwrap_or(0)
+    async fn search_vertex_count(&self, query: String) -> usize {
+        self.graph.search_vertex_count(&query).unwrap_or(0)
+    }
+
+    async fn search_edge_count(&self, query: String) -> usize {
+        self.graph.search_edge_count(&query).unwrap_or(0)
     }
 
     async fn node_count(&self, filter: Option<NodeFilter>) -> usize {
@@ -175,6 +179,10 @@ impl GqlGraph {
             (None, None, None) => {
                 return self.graph.edges().into_iter().map(|ev| ev.into()).collect()
             }
+            (Some(filter), None, None) => return get_edges_with_filter(&self.graph, filter),
+            (None, Some(limit), Some(offset)) => {
+                return get_edges_with_limit(&self.graph, limit, offset)
+            }
             (Some(filter), Some(limit), Some(offset)) => {
                 return get_edges_with_filter_limit(&self.graph, filter, limit, offset);
             }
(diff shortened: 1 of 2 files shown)


raphtory-graphql/src/model/graph/graph.rs
```diff
@@ -135,16 +135,122 @@ impl GqlGraph {
...
+    async fn node_count(&self, filter: Option<NodeFilter>) -> usize {
+        if let Some(filter) = filter {
+            self.graph
+                .vertices()
+                .iter()
+                .map(|vv| vv.into())
+                .filter(|n| filter.matches(n))
+                .count()
+        } else {
+            self.graph.vertices().len()
```
miratepuffin, line 155: We have a count_vertices function we can use here

raphtory-graphql/src/model/graph/graph.rs
```diff
@@ -135,16 +135,122 @@ impl GqlGraph {
...
+            self.graph
                 .edges()
                 .into_iter()
                 .map(|ev| ev.into())
                 .filter(|ev| filter.matches(ev))
-                .collect(),
-            None => self.graph.edges().into_iter().map(|ev| ev.into()).collect(),
+                .count()
+        } else {
+            self.graph.edges().count()
```
miratepuffin, line 168: there is a graph.count_edges function which can be used here

raphtory-graphql/src/model/graph/graph.rs
```diff
@@ -135,16 +135,122 @@ impl GqlGraph {
             .collect()
     }
 
-    async fn edges<'a>(&self, filter: Option<EdgeFilter>) -> Vec<Edge> {
-        match filter {
-            Some(filter) => self
-                .graph
+    async fn search_count(&self, query: String) -> usize {
```
miratepuffin, line 138: Tantivy has a collector https://docs.rs/tantivy/latest/tantivy/collector/struct.Count.html which allows us to do this a bit better - will need to push it internally

miratepuffin (changes requested):

raphtory-graphql/src/model/graph/graph.rs
```diff
@@ -135,16 +135,118 @@ impl GqlGraph {
...
+            (None, Some(limit), None) => return get_edges_with_limit(&self.graph, limit, 0),
+        }
+
+        fn get_edges_with_filter_limit(
+            graph: &IndexedGraph<DynamicGraph>,
+            filter: EdgeFilter,
+            limit: usize,
+            offset: usize,
+        ) -> Vec<Edge> {
+            let start_time = limit * offset;
```
miratepuffin, line 201: wwanna change these to start_pos/end_pos to make it clearer

raphtory-graphql/src/model/graph/graph.rs
```diff
@@ -135,16 +135,118 @@ impl GqlGraph {
...
+            (Some(filter), Some(limit), Some(offset)) => {
+                return get_edges_with_filter_limit(&self.graph, filter, limit, offset);
+            }
+            (Some(filter), None, Some(offset)) => {
+                return get_edges_with_filter_limit(&self.graph, filter, 10, offset)
+            }
+            (Some(filter), Some(limit), None) => {
+                return get_edges_with_filter_limit(&self.graph, filter, limit, 0)
+            }
+            (Some(filter), None, None) => return get_edges_with_filter(&self.graph, filter),
```
miratepuffin, line 187: Probably want to put (None, None, None) (Some, None, None) and (None, Some, Some) and (Some, Some, Some) at the top and the weirder ones lower down.

raphtory-graphql/src/model/graph/graph.rs
```diff
@@ -135,16 +135,118 @@ impl GqlGraph {
...
+                .filter(|ev| filter.matches(ev))
+                .collect()
+        }
+
+        fn get_edges_with_limit(
+            graph: &IndexedGraph<DynamicGraph>,
+            limit: usize,
+            offset: usize,
+        ) -> Vec<Edge> {
+            let start_time = limit * offset;
```
miratepuffin, line 235: Same as above.

miratepuffin (changes requested):

PR #1368, opened 2023-11-09
fix edge_latest_time() in GraphWithDeletions and add a test

ljeub-pometry:

fix a bug where GraphWithDeletions doesn't calculate the latest time for edges correctly and instead always returns the time of the latest deletion.

added a test


iamsmkr (approved):
LGTM

PR #1393, opened 2023-11-24
Feature/one hop filters

### LINKED ISSUE #1382: Update window and layer semantics
Currently when windowing/applying a layer view this is always a restrictive process where you cannot undo the filter set. This makes it a pain to hop around the graph on different layers or at different time points without going back to the base graph and getting a new vertex/edge object and applying your new view.

In the new semantics when applying a window/layer on a vertex or edge this will only last one hop with the returned objects resetting back to a base graph.

ljeub-pometry:

rework of the filtering apis for vertices and edges so they only apply for one hop

makes exploring paths more intuitive

yes, we need to explain the semantics

the tests

### Issues

closes #1382


ljeub-pometry:
> As we discussed, worth throwing in some python tests just for sanity and then we can merge as long as @fabianmurariu doesn't have anything major.
> 
> Have you run any benchmarks against master, out of curiocity?

I just ran the graph-ops benchmark. Performance is largely unchanged (for some reason figuring out that a vertex does not exist is about 8-10% slower though we are only talking about  1-2ns slower and finding max id is actually about 20-30% faster)

fabianmurariu (approved):
LGTM

PR #1401, opened 2023-12-01
Syncing functionalty between graphql and python

miratepuffin:

This PR is to bring the GraphQL APIs in line with the rest of Raphtory.

There is a lot of understandable confusion from incorrectly named functions/missing functions.

Yes - GraphQl functions have a schema, but full documentation is needed.

New tests added

Not currently


raphtory-graphql/src/model/graph/graph.rs
```diff
@@ -1,111 +1,250 @@
...
-    async fn node_names(&self) -> Vec<String> {
+    async fn layers(&self, names: Vec<String>) -> Option<GqlGraph> {
+        let name = self.name().await;
         self.graph
-            .vertices()
-            .into_iter()
-            .map(|v| v.name())
-            .collect_vec()
+            .layer(names)
+            .map(move |g| GqlGraph::new(name, g.into_dynamic_indexed()))
```
ljeub-pometry, line 60: If we pull out `into_dynamic_indexed` into a trait, we could make `new` here generic and not have to worry about calling `into_dynamic_indexed` everywhere explicitly.

raphtory-graphql/src/model/graph/graph.rs
```diff
@@ -1,111 +1,250 @@
...
-    graph: IndexedGraph<DynamicGraph>,
-}
 
-impl<G: StaticGraphViewOps + IntoDynamic> From<IndexedGraph<G>> for GqlGraph {
-    fn from(value: IndexedGraph<G>) -> Self {
-        Self {
-            graph: value.into_dynamic_indexed(),
-        }
+    async fn subgraph(&self, nodes: Vec<String>) -> GqlGraph {
+        let nodes: Vec<VertexRef> = nodes.iter().map(|v| v.clone().into()).collect();
```
ljeub-pometry, line 70: Should be able to call `as_ref` instead of `clone`. Even better would be to use `Borrow` in the implementation for `InputVertex` so this would just work.

raphtory-graphql/src/model/graph/graph.rs
```diff
@@ -1,111 +1,250 @@
...
+        if let Some(filter) = filter {
+            self.graph
+                .edges()
+                .into_iter()
+                .map(|ev| ev.into())
+                .filter(|ev| filter.matches(ev))
+                .count()
+        } else {
+            self.graph.count_edges()
+        }
```
ljeub-pometry, line 177: why are these named differently here than in the underlying graph?

raphtory-graphql/src/model/graph/graph.rs
```diff
@@ -320,6 +381,57 @@ impl GqlGraph {
         }
     }
 
+    async fn search_edges(&self, query: String, limit: usize, offset: usize) -> Vec<Edge> {
+        self.graph
+            .search_edges(&query, limit, offset)
+            .into_iter()
+            .flat_map(|vv| vv)
```
ljeub-pometry, line 388: this is just `flatten()`?

raphtory-graphql/src/model/graph/node.rs
```diff
@@ -38,201 +39,141 @@ impl Node {
...
-        self.vv.start()
-    }
+    ////////////////////////
+    // LAYERS AND WINDOWS //
+    ////////////////////////
 
-    async fn end(&self) -> Option<i64> {
-        self.vv.end()
+    async fn layers(&self, names: Vec<String>) -> Option<Node> {
+        self.vv.layer(names).map(move |v| v.into())
```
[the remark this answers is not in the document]
ljeub-pometry: Indeed

raphtory-graphql/src/model/graph/property.rs
```diff
@@ -1,25 +1,182 @@
...
+pub(crate) struct GqlProperties {
+    props: DynProperties,
+}
+impl GqlProperties {
+    pub(crate) fn new(props: DynProperties) -> Self {
+        Self { props }
+    }
+}
+
+impl From<DynProperties> for GqlProperties {
```
ljeub-pometry, line 81: Make this `impl<P: Into<DynProperties>> From<P>` and a lot of weird conversions go away.

ljeub-pometry (changes requested):
Got a bunch of suggestions to make this code a bit easier on the eyes...

The broken test seems simply a case of an outdated query so should be an easy fix.

PR #1405, opened 2023-12-04
pull out the EdgeStore functionality into trait EdgeLike 

fabianmurariu:
EdgeStore is specific to the current storage layer
TimeIndexOps needs to be decoupled from the underlying storage
Add External variant in enum to support TimeIndexOps from other storage variants

Need to be permit different storage layers

Nope

Unit


ljeub-pometry (approved):
This looks ok. Please check the performance impact using the `graph_ops` benchmark

PR #1412, opened 2023-12-08
Vetrex 2 node

miratepuffin:
I have changed every reference to vertex to node as this appears to be the preferred term used and brings all the APIs in sync.
We had a mixture of vertex/node throughout the project and needed to bring them in line

Yes a huge amount, the documentation will be updated accordingly 

All tests passing and project compiling.

No


examples/py/enron/lib/bindings/utils.js
fabianmurariu, line 1: did we add these in by mistake?

raphtory-graphql/src/model/graph/edge.rs
```diff
@@ -30,6 +30,30 @@ impl<G: StaticGraphViewOps + IntoDynamic, GH: StaticGraphViewOps + IntoDynamic>
 
 #[ResolvedObjectFields]
 impl Edge {
+    ////////////////////////
```
fabianmurariu, line 33: there seems to be other stuff on top of the rename

fabianmurariu (approved):
LGTM

PR #1423, opened 2023-12-20
Added date_time history/items to Temporal properties

miratepuffin:
Added a histories_date_time function to the TemporalProps object
Added a history_date_time and items_date_time function to TemporalProp

We now have datetime objects being returns for all structural histories, seems odd to miss these out.
 
Yes, these have documentation and it will be reflected in the tutorial as well.

New tests added.

Yes, but as a larger refactor of datetimes


ljeub-pometry (approved):
Looks good but we should add more descriptive errors in the future instead of just returning None

PR #1443, opened 2024-01-07
Added timezoned datetime support to python

miratepuffin:
Added support for datetimes with an offset
Offsets were exploding!

Yes, but we need a section on dates
New tests added
Need to add similar support in rust and in pandas loaders


raphtory/src/python/utils/mod.rs
```diff
@@ -71,11 +75,45 @@ impl<'source> FromPyObject<'source> for PyTime {
...
+                } else {
+                    FixedOffset::east_opt(0).ok_or(PyTypeError::new_err("Could not pass timezone"))
+                    // No timezone information, assume UTC
+                }?;
+
+                let utc_datetime = tz_offset.from_local_datetime(&naive_datetime).unwrap();
+                let timestamp = utc_datetime.timestamp() * 1000;
+                return Ok(PyTime::new(timestamp.try_into_time()?));
+            }
+        }
```
ljeub-pometry, line 111: Is manually parsing the PyDateTime object necessary? I thought pyo3 comes with support for chrono `DateTime<UTC>` (see https://pyo3.rs/main/doc/pyo3/chrono/)

basically can you not replace most of this code with 
```rust
if let Ok(date_time) = time.extract::<DateTime<UTC>>() {}
```

ljeub-pometry (changes requested):
I am a bit worried about the manual parsing of python DateTime objects, can we not just use the built-in pyo3 functionality for this?

PR #1454, opened 2024-01-15
Python plugins for document search

ricopinazo:

This PR adds a function `with_document_search_function` to the GraphQL API that enables providing a python function for document search queries

This change is needed so that we can run a GraphQL server from python with plugins

Yes, it adds a new python method to RaphtoryServer and a 

It was tested manually

No


python/src/graphql.rs
```diff
@@ -125,10 +136,76 @@ impl PyRaphtoryServer {
...
+
+                FieldFuture::Value(Some(FieldValue::list(gql_documents)))
+            });
+            for (name, type_name) in input {
+                let ty = match type_name.as_str() {
+                    // TODO: try to use PyType here!!
+                    "str" => TypeRef::named_nn(TypeRef::STRING),
+                    "int" => TypeRef::named_nn(TypeRef::INT),
+                    "float" => TypeRef::named_nn(TypeRef::FLOAT),
+                    _ => panic!("type allowed are only: 'str', 'int' and 'float'"),
```
miratepuffin, line 196: Can we change this into a pyerror being returned.

python/src/graphql.rs
```diff
@@ -125,10 +136,76 @@ impl PyRaphtoryServer {
...
+                };
+                field = field.argument(InputValue::new(name, ty));
+            }
+            let parent = parent.field(field);
+            (registry, parent)
+        };
+        println!("inserting algorithm with name {name}");
```
miratepuffin, line 203: Can we remove this print statement

raphtory/src/python/packages/vectors.rs
```diff
@@ -56,12 +56,36 @@ impl<'source> FromPyObject<'source> for PyQuery {
...
+                content: self.content.clone(),
+            })
+        } else if let Ok(edge) = edge {
+            Ok(Document::Edge {
+                src: edge.src().name(),
+                dst: edge.dst().name(),
+                content: self.content.clone(),
+            })
+        } else {
+            Err(PyException::new_err(
```
miratepuffin, line 82: I assume once we have graph docs this will also be added here

miratepuffin (approved):

PR #1455, opened 2024-01-15
Edges api improvements

### LINKED ISSUE #1403: Create new Edges object and bin the Edge iterators

ljeub-pometry:

Bring Edges api in line with nodes

Edges was returning iterators rather than structs which means not all expected methods were implemented properly

Yes, apis have changed in some cases

the tests

### Issues

closes #1403


miratepuffin (approved):
Need to check the docs are all ok with the swap to macros, but can do that as part of an overall doc review.

PR #1461, opened 2024-01-18
Allow saving arbitrary serialized data to nodes

louisch:

The ability to save arbitrary serialized data on a per-node basis through GraphQL

To enable saving of per-node properties (in a slightly hacky way, as a first step).

No but we plan on trying to move towards using the actual rust graphs inside JS in the future instead of taking this route.

The existing save_graph test was patched for this new API

In the future we need to evaluate using rust graphs on client-side instead of using this method.

### DIFF
--- a/raphtory-graphql/src/model/mod.rs
+++ b/raphtory-graphql/src/model/mod.rs
@@ -11,11 +11,12 @@ use dynamic_graphql::{
 };
 use itertools::Itertools;
 use raphtory::{
-    core::{entities::nodes::node::Node, utils::errors::GraphError, ArcStr, Prop},
-    db::api::{mutation::CollectProperties, view::MaterializedGraph},
+    core::{utils::errors::GraphError, ArcStr, Prop},
+    db::api::view::MaterializedGraph,
     prelude::{GraphViewOps, NodeViewOps, PropertyAdditionOps},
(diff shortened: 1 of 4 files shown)


raphtory-graphql/src/model/mod.rs
```diff
@@ -239,6 +244,22 @@ impl Mut {
...
+            }
+            let node_props = node_map
+                .get(node.name().to_string().as_str())
+                .ok_or(format!(
+                    "Could not find node {} in provided map",
+                    node.name()
+                ))?;
+            let node_props_as_str = node_props.as_str().ok_or("node_props must be string")?;
+            node.update_constant_properties([("props", node_props_as_str)])?;
+        }
```
ljeub-pometry, line 261: you could have made this less hacky by making the properties an object? So you can have properties other than `"props"` and should be able to have other types as well I assume?

ljeub-pometry: i.e., why not accept strings like this? "{"node1": {"x": 1, "y": 1}, "node2: {"x": 2, "y": 2}}"?

ljeub-pometry (changes requested):
I think you should make your hack a little nicer and accept properties that are not named "props" and are not a string?

PR #1482, opened 2024-02-09
Allow props to be returned as JSON & Upgrade rust to 1.75

Haaroon:

- Allows props to be returned as a JSON object via the `to_json` function in both rust and graphql
- Upgraded rust env to all use 1.75

For GQLProps via GraphQL

Multiple tests


Haaroon:
the test pass locally what are u doin github

PR #1493, opened 2024-02-16
Document props

ricopinazo:

This PR adds a new Document property type

These changes are needed so that a user can add documents to entities that can then be extracted in order to create a vectorised graph

It allows to use documents as properties, but the interfaces are the same.

No tests were added, was only tested manually

Some new tests should be probably included


raphtory/src/python/packages/vectors.rs
```diff
@@ -128,31 +140,43 @@ pub fn into_py_document(
...
+                entity: Some(edge.into_py(py)),
+                life,
             }
         }
     }
 }
 
 #[allow(dead_code)]
-#[cfg(feature = "python")]
+#[cfg(feature = "python")] // we dont need this, we already have the flag for the full module
```
miratepuffin, line 179: Test if we can remove

miratepuffin (approved):

PR #1521, opened 2024-02-28
Fixed node_type in graphql

miratepuffin:

The node_type function in graphql was still looking at properties.


rachchan (approved):
works in playground

PR #1546, opened 2024-03-22
Add the lost pip package matplotlib in the Python package

haoxins:

will get `ModuleNotFoundError: No module named 'matplotlib'` error when running the example from the website

NO

publish a new version of the Python package


miratepuffin (approved):

miratepuffin:
Thanks for this @haoxins - approved :shipit:

PR #1552, opened 2024-03-29
Add largest connected component method

D4rkisek:
This pull request introduces a method to extract the largest connected component from a graph. The method, utilises the weakly connected components algorithm to identify and return the largest connected component as a new sub-graph.

Resolving an issue.

Perhaps not, it is just a method.

Conducted several testings; when there is an empty graph, single connected component, multiple connected components, and same size connected components.

### Issues

resolves@1214

Return a neater warning message when there are two or more largest connected components.

### DIFF
--- a/python/tests/test_algorithms.py
+++ b/python/tests/test_algorithms.py
@@ -31,6 +31,14 @@ def test_connected_components():
     assert actual.get("1") == 1
 
 
+def test_largest_connected_component():
+    g = gen_graph()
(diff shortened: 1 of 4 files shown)


miratepuffin:
Hi @D4rkisek, Thanks for contributing! 

Would you be able to expose this on the graph object in python? I can point you in the right direction of it if you like

D4rkisek:
Yes, I would like that @miratepuffin, thank you.

Issue #1214, opened 2023-08-31
Add get_largest_component function to Raphtory


Alnaimi-:
We should add a get_largest_component function which returns the largest subgraph after running connected components.

PR #1562, opened 2024-04-09
Add largest connected component method

D4rkisek:
This pull request introduces a method that extracts the largest connected component from a graph. The method, utilises the weakly connected components algorithm to identify and return the largest connected component as a new sub-graph.

Resolves an issue.

Yes, it adds a new algorithm method that is exposed through the graph.rs API. While this change is currently not documented, I am prepared to provide documentation upon request.

Unit tests, in Python and Rust: tested against an empty graph, single connected component, multiple connected components, and same size connected components.

### Issues
`resolves`#1214

Return a neater warning message when there are two or more largest connected components. Can do that upon request.

### DIFF
--- a/raphtory/src/algorithms/components/lcc.rs
+++ b/raphtory/src/algorithms/components/lcc.rs
@@ -18,18 +18,18 @@ use std::collections::HashMap;
 ///
 /// # Returns:
 ///
-/// A raphtory graph, which essentially is a sub-graph of the `g`
+/// A raphtory graph, which essentially is a sub-graph of the graph `g`
 ///
 pub trait LargestConnectedComponent {
     fn largest_connected_component(&self) -> NodeSubgraph<Self>
         where
-            Self: Sized + StaticGraphViewOps;
+            Self: StaticGraphViewOps;
 }
 
 impl LargestConnectedComponent for Graph {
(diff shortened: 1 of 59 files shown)


raphtory/src/algorithms/components/lcc.rs
```diff
@@ -0,0 +1,147 @@
...
+        where
+            Self: StaticGraphViewOps;
+}
+
+impl LargestConnectedComponent for Graph {
+    fn largest_connected_component(&self) -> NodeSubgraph<Self>
+        where
+            Self: StaticGraphViewOps,
+    {
+        let connected_components_map = weakly_connected_components(self, usize::MAX, None).get_all_with_names();
```
miratepuffin, line 34: The Algorithm result has a group_by function which should allow you to easily get both the largest component and the ids of the nodes within it

miratepuffin (changes requested):
Looks great! A minor optimisation for the way you get the LCC and this can go in.

Thanks again for contributing

raphtory/src/algorithms/components/lcc.rs
D4rkisek: That sounds good. 👍

miratepuffin (approved):

PR #1586, opened 2024-04-26
Added functionality to run plugins in python in graphql

miratepuffin:
Test feature - needs to be disabled by default within release

### DIFF
--- a/raphtory-graphql/src/model/mod.rs
+++ b/raphtory-graphql/src/model/mod.rs
@@ -26,6 +26,7 @@ use std::{
     fmt::{Display, Formatter},
     io::Read,
     path::Path,
+    process::Command,
 };
 use uuid::Uuid;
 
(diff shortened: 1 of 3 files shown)


raphtory-graphql/src/model/mod.rs
```diff
@@ -55,6 +56,32 @@ impl QueryRoot {
...
+    ) -> String {
+        let output = match args {
+            None => Command::new("python")
+                .arg(format!("plugins/{}.py", script))
+                .output(),
+            Some(args) => Command::new("python")
+                .arg(format!("plugins/{}.py", script))
+                .args(&args)
+                .output(),
+        };
```
ljeub-pometry, line 72: this can run any python script on the box!

ljeub-pometry: you need to keep a list of allowed files and only look up by name or something

ljeub-pometry (changes requested):
terrible remote execution possibilities

PR #1600, opened 2024-05-14
fix node type filter

### LINKED ISSUE #1599: Fix issue with nodes type filter
Image

iamsmkr:
Fixes #1599 

Same as above

No

Unit test

### Issues

#1599 

No


miratepuffin (changes requested):
This needs to be exposed in python - we also need to just swap the semantics for the `node_filter` function on nodes instead of adding a new name.

Can you also add some tests to make sure this works with layered graphs, windowed graphs, subgraphs etc as you are utilising the base graph.

What happens when you call the function twice?

i.e. g.nodes().node_filter(["a","b"]).neighbours.node_filter(["b","c"]).collect()

PR #1619, opened 2024-05-24
fix arrowgraph for gql

shivamka1:
Wire arrowgraph to work with gql

Arrow graph was not supported for gql

No

Unit test

### Issues
fix https://github.com/Pometry/raphtory-arrow/issues/2

No


miratepuffin (approved):
Please add one test for what happens if someone tries to update an arrowgraph via graphql

PR #1622, opened 2024-05-27
Fix/performance

ljeub-pometry:

Make it possible to get eval nodes and edges based on node ids to avoid hitting the locks


narnolddd (approved):
This looks great to me! (focusing mainly on the top level api changes/changes to the motifs algo rather than storage bits ofc). Will be really helpful being able to access edges by id in eval mode.

miratepuffin (approved):

PR #1623, opened 2024-05-28
Motifs tidy ii

narnolddd:

small tidy up of variable names/removing dead code and also making a small optimisation
fix local motifs for windowed graphs.

Noticed I'd left some dead code in there. Optimisation was suggested a while ago by @LJeub which seems to shave off a few seconds in Alphabay but may be even better for bigger datasets.

Local motifs was throwing an error for nodes filtered out by window.

No

Alphabay motifs produce same output in similar to slightly shorter time. Added windowed graph test for local motifs.

### Issues
N/A

No


ljeub-pometry (approved):
Changes look good, note that you should be able to speed up the edge lookups by tracking internal instead of global node ids

narnolddd:
I unfortunately found what was secretly an n^2 routine in the local motif counting -- the algo was never seeming to finish on the large dataset and I found I was unintentionally calling neighbours about n^2 times. Changed now and it seems to be about the same time as the global motifs, just a little longer which is what I was expecting.

PR #1632, opened 2024-06-04
simplify the testing

ljeub-pometry:

Make the testing of the different storage options much more streamlined and avoid the warnings.

no

the tests

no


raphtory/src/algorithms/dynamics/temporal/epidemics.rs
```diff
@@ -252,6 +252,7 @@ mod test {
     use rand_distr::{Distribution, Exp};
     use rayon::prelude::*;
     use stats::{mean, stddev};
+    #[cfg(feature = "storage")]
```
miratepuffin, line 255: Does this need to be under storage?

miratepuffin (approved):

PR #1647, opened 2024-06-07
Remove par iter from hot path

fabianmurariu:
Fixes the slowdown from risks


ljeub-pometry (approved):
I should have benchmarked this...

PR #1648, opened 2024-06-08
Fixed documentation to state nodes instead of edges for find_nodes

brandon-haugen:

Documentation fix

The documentation currently states that `find_nodes` returns `edges` which is wrong.

```python
find_nodes(properties_dict)
Get the edges that match the properties name and value :param property_dict: the properties name and value :type property_dict: dict
```

This PR corrects the documentation so it says `nodes` instead of `edges`.

No, it just updates documentation strings so they use the correct object for the function being described

Unfortunately, I have not been able to get the documentation to build locally yet...

### Issues

None

No

### DIFF
--- a/raphtory/src/python/graph/views/graph_view.rs
+++ b/raphtory/src/python/graph/views/graph_view.rs
@@ -218,11 +218,11 @@ impl PyGraphView {
         self.graph.node(id)
     }
 
-    /// Get the edges that match the properties name and value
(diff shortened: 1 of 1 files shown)


miratepuffin (approved):
Thanks for noticing this @brandon-haugen, will merge it just in time for 0.9.2 to be released :) 

If you would like to open an issue or describe here what was going wrong with your docs build I can help you get it running if you like.

PR #1665, opened 2024-06-18
Fix/graphql server

miratepuffin:
Changed where the graphql server binds to so that it works within a container.


ricopinazo (approved):
LGTM

PR #1677, opened 2024-07-01
Improve python extensibility

miratepuffin:
In this PR we broke the raphtory python module function down into several helper functions which return the submodules (algorithms, graphql, etc) making it much easier to build custom versions of the raphtory python module.

The python module was set up in a way which made it difficult to extend for users who wish to build custom raphtory distributions with private algorithms etc.  

Not for the average user - documentation on custom distros is on way.

Tested via the standard unit tests, via the example extended raphtory library and privately.

There are several further project rejigs planned, but nothing more to do here.


raphtory-graphql/Cargo.toml
```diff
@@ -51,8 +51,13 @@ base64-compat = { workspace = true }
 time = { workspace = true }
 reqwest = { workspace = true }
 
+# python binding optional dependencies
+pyo3 = { workspace = true, optional = true }
+crossbeam-channel = "0.5.13"
```
ljeub-pometry, line 56: also put this in the workspace?

ljeub-pometry (approved):
Looks good, one dependency got added that is not pointing at the workspace, any reason for that?

PR #1683, opened 2024-07-04
Stable serialise

### LINKED ISSUE #1679: Standardise bin format

fabianmurariu:
add support to serialise the graph into protobuf format

allow us to encode/decode a graph even if the internal structure changes

new functions

unit

### Issues

closes #1679


raphtory/src/db/api/view/serialise.rs
```diff
@@ -0,0 +1,1092 @@
...
+
+impl<
+        'graph,
+        G: InternalAdditionOps
+            + GraphViewOps<'graph>
+            + DelegatePropertyAdditionOps
+            + InternalDeletionOps,
+    > StableDecode for G
+{
+    fn decode_from_bytes(buf: &[u8], graph: &Self) -> Result<(), GraphError> {
```
ljeub-pometry, line 365: Probably would be better to make a new graph here (you can use an `InternalGraph` here and then add `decode_from_bytes` constructors to `Graph` and `PersistentGraph` which call this function and wrap the `InternalGraph`)

raphtory/src/db/api/view/serialise.rs
```diff
@@ -0,0 +1,1092 @@
...
+    value
+}
+
+fn intern_str(intern_map: &mut HashMap<String, ArcStr>, s: &String) -> ArcStr {
+    let s = intern_map.entry(s.clone()).or_insert_with(|| {
+        let s: ArcStr = s.clone().into();
+        s
+    });
+    s.clone()
+}
```
ljeub-pometry, line 664: why did you rebuild the `String` interning? You can just use `graph.process_prop_value(prop)` for this?

raphtory/src/db/api/view/serialise.rs
```diff
@@ -0,0 +1,1092 @@
...
+            let AddNode {
+                id,
+                properties,
+                time,
+                type_id,
+            } = node_res?;
+            let v = VID(id as usize);
+            let props = properties
+                .as_ref()
+                .map(|p| as_prop(p, &mut string_intern_map))
```
ljeub-pometry, line 464: The graph has the `process_prop_value` method for handling the string interning

raphtory/src/db/api/view/serialise.rs
```diff
@@ -0,0 +1,1092 @@
...
+impl<
+        'graph,
+        G: InternalAdditionOps
+            + GraphViewOps<'graph>
+            + DelegatePropertyAdditionOps
+            + InternalDeletionOps,
+    > StableDecode for G
+{
+    fn decode_from_bytes(buf: &[u8], graph: &Self) -> Result<(), GraphError> {
+        let mut buf = buf;
```
ljeub-pometry, line 366: you could just write `mut buf: &[u8]` for the input argument

raphtory/src/db/graph/graph.rs
```diff
@@ -46,11 +46,26 @@ pub fn graph_equal<'graph1, 'graph2, G1: GraphViewOps<'graph1>, G2: GraphViewOps
...
+            let c1 = e.properties().constant().into_iter().count();
+            let t1 = e.properties().temporal().into_iter().count();
                 g2
                     .edge(e.src().id(), e.dst().id())
-                    .filter(|ee| ee.active(e.time().expect("exploded")))
+                    .filter(|ee| {
+                        ee.active(e.time().expect("exploded")) &&
+                        c1 == e.properties().constant().into_iter().count() &&
+                        t1 == e.properties().temporal().into_iter().count()
+                    })
```
ljeub-pometry, line 68: This was supposed to check the structural equality of two graphs, can we move the property checking to a different function? Also, if we do this, we should actually look at the values of the properties, not just the counts.

raphtory/src/db/graph/graph.rs
```diff
@@ -203,6 +218,16 @@ impl Graph {
...
+    pub fn bincode(&self) -> Result<Vec<u8>, GraphError> {
+        let encoded = bincode::serialize(&self)?;
+        Ok(encoded)
+    }
+
+    pub fn from_bincode(b: &[u8]) -> Result<Self, GraphError> {
+        let g: Graph = bincode::deserialize(b)?;
+        Ok(g)
+    }
 }
```
ljeub-pometry, line 231: this doesn't get any of the versioning information, have a look at `PyGraph.bincode` and then move that implementation here.

ljeub-pometry (changes requested):
Some quick changes:
- make the decode methods into actual constructors to avoid panics with calling them on non-empty graphs
- use the existing functionality for the string interning rather than reinventing it
- make sure the bincode methods get the versioning and add them to the `PersistentGraph` as well (also make the corresponding python methods that already exist call this new implementation)

raphtory/src/db/api/view/serialise.rs
```diff
@@ -0,0 +1,1051 @@
...
+
+        Ok(())
+    }
+}
+
+pub trait StableDecode {
+    fn decode_from_bytes(bytes: &[u8], g: &Self) -> Result<(), GraphError>;
+    fn decode(path: impl AsRef<Path>, g: &Self) -> Result<(), GraphError> {
+        let file = File::open(path)?;
+        let buf = unsafe { memmap2::MmapOptions::new().map(&file)? };
```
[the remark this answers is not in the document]
ljeub-pometry: I'm ok with keeping MMap for now as it is not a format-breaking change to change our mind about that later.

raphtory/src/db/api/view/serialise.rs
fabianmurariu: I wanted this to apply to all graphs not just Graph and PersistentGraph

raphtory/src/db/graph/graph.rs
fabianmurariu: my concern was that we start looking at everything, we'll slow it down even more, whereas here was the bare minimum

raphtory/src/db/graph/graph.rs
fabianmurariu: gone

ljeub-pometry (approved):
LGTM

PR #1692, opened 2024-07-15
implementation of temporal rich club

narnolddd:

Temporal rich club coefficient -- measures the stability of edges connecting the highest degree nodes, see [here](https://www.nature.com/articles/s41567-022-01634-8) for an in-depth definition

A useful feature for comparing snapshots of temporal networks

Yes, I have documented the python wrapper of this which I assume will be the most used one.

Implemented the toy graph example from the paper.

* Doing a bit of cloning which might be avoidable. 
* I am calling the `assert_eq_f64` from the pagerank tests, could this be moved to a more centralised `utils` class? Tbh same with the sliding window struct I have made.


narnolddd:
This is quite slow on persistent graphs, I will try and understand why.

raphtory/src/algorithms/motifs/temporal_rich_club_coefficient.rs
```diff
@@ -0,0 +1,181 @@
...
+            let new_edges: HashSet<UndirEdge> = g
+                .subgraph(s_k.clone())
+                .edges()
+                .into_iter()
+                .map(|e| undir_edge(e.src().node, e.dst().node))
+                .collect();
+            new_edges
+        })
+        .into_iter()
+        .reduce(|acc_edges, item_edges| acc_edges.intersection(&item_edges).cloned().collect());
```
ljeub-pometry, line 89: possibly worth it to use `into_par_iter` here?

ljeub-pometry (approved):
Neat algorithm to have :)

raphtory/src/algorithms/motifs/temporal_rich_club_coefficient.rs
narnolddd: Thank you! I'll have a look now.

PR #1694, opened 2024-07-15
Fix benchmark names

ljeub-pometry:

Make sure all the benchmark names are consistent with past choices so we get reasonable comparisons

The benchmark page was getting messy because some benchmarks had changed names

no


fabianmurariu (approved):
Oops!

PR #1743, opened 2024-09-03
Fix server issue in docker

### LINKED ISSUE #1728: missing ipywidgets dependency for python
`PanicException: called `Result::unwrap()` on an `Err` value: PyErr { type: <class 'ModuleNotFoundError'>, value: ModuleNotFoundError("No module named 'ipywidgets'"), traceback: None }`

miratepuffin:
Found an issue in the server where creating a client at Sever creation causes the server to die in Docker.
Added Ipywidgets as a dependency as its required for the loading bars.

Docker graphql server dies

No

Unit tests

### Issues

fixes #1728 

Need to add some docker tests

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -19,6 +19,7 @@ dependencies = [
     "requests >= 2.31.0",
     "gql[all] == 3.5.0",
     "matplotlib >= 3.4.3",
+    "ipywidgets >= 8.1.5",
 ]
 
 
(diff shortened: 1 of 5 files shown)


python/pyproject.toml
```diff
@@ -19,6 +19,7 @@ dependencies = [
     "requests >= 2.31.0",
     "gql[all] == 3.5.0",
     "matplotlib >= 3.4.3",
+    "ipywidgets >= 8.1.5",
```
ricopinazo, line 22: is this accidental? If it isn't just ignore this

ricopinazo (approved):
LGTM

python/pyproject.toml
miratepuffin: It is another fix that I shoved in here :)

PR #1745, opened 2024-09-04
Don't drop the update in the Cache if the write fails

ljeub-pometry:

Improve the error recovery in the GraphWriter

Currently, if the write fails the updates in the cache are lost.

No

Existing tests, need to figure out how to make this failure actually occur.


miratepuffin (approved):
LGTM

PR #1746, opened 2024-09-04
Graphql logging

### LINKED ISSUE #1697: impl tracing in graphql, rid printlns

miratepuffin:

1. Fixed tracing within the graphql server
2. Pushed logging throughout raphtory - removing all printlns
3. Fixed a constant properties function in graphql which was pub instead of async

- Logging and tracing make it a lot easier to see what is going on in the server

No

Current test suite

### Issues

Fixes #1697


ljeub-pometry (approved):
looks good, some minor formatting cleanup would be good.

python/tests/graphql/misc/test_tracing.py
```diff
@@ -0,0 +1,37 @@
+import tempfile
+
+from raphtory import Graph
+from raphtory.graphql import GraphServer, RaphtoryClient
+
+# TODO this doesn't work currently due to how we terminate with Tracing
```
fabianmurariu, line 6: Have we fixed this?

miratepuffin: Yes, we can turn this test back on.

raphtory-graphql/src/config/otlp_config.rs
```diff
@@ -0,0 +1,71 @@
...
+                Ok(tracer_provider) => {
+                    info!(
+                        "Sending traces to {}:{}",
+                        self.otlp_agent_host.clone(),
+                        self.otlp_agent_port.clone()
+                    );
+                    Some(tracer_provider)
+                }
+                Err(e) => {
+                    error!("{}", e.to_string());
```
fabianmurariu, line 63: do we just move on if we were told to enable tracing but we couldn't?

fabianmurariu:
This was a lot of work, but I can't help but thinking 90% of these `println!` should have been removed instead of replaced

PR #1747, opened 2024-09-04
More cache writer error handling improvements

ljeub-pometry:

Attempt to reset the file if the write fails

Slight differences in error handling

Added a test using a read-only file which at least tests some of the behaviour


fabianmurariu (approved):
LGTM

PR #1752, opened 2024-09-06
fix latest_time for exploded edge on PersistentGraph for edges that are deleted at the same time as created

ljeub-pometry:

fixes:

```
G1 = PersistentGraph()

G1.add_edge(1, 1, 2, properties={"message":"hi"})
G1.delete_edge(1, 1, 2)

print(f"G1's edges are {G1.edges.explode()}")
```

which was returning

```
G1's edges are Edges(Edge(source=1, target=2, earliest_time=1, latest_time=9223372036854775807, properties={message: hi}))
```

but the `latest_time` should have been `1`.

Added test for this case

### DIFF
--- a/raphtory-api/src/core/storage/timeindex.rs
+++ b/raphtory-api/src/core/storage/timeindex.rs
@@ -41,6 +41,16 @@ impl TimeIndexEntry {
         Self(t, 0)
     }
 
+    pub fn next(&self) -> Self {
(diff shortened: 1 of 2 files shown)


narnolddd (approved):
Yay! Looks good :)

miratepuffin (approved):

PR #1767, opened 2024-09-09
Bool properties and test_graphdb changes

fabianmurariu:
Add tests for diskgraph in test_graphdb.py
Add support for bool columns


ljeub-pometry (changes requested):
- A bunch of debug prints to clean up. 
- I am worried that our `at` implementation for the DiskGraph properties acts differently than the one for in-memory graph. It seems we are getting away with it incidentally because it is currently only used with timestamps that actually exist but there is no guarantee that will stay this way as we improve the higher-level property apis.

ljeub-pometry (approved):
LGTM

PR #1783, opened 2024-09-19
More edge property filters

ljeub-pometry:

- filter based on property existence
- filter based on set of values
- `ne`, `ge`, `le`

we need to filter properties

yes, more methods with python doc strings

needs more tests

more tests


raphtory/src/core/entities/properties/props.rs
```diff
@@ -275,6 +275,28 @@ impl PropMapper {
...
     }
+
+    pub fn get_and_validate(
+        &self,
+        prop: &str,
+        dtype: PropType,
+    ) -> Result<Option<usize>, GraphError> {
+        match self.get_id(prop) {
+            Some(id) => {
+                let existing_dtype = self.get_dtype(id).unwrap();
```
miratepuffin, line 286: I know this cannot fail, but worth turning the unwrap into a '?'

raphtory/src/db/api/view/edge_property_filter.rs
```diff
@@ -1,78 +1,207 @@
...
+        value: Prop,
+    ) -> Result<Self::FilteredViewType, GraphError> {
+        let (t_prop_id, c_prop_id) =
+            get_ids_and_check_type(self.current_filter().edge_meta(), property, value.dtype())?;
+
+        Ok(self.one_hop_filtered(EdgePropertyFilteredGraph::new(
+            self.current_filter().clone(),
+            t_prop_id,
+            c_prop_id,
+            PropValueFilter::new(value, |left, right| left < right),
```
miratepuffin, line 148: Is there a reason these all read right to left?

miratepuffin (approved):
Couple minor points/questions but otherwise LGTM

raphtory/src/core/entities/properties/props.rs
ljeub-pometry: Can turn it into an expect, if this is ever not true, we want shit to blow up!

raphtory/src/db/api/view/edge_property_filter.rs
ljeub-pometry: Swapped the arguments around so this is less confusing

raphtory/src/db/api/view/edge.rs
```diff
@@ -145,16 +145,11 @@ impl<'graph, E: BaseEdgeViewOps<'graph>> EdgeViewOps<'graph> for E {
 
     /// list the activation timestamps for the edge
     fn history(&self) -> Self::ValueType<Vec<i64>> {
-        self.map(|g, e| g.edge_history(e, g.layer_ids().constrain_from_edge(e)))
+        self.map(|g, e| g.edge_history(e, g.layer_ids()).map(|ti| ti.t()).collect())
```
fabianmurariu, line 148: No iter here?

raphtory/src/db/api/view/edge_property_filter.rs
fabianmurariu, line 1: Much repetition, this file has

fabianmurariu (approved):
LGTM

PR #1784, opened 2024-09-19
impl create update graph gql apis as plugins

shivam-880:
Impl create/update graph gql apis as plugins

These apis are custom apis which are local to a use case

Yes

Integration tests

### Issues

No


examples/custom-gql-apis/src/hello_world.rs
```diff
@@ -0,0 +1,43 @@
...
+use raphtory_core::core::utils::errors::GraphError;
+use raphtory_core::prelude::{CacheOps, GraphViewOps, ImportOps, NodeViewOps, PropertyAdditionOps};
+use raphtory_graphql::data::Data;
+use itertools::Itertools;
+use raphtory_graphql::model::plugins::query::Query;
+use raphtory_graphql::model::plugins::query_plugins::QueryPlugins;
+
+pub(crate) struct HelloWorld;
+
+impl<'a> Query<'a, QueryPlugins> for HelloWorld {
```
miratepuffin, line 16: Should QueryPlugins be singular i.e. QueryPlugin

examples/custom-gql-apis/src/python.rs
```diff
@@ -0,0 +1,16 @@
...
+use raphtory_core::python::types::repr::{Repr, StructReprBuilder};
+use raphtory_graphql::python::server::server::PyGraphServer;
+use raphtory_graphql::python::server::take_server_ownership;
+use crate::hello_world::HelloWorld;
+
+#[pyfunction(name = "add_custom_gql_apis")]
+pub fn py_add_custom_gql_apis(
+    server: PyRefMut<PyGraphServer>,
+) -> Result<PyGraphServer, PyErr> {
+    let server = take_server_ownership(server)?;
```
miratepuffin, line 13: Can this fail?

examples/custom-gql-apis/src/python.rs
```diff
@@ -0,0 +1,16 @@
...
+    let server = server.register_query_plugins::<_, HelloWorld>("hello_world");
```
miratepuffin, line 14: Can we add a mutation query as well as an example of how these are different.

raphtory-graphql/src/model/plugins/mutation_entry_point.rs
```diff
@@ -0,0 +1,26 @@
+use crate::model::algorithms::RegisterFunction;
+use async_graphql::dynamic::Object;
+use dynamic_graphql::internal::{OutputTypeName, Register, Registry, ResolveOwned, TypeName};
+use itertools::Itertools;
+use std::{collections::HashMap, sync::MutexGuard};
+
+pub trait MutationEntryPoint<'a>:
+Register + TypeName + OutputTypeName + ResolveOwned<'a> + Sync
+{
+    fn lock_plugins() -> MutexGuard<'static, HashMap<String, RegisterFunction>>;
```
miratepuffin, line 10: I assume this is called automatically and isn't up to the user to use?

examples/custom-gql-apis/src/python.rs
ricopinazo: yes if you do

```
server = ???
py_add_custom_gql_apis(server)
py_add_custom_gql_apis(server)
```
In the second call, the server would be already consumed. To be fair, we should change the server API to be mutable, as we did for the vector API

raphtory-graphql/src/model/plugins/mutation.rs
```diff
@@ -0,0 +1,59 @@
+use crate::model::plugins::mutation_entry_point::MutationEntryPoint;
+use async_graphql::{
+    dynamic::{Field, FieldFuture, FieldValue, InputValue, Object, ResolverContext, TypeRef},
+    FieldResult,
+};
+use dynamic_graphql::internal::{Register, Registry};
+use futures_util::future::BoxFuture;
+use crate::model::plugins::mutation_plugins::MutationPlugins;
+
+pub trait Mutation<'a, A: MutationEntryPoint<'a> + 'static> {
```
ricopinazo, line 10: this trait looks the same as the Query trait. Can't we have a Function trait with an apply function that can be used for both mutations and queries?

raphtory-graphql/src/model/plugins/mutation_entry_point.rs
```diff
@@ -0,0 +1,26 @@
+use crate::model::algorithms::RegisterFunction;
+use async_graphql::dynamic::Object;
+use dynamic_graphql::internal::{OutputTypeName, Register, Registry, ResolveOwned, TypeName};
+use itertools::Itertools;
+use std::{collections::HashMap, sync::MutexGuard};
+
+pub trait MutationEntryPoint<'a>:
```
ricopinazo, line 7: This trait also seems to be exactly the same as the QueryEntryPoint. Maybe we can also have a PluginEntryPoint and implement it for both MutationPlugins and QueryPlugins

raphtory-graphql/src/model/plugins/mutation_entry_point.rs
ricopinazo: yeah it's only for internal use, maybe we should do this trait pub(crate), don't think anyone should be using it outside

miratepuffin (approved):

PR #1785, opened 2024-09-19
Persistent Graph boundry fix

miratepuffin:
Change the check for node inclusion in Persistent Graphs to not include nodes that were created outside of the window.
Added several tests for all of our window functions on the Persistant Graph

Empty  nodes were being returned when on a boundry

This was creating becuase of the new docs

New suite of tests

Yes, new bug discovered with hanging deletions, to be fixed after this.


ljeub-pometry (approved):
Looks good

PR #1786, opened 2024-09-21
Feature/latest and active

miratepuffin:

1. Added a `latest()` function on all classes which equates to `x.at(graph.latest_time)` 
2. Exposed `is_active` on edge, edges,  and node as a short hand to see if the entitiy has any updates within the window - this is not added only nodes as they are filtered through windows so it would basically always return true.
3. Add event_graph function to Graph and persistent_graph function to PersistentGraph - these are basically just noops, but make it so in python you can call them without knowing what type you current have.

TODO:

- [x] Need to add all of the above to graphql

- It is a massive pain (expecially in graphql) to always look up the current time or latest time in the graph
- It can often be the case (say when doing a rolling) that you end up with an 'Empty' edge/node - this is done becuase dealing with Nones is a pain in the ass - so is_active() allows you to check this.

Yes and yes

New tests

### Issues
Created #1787

This has opened a couple of questions on if other filters should apply a window, and how should the DSL filtering be read.

### DIFF
--- a/python/python/raphtory/__init__.pyi
+++ b/python/python/raphtory/__init__.pyi
@@ -2925,6 +2925,7 @@ class MutableNode:
         An iterator over the neighbours of this node that point into this node.
         """
 
+    def is_active(self): ...
     def latest(self):
         """
          Create a view of the Node including all events at the latest time.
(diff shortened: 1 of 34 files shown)


ljeub-pometry (approved):
LGTM

PR #1794, opened 2024-09-30
add float timestamp support in python

### LINKED ISSUE #1754: Support for floats as timestamps

ljeub-pometry:

Add conversion from float for python timestamp input

Supports the format returned by the `timestamp` method on python datetime objects

Added test to make sure this works

### Issues

fixes #1754


miratepuffin (approved):
LGTM

PR #1817, opened 2024-10-11
Support multiple layers for disk storage

fabianmurariu:
make the edge ids global for disk storage

because our APIs need edge ids to be aligned

unit tests


raphtory/src/core/entities/properties/graph_meta.rs
```diff
@@ -60,7 +60,9 @@ impl GraphMeta {
         if is_static {
             Ok(self.constant_mapper.get_or_create_id(name))
         } else {
-            self.temporal_mapper.get_or_create_and_validate(name, dtype)
+            self.temporal_mapper
+                .get_or_create_and_validate(name, dtype)
+                .map_err(|e| e.into())
```
ljeub-pometry, line 65: Maybe we should just move the GraphError enum to `raphtory-api` instead of having to worry about different error types?

raphtory/src/core/mod.rs
```diff
@@ -17,15 +17,13 @@
 //!
 //! `raphtory` supports  support for the following platforms:
 //!
-//! **Note** they must have Rust 1.53 or later.
+//! **Note** they must have Rust 1.80 or later.
```
ljeub-pometry, line 20: Makes me wonder if there is a way to get this from the Cargo.toml automatically? This is bound to be out of date again next time we update it.

raphtory/src/serialise/serialise.rs
```diff
@@ -215,7 +215,7 @@ impl StableEncode for GraphStorage {
         let edges = storage.edges();
         for eid in 0..edges.len() {
             let eid = EID(eid);
-            let edge = edges.edge(ELID::new(eid, Some(0))); // This works for DiskStorage as long as it has a single layer
+            let edge = edges.edge(eid); // This works for DiskStorage as long as it has a single layer
```
ljeub-pometry, line 218: this comment is no longer relevant

ljeub-pometry (approved):
No complaints from my end, this looks good!

raphtory/src/core/entities/properties/graph_meta.rs
fabianmurariu: I don't want to do this as part of this PR

PR #1820, opened 2024-10-15
fix the `to_df` in AlgorithmResult and Edges

ljeub-pometry:

- `to_df` in `AlgorithmResult` no longer returns internal ids
- `Graph.edges.explode().to_df()` is now equivalent to `Graph.edges.to_df(explode=True)`, in particular the history is no longer duplicated for each exploded edge.
- `PyDirection` is no more and direction arguments now take strings as input directly (The only way to construct a `PyDirection` was via passing in a string anyway so this seemed entirely confusing and useless)

- internal ids are useless
- Direction input in python was confusing
- `to_df()` on exploded edges was broken

extended data frame conversion tests to include exploded edges

### DIFF
--- a/python/python/raphtory/__init__.pyi
+++ b/python/python/raphtory/__init__.pyi
@@ -5656,14 +5656,6 @@ class PropertyFilter:
     def __init__(self):
         """Initialize self.  See help(type(self)) for accurate signature."""
 
-class PyDirection:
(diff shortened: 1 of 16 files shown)


fabianmurariu (approved):
LGTM

miratepuffin (approved):

PR #1832, opened 2024-10-18
Snapshot at/latest

ricopinazo:
This PR adds two new window functions `snapshot_at` and `snapshot_latest`

These two new functions are needed for those cases where users work with graphs that might be event or persistent and they just want to get all of those entities that have not been deleted at some point in time, which corresponds to `before(t+1)` for event graphs and `at` for persistent graphs.

Yes, docs were added to rust and python

A snapshot test was added to `raphtory/src/db/api/view/time.rs`

No

### DIFF
--- a/raphtory/src/db/api/view/time.rs
+++ b/raphtory/src/db/api/view/time.rs
@@ -120,10 +120,14 @@ pub trait TimeOps<'graph>:
     /// Create a view that only includes events at the latest time
     fn latest(&self) -> Self::WindowedViewType;
 
-    // TODO: add docs
+    /// Create a view including all events that have not been explicitly deleted at `time`
+    ///
+    /// This is equivalent to `before(time + 1)` for `EventGraph`s and `at(time)` for `PersitentGraph`s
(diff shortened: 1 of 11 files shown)


raphtory/src/db/api/view/time.rs
```diff
@@ -203,6 +215,20 @@ impl<'graph, V: OneHopFilter<'graph> + 'graph + InternalTimeOps<'graph>> TimeOps
...
     }
 
+    fn snapshot_at<T: IntoTime>(&self, time: T) -> Self::WindowedViewType {
+        match self.current_filter().graph_type() {
+            GraphType::EventGraph => self.before(time.into_time() + 1),
+            GraphType::PersistentGraph => self.at(time),
+        }
+    }
+
+    fn snapshot_latest(&self) -> Self::WindowedViewType {
```
miratepuffin, line 225: Possibly worth for the event graph just returning self? Will be faster this way as there wont be a window

miratepuffin (changes requested):
Looks good! One minor change and some tests please :)

miratepuffin (approved):
LGTM, just need to remove @with_disk_graph for the time being

PR #1854, opened 2024-11-08
Feature/node state ops

ljeub-pometry:

Make lazy node state views support time ops and layer ops

so you can get a windowed degree for all nodes in the graph for example

Adds additional APIs that are documented

Needs more tests


raphtory/src/db/api/state/ops/node.rs
```diff
@@ -0,0 +1,146 @@
...
+    type Filtered<G: GraphViewOps<'graph>>: NodeOp<Output = Self::Output>
+        + NodeOpFilter<'graph, Graph = G>
+        + 'graph;
+
+    fn graph(&self) -> &Self::Graph;
+
+    fn filtered<G: GraphViewOps<'graph>>(&self, graph: G) -> Self::Filtered<G>;
+}
+
+#[derive(Debug, Clone, Copy)]
```
fabianmurariu, line 36: I wonder if many of these could just be special cases of a more generic 
``` struct NodeOpFn<O, F: Fn(&GraphStorage, VID) -> O>(F)```
and could be just made from constructors like 
``` impl NodeOpFn {
fn id()-> ...
```

raphtory/src/python/graph/node_state.rs
fabianmurariu, line 1: This file alone makes me SERIOUSLY want to reconsider the way we use generics to filter the graph

raphtory/src/db/api/state/ops/node.rs
ljeub-pometry: If you do this you can no longer name the type which causes a bunch of problems, though maybe we can get away with `fn`

fabianmurariu (approved):
LGTM

PR #1855, opened 2024-11-08
Feature/create node

shivam-880:
Add create node api to create node if doesn't already exits. Fail otherwise.
Removes the ability to create RemoteGraph directly, can now only be done through the client

An alternative way to create node besides add node which updates the node if node already exists in the graph

Documentation and Graphql docs

Unit tests

No

### DIFF
--- a/python/tests/graphql/edit_graph/test_graphql.py
+++ b/python/tests/graphql/edit_graph/test_graphql.py
@@ -435,7 +435,7 @@ def test_create_node_using_client():
             }
         }
 
-        remote_graph = RemoteGraph(path="g", client=client)
+        remote_graph = client.remote_graph(path="g")
         remote_graph.create_node(timestamp=0, id="oogway")
         assert client.query(query_nodes) == {
             "graph": {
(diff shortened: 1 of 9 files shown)


python/tests/graphql/edit_graph/test_graphql.py
```diff
@@ -382,6 +382,75 @@ def test_graph_properties_query():
...
+        query_nodes = """{graph(path: "g") {nodes {list {name}}}}"""
+        assert client.query(query_nodes) == {
+            "graph": {
+                "nodes": {
+                    "list": [{"name": "ben"}, {"name": "shivam"}]
+                }
+            }
+        }
+
+        remote_graph = RemoteGraph(path="g", client=client)
```
ljeub-pometry, line 438: do we not have a `client.remote_graph(path="g")` method?

raphtory/src/db/graph/graph.rs
```diff
@@ -3238,4 +3238,19 @@ mod db_tests {
         let graph = pool.install(|| Graph::new());
         assert_eq!(graph.core_graph().internal_num_nodes(), 0);
     }
+
+    #[test]
+    fn test_create_node() {
+        let g = Graph::new();
+        g.create_node(0, 1, [("test".to_string(), Prop::Bool(true))], None)
```
ljeub-pometry, line 3245: the `to_string` is not needed, right?

ljeub-pometry (approved):
please run tidy, otherwise looks good

miratepuffin (changes requested):
Couple of small suggestions and then we can get it in :)

miratepuffin (approved):

Issue #1737, opened 2024-09-02
Add an `import_as` function for edges and nodes


miratepuffin:
This will allow us to 'copy' nodes/edges under a new id in the same graph and change id when moving graph (useful when one graph is indexed by ints and the other by strings or vise versa)

PR #1859, opened 2024-11-18
Feature/import as

shivamka1:
#1737: Add an import_as function for edges and nodes
Changed the force argument in import to 'merge' to better indicate what it is doing

When importing node/nodes or edge/edges from an existing graph user should be able to provide new id.

Yes, yes.

Unit tests

No

### DIFF
--- a/raphtory/src/db/api/mutation/import_ops.rs
+++ b/raphtory/src/db/api/mutation/import_ops.rs
@@ -36,14 +36,12 @@ pub trait ImportOps:
 {
     /// Imports a single node into the graph.
     ///
-    /// This function takes a reference to a node and an optional boolean flag `merge`.
-    /// If `merge` is `false`, the function will return an error if the node already exists in the graph.
-    /// If `merge` is `true`, the function will overwrite the existing node in the graph.
-    ///
     /// # Arguments
     ///
     /// * `node` - A reference to the node to be imported.
-    /// * `merge` - An optional boolean flag. If `true`, the function will overwrite the existing node.
+    /// * `merge` - An optional boolean flag.
+    ///             If `merge` is `false`, the function will return an error if the imported node already exists in the graph.
+    ///             If `merge` is `true`, the function merges the histories of the imported node and the existing node (in the graph).
     ///
     /// # Returns
     ///
@@ -56,15 +54,13 @@ pub trait ImportOps:
 
     /// Imports a single node into the graph.
     ///
-    /// This function takes a reference to a node and an optional boolean flag `merge`.
-    /// If `merge` is `false`, the function will return an error if the node already exists in the graph.
-    /// If `merge` is `true`, the function will overwrite the existing node in the graph.
-    ///
     /// # Arguments
     ///
     /// * `node` - A reference to the node to be imported.
     /// * `new_id` - The new node id.
-    /// * `merge` - An optional boolean flag. If `true`, the function will overwrite the existing node.
+    /// * `merge` - An optional boolean flag.
+    ///             If `merge` is `false`, the function will return an error if the imported node already exists in the graph.
+    ///             If `merge` is `true`, the function merges the histories of the imported node and the existing node (in the graph).
     ///
     /// # Returns
     ///
@@ -83,14 +79,12 @@ pub trait ImportOps:
 
     /// Imports multiple nodes into the graph.
     ///
-    /// This function takes a vector of references to nodes and an optional boolean flag `merge`.
-    /// If `merge` is `false`, the function will return an error if any of the nodes already exist in the graph.
-    /// If `merge` is `true`, the function will overwrite the existing nodes in the graph.
-    ///
     /// # Arguments
     ///
     /// * `nodes` - A vector of references to the nodes to be imported.
-    /// * `merge` - An optional boolean flag. If `true`, the function will overwrite the existing nodes.
+    /// * `merge` - An optional boolean flag.
+    ///             If `merge` is `false`, the function will return an error if any of the imported nodes already exists in the graph.
+    ///             If `merge` is `true`, the function merges the histories of the imported nodes and the existing nodes (in the graph).
     ///
     /// # Returns
     ///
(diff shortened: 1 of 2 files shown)


raphtory/src/db/graph/graph.rs
```diff
@@ -726,6 +729,337 @@ mod db_tests {
...
+        let g = Graph::new();
+        let g_a = g.add_node(0, "A", NO_PROPS, None).unwrap();
+        let g_b = g
+            .add_node(1, "B", vec![("temp".to_string(), Prop::Bool(true))], None)
+            .unwrap();
+        let _ = g_b.add_constant_properties(vec![("con".to_string(), Prop::I64(11))]);
+
+        let gg = Graph::new();
+        gg.add_node(1, "Q", NO_PROPS, None).unwrap();
+        let res = gg.import_nodes_as(vec![&g_a, &g_b], vec!["P", "Q"], false);
```
miratepuffin, line 808: I think we would want this to either fail from the beginning (and import none) or to return a list of failed imports such that you can rename them

raphtory/src/db/graph/graph.rs
```diff
@@ -726,6 +729,337 @@ mod db_tests {
...
+
+        let gg = Graph::new();
+        let e = gg.add_edge(1, "X", "Y", NO_PROPS, None).unwrap();
+        let res = gg.import_edge_as(&e_a_b, ("X", "Y"), false);
+        match res {
+            Err(EdgeExistsError(src_id, dst_id)) => {
+                assert_eq!(
+                    src_id.to_string(),
+                    "X",
+                    "Unexpected node ID for NodeExistsError"
```
miratepuffin, line 880: This reads like there is a problem becuase the nodes existed, not because the edge was there no?

raphtory/src/db/graph/graph.rs
```diff
@@ -726,6 +729,337 @@ mod db_tests {
...
+        assert_eq!(nodes, vec!["P", "Q"]);
+        let y = gg.node("Q").unwrap();
+        assert_eq!(y.name(), "Q");
+        assert_eq!(y.history(), vec![1]);
+        assert_eq!(y.properties().get("temp").unwrap(), Prop::Bool(true));
+        assert_eq!(y.properties().constant().get("con").unwrap(), Prop::I64(11));
+    }
+
+    #[test]
+    fn import_edge_as() {
```
miratepuffin, line 855: Need to check the pathway where the edge is not there and no force

raphtory/src/db/graph/graph.rs
```diff
@@ -726,6 +729,337 @@ mod db_tests {
         assert_eq!(gg.edges().len(), 2);
     }
 
+    #[test]
+    fn import_node_as() {
```
miratepuffin, line 733: Need to check the pathway where the node is not there an no force

raphtory/src/db/graph/graph.rs
```diff
@@ -726,6 +729,337 @@ mod db_tests {
...
+        let g_b = g
+            .add_node(1, "B", vec![("temp".to_string(), Prop::Bool(true))], None)
+            .unwrap();
+        let _ = g_b.add_constant_properties(vec![("con".to_string(), Prop::I64(11))]);
+        let e_a_b = g
+            .add_edge(
+                2,
+                "A",
+                "B",
+                vec![("e_temp".to_string(), Prop::Bool(false))],
```
miratepuffin, line 925: We never check this property below -

raphtory/src/db/graph/graph.rs
```diff
@@ -726,6 +729,337 @@ mod db_tests {
...
+    }
+
+    #[test]
+    fn import_edges_as() {
+        let g = Graph::new();
+        let g_a = g.add_node(0, "A", NO_PROPS, None).unwrap();
+        let g_b = g
+            .add_node(1, "B", vec![("temp".to_string(), Prop::Bool(true))], None)
+            .unwrap();
+        let _ = g_b.add_constant_properties(vec![("con".to_string(), Prop::I64(11))]);
```
miratepuffin, line 956: Whats the purpose of adding these props on the original nodes?

raphtory/src/db/graph/graph.rs
```diff
@@ -726,6 +729,337 @@ mod db_tests {
...
+        let e_b_c = g.add_edge(2, "B", "C", NO_PROPS, None).unwrap();
+
+        let gg = Graph::new();
+        let e = gg.add_edge(1, "Y", "Z", NO_PROPS, None).unwrap();
+        let res = gg.import_edges_as([&e_a_b, &e_b_c], [("X", "Y"), ("Y", "Z")], false);
+        match res {
+            Err(EdgeExistsError(src_id, dst_id)) => {
+                assert_eq!(
+                    src_id.to_string(),
+                    "Y",
```
miratepuffin, line 976: As above this reads like a problem with the nodes

raphtory/src/db/graph/graph.rs
```diff
@@ -726,6 +729,337 @@ mod db_tests {
...
+                "B",
+                vec![("e_temp".to_string(), Prop::Bool(false))],
+                None,
+            )
+            .unwrap();
+        let e_b_c = g.add_edge(2, "B", "C", NO_PROPS, None).unwrap();
+
+        let gg = Graph::new();
+        let e = gg.add_edge(1, "Y", "Z", NO_PROPS, None).unwrap();
+        let res = gg.import_edges_as([&e_a_b, &e_b_c], [("X", "Y"), ("Y", "Z")], false);
```
miratepuffin, line 971: This should add nothing, or everything which doesn't fail

raphtory/src/db/api/mutation/import_ops.rs
```diff
@@ -50,6 +54,33 @@ pub trait ImportOps:
         force: bool,
     ) -> Result<NodeView<Self, Self>, GraphError>;
 
+    /// Imports a single node into the graph.
+    ///
+    /// This function takes a reference to a node and an optional boolean flag `force`.
+    /// If `force` is `Some(false)` or `None`, the function will return an error if the node already exists in the graph.
```
miratepuffin, line 60: Seeing as you have to give a value should we change this to a bool instead of Option Bool

raphtory/src/db/api/mutation/import_ops.rs
```diff
@@ -165,6 +277,63 @@ impl<
...
+        GHH: GraphViewOps<'a>,
+        GH: GraphViewOps<'a>,
+        V: AsNodeRef + Clone + Debug,
+    >(
+        &self,
+        node: &NodeView<GHH, GH>,
+        new_id: V,
+        force: bool,
+    ) -> Result<NodeView<Self, Self>, GraphError> {
+        if !force {
```
miratepuffin, line 291: If this is all the same code as import_node, I would create an internal function which does the actual work and then have `import_node`/`import_node_as` call it - this way tests cover both and we can't accidentally make them do different things.

We should do the same with edges below

python/tests/test_graphdb/test_graphdb_imports.py
```diff
@@ -56,3 +56,221 @@ def test_import_with_int():
     g2.import_node(g.node(1))
     g2.import_nodes([g.node(2), g.node(3)])
     assert g2.count_nodes() == g.count_nodes()
+
```
miratepuffin, line 59: If you can take the comments I made within the rust tests and also apply them here that would be great

miratepuffin (changes requested):
Couple of change suggested. 

We also need to add a test for added edges from the same graph which lucas noticed deadlocks. 

Can you make sure to run `make tidy` as you need to generate the stubs for these new functions.

If you can also quickly merge master to bring in the changes to allow iterators.

miratepuffin (approved):
Can you just change the force flag to 'merge'

PR #1869, opened 2024-11-25
improve subgraph count_nodes performance

ljeub-pometry:
### What changes were proposed in this pull request?

Refactor NodeSubgraph so it actually uses the NodeList optimisation

### Why are the changes needed?

calling count_nodes on a subgraph was unnecessarily iterating through all nodes

### Does this PR introduce any user-facing change? If yes is this documented?

No api changes

### How was this patch tested?

the tests

PR #1874, opened 2024-11-27
no reason to make a Hashset when building a subgraph anymore

ljeub-pometry:

remove the Hashset building step from `subgraph` as it is no longer needed.

Missed removing this in #1869

No

The tests


fabianmurariu (approved):
LGTM

PR #1885, opened 2024-12-04
Fix the private storage when loading single layer graphs

ljeub-pometry:
- fix private storage
- tidy up warnings


fabianmurariu (approved):
LGTM

PR #1887, opened 2024-12-04
fixes #1884

### LINKED ISSUE #1884: Searching for node by name returns all nodes in the graph
Following query returns wrong results:
```gql
{
  graph(path: "0x0a5e1db3671faccd146404925bda5c59929f66c3") {
    searchNodeCount(query:"node_type:collapsed AND cluster_id:\"0x941900204497226bede1324742eb83af6b0b5eec\"")
    searchNodes(
      query: "node_type:collapsed AND cluster_id:\"0x941900204497226bede1324742eb83af6b0b5eec\"",
      limit: 4,
      offset: 0
    ) {
      id,
      nodeType,
      firstUpdate,
      lastUpdate,
      properties {
        keys,
        values(keys: ["name", "doc", "cluster_id"]) {
          key
          value
        }
      }
    }
  }
}
```

It should have instead returned single node:
```json
{
  "data": {
    "graph": {
      "searchNodeCount": 2,
      "searchNodes": [
        {
          "id": "0x1c5e2c8e97f34a5ca18dc7370e2bfc0da3baed5c",
          "nodeType": "collapsed",
          "firstUpdate": 1718937655000,
          "lastUpdate": 1727062711000,
          "properties": {
            "keys": [
              "balance",
              "cluster_id"
            ],
            "values": [
              {
                "key": "cluster_id",
                "value": "0x1c5e2c8e97f34a5ca18dc7370e2bfc0da3baed5c"
              }
            ]
          }
        },
        {
          "id": "0x941900204497226bede1324742eb83af6b0b5eec",
          "nodeType": "collapsed",
          "firstUpdate": 1720147309000,
          "lastUpdate": 1720233593000,
          "properties": {
            "keys": [
              "balance",
              "cluster_id"
            ],
            "values": [
              {
                "key": "cluster_id",
                "value": "0x941900204497226bede1324742eb83af6b0b5eec"
              }
            ]
          }
        }
      ]
    }
  }
}
```

shivam-880:
Fixes #1884 

Indexing props, node types, names, etc as TEXT allows tantivy default tokenizer to tokenize the text which is not desirable if we want exact match while looking for keywords.

No

Unit tests

No

### DIFF
--- a/python/tests/test_index.py
+++ b/python/tests/test_index.py
@@ -1,5 +1,6 @@
 from raphtory import Graph
 
+# Range queries are not supported for properties in tantivy (as of 0.22) because they are implemented as json.
 
 def test_search_in_python():
     g = Graph()
@@ -52,10 +53,11 @@ def test_search_in_python():
     assert len(index.search_nodes("name:ben AND name:hamza")) == 0
 
     # Property tests
-    assert len(index.search_nodes("value:<120 OR value_f:>30")) == 3
-    assert len(index.search_nodes("value:[0 TO 60]")) == 2
-    assert len(index.search_nodes("value:[0 TO 60}")) == 1  # } == exclusive
-    assert len(index.search_nodes("value:>59 AND value_str:abc123")) == 1
+    # assert len(index.search_nodes("value:<120 OR value_f:>30")) == 3
+    assert len(index.search_nodes("value:199")) == 1
+    # assert len(index.search_nodes("value:[0 TO 60]")) == 2
+    # assert len(index.search_nodes("value:[0 TO 60}")) == 1  # } == exclusive
+    # assert len(index.search_nodes("value:>59 AND value_str:abc123")) == 1
 
     # edge tests
     assert len(index.search_edges("from:ben")) == 2
@@ -64,10 +66,10 @@ def test_search_in_python():
     assert len(index.search_edges("to: IN [ben, hamza]")) == 2
 
     # edge prop tests
-    assert len(index.search_edges("value:<120 OR value_f:>30")) == 3
-    assert len(index.search_edges("value:[0 TO 60]")) == 2
-    assert len(index.search_edges("value:[0 TO 60}")) == 1  # } == exclusive
-    assert len(index.search_edges("value:>59 AND value_str:abc123")) == 1
+    # assert len(index.search_edges("value:<120 OR value_f:>30")) == 3
+    # assert len(index.search_edges("value:[0 TO 60]")) == 2
+    # assert len(index.search_edges("value:[0 TO 60}")) == 1  # } == exclusive
+    # assert len(index.search_edges("value:>59 AND value_str:abc123")) == 1
 
     # Multiple history points test
     g = Graph()
@@ -84,13 +86,13 @@ def test_search_in_python():
     index = g.index()
 
     # The semantics here are that the expressions independently need to evaluate at ANY point in the lifetime of the node - hence hamza is returned even though at no point does he have both these values at the same time
-    assert len(index.search_nodes("value:<70 AND value_f:<19.2")) == 1
+    # assert len(index.search_nodes("value:<70 AND value_f:<19.2")) == 1
 
     g.add_node(
         4, "hamza", properties={"value": 100, "value_f": 11.3, "value_str": "dsc2312"}
     )
     # the graph isn't currently reindexed so this will not return hamza even though he now has a value which fits the bill
-    assert len(index.search_nodes("value:>99")) == 0
+    # assert len(index.search_nodes("value:>99")) == 0
 
 
 def test_type_search():
@@ -150,7 +152,7 @@ def test_search_with_windows():
     # Testing if windowing works - ben shouldn't be included and Hamza should only have max value of 70
     assert len(w_index.search_nodes("name:ben")) == 0
     assert len(w_index.search_nodes("value:70")) == 1
-    assert len(w_index.search_nodes("value:>80")) == 0
+    # assert len(w_index.search_nodes("value:>80")) == 0
 
     assert len(w_index.search_edges("from:ben")) == 0
     assert len(w_index.search_edges("from:haaroon AND value:>70")) == 0
--- a/python/tests/test_index.py
+++ b/python/tests/test_index.py
@@ -50,7 +50,6 @@ def test_search_in_python():
     assert len(index.search_nodes("name:ben")) == 1
     assert len(index.search_nodes("name:ben OR name:hamza")) == 2
     assert len(index.search_nodes("name:ben AND name:hamza")) == 0
-    assert len(index.search_nodes("name: IN [ben, hamza]")) == 2
 
     # Property tests
     assert len(index.search_nodes("value:<120 OR value_f:>30")) == 3
--- a/python/tests/test_index.py
+++ b/python/tests/test_index.py
@@ -1,6 +1,7 @@
 from raphtory import Graph
 
 # Range queries are not supported for properties in tantivy (as of 0.22) because they are implemented as json.
+# See https://github.com/quickwit-oss/tantivy/issues/1709
 
 def test_search_in_python():
     g = Graph()
--- a/python/tests/test_index.py
+++ b/python/tests/test_index.py
@@ -52,6 +52,7 @@ def test_search_in_python():
     assert len(index.search_nodes("name:ben")) == 1
     assert len(index.search_nodes("name:ben OR name:hamza")) == 2
     assert len(index.search_nodes("name:ben AND name:hamza")) == 0
+    assert len(index.search_nodes("name: IN [ben, hamza]")) == 2
 
     # Property tests
     # assert len(index.search_nodes("value:<120 OR value_f:>30")) == 3
--- a/python/tests/test_index.py
+++ b/python/tests/test_index.py
@@ -157,7 +157,7 @@ def test_search_with_windows():
     # assert len(w_index.search_nodes("value:>80")) == 0
 
     assert len(w_index.search_edges("from:ben")) == 0
-    assert len(w_index.search_edges("from:haaroon AND value:>70")) == 0
+    # assert len(w_index.search_edges("from:haaroon AND value:>70")) == 0
     assert len(w_index.search_edges("from:haaroon AND to:hamza")) == 1
 
 
--- a/raphtory-graphql/src/lib.rs
+++ b/raphtory-graphql/src/lib.rs
@@ -76,7 +76,7 @@ mod graphql_test {
         let query = r#"
         {
           graph(path: "lotr") {
-            searchNodes(query: "temporal_properties.kind:wizard", limit: 10, offset: 0) {
+            searchNodes(query: "kind:wizard", limit: 10, offset: 0) {
               name
             }
           }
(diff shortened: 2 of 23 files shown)


python/tests/test_index.py
```diff
@@ -50,7 +50,6 @@ def test_search_in_python():
     assert len(index.search_nodes("name:ben")) == 1
     assert len(index.search_nodes("name:ben OR name:hamza")) == 2
     assert len(index.search_nodes("name:ben AND name:hamza")) == 0
-    assert len(index.search_nodes("name: IN [ben, hamza]")) == 2
```
ljeub-pometry, line 53: this should work (see https://docs.rs/tantivy/latest/tantivy/query/struct.QueryParser.html)

raphtory/src/search/mod.rs
```diff
@@ -117,9 +117,10 @@ impl<'graph, G: GraphViewOps<'graph>> IndexedGraph<G> {
         // reverse to sort by it
         schema.add_u64_field(fields::VERTEX_ID_REV, FAST | STORED);
         // add name
-        schema.add_text_field(fields::NAME, TEXT);
```
ljeub-pometry, line 120: while this does indeed fix the issue with searching for long strings, it breaks normal full-text search completely

raphtory/src/search/mod.rs
```diff
@@ -943,13 +944,121 @@ impl<G: InternalDeletionOps> InternalDeletionOps for IndexedGraph<G> {
...
+            results,
+            vec![
+                "0x1c5e2c8e97f34a5ca18dc7370e2bfc0da3baed5c",
+                "0x941900204497226bede1324742eb83af6b0b5eec"
+            ]
+        );
+
+        let results = ig
+            .search_nodes(
+                "cluster_id:\"0x941900204497226bede1324742eb83af6b0b5eec\"",
```
ljeub-pometry, line 1020: quotes should not be necessary (this is an artefact of the `STRING` field)

raphtory/src/search/mod.rs
```diff
@@ -943,13 +944,121 @@ impl<G: InternalDeletionOps> InternalDeletionOps for IndexedGraph<G> {
...
 
+    #[test]
+    fn test_decode_sk() {
+        let graph = Graph::new();
+        graph
+            .add_node(
+                0,
+                "0x0a5e1db3671faccd146404925bda5c59929f66c3",
+                [
+                    ("balance", Prop::F32(0.0011540000414242968)),
```
ljeub-pometry, line 963: remove the balance property as it is not used

raphtory/src/search/mod.rs
```diff
@@ -1164,7 +1273,7 @@ mod test {
         graph.reload().expect("reload failed");
         // Find the Wizard
         let nodes = graph
-            .search_nodes(r#"description:wizard"#, 10, 0)
+            .search_nodes(r#"description:"A wizard""#, 10, 0)
```
ljeub-pometry, line 1276: the old one should work!

ljeub-pometry (changes requested):
While changing the field type to STRING does fix the long id bug, it also disables all normal full-text searching with partial matches! We need to set a custom tokeniser that just removes the default 40 character limit for long strings

raphtory/src/search/mod.rs
```diff
@@ -76,12 +82,11 @@ pub(in crate::search) mod fields {
...
     pub const NODE_TYPE: &str = "node_type";
-    // edges
-    // pub const SRC_ID: &str = "src_id";
+    pub const EDGE_ID: &str = "edge_id";
     pub const SOURCE: &str = "from";
-    // pub const DEST_ID: &str = "dest_id";
     pub const DESTINATION: &str = "to";
-    pub const EDGE_ID: &str = "edge_id";
+    pub const CONSTANT_PROPERTIES: &str = "constant_properties";
+    pub const TEMPORAL_PROPERTIES: &str = "temporal_properties";
```
ljeub-pometry, line 89: would it be better name these `cprop` and `tprop`? @miratepuffin

ljeub-pometry: these are a lot of characters to add to the query in addition to the property name

raphtory/src/search/mod.rs
```diff
@@ -106,256 +111,113 @@ impl<'graph, G: GraphViewOps<'graph>> IndexedGraph<G> {
...
-        // ensure we add node_id as stored to get back the node id after the search
         schema.add_u64_field(fields::VERTEX_ID, FAST | STORED);
-        // reverse to sort by it
         schema.add_u64_field(fields::VERTEX_ID_REV, FAST | STORED);
-        // add name
-        schema.add_text_field(fields::NAME, STRING | STORED);
-        // add node_type
-        schema.add_text_field(fields::NODE_TYPE, STRING | STORED);
+        schema.add_text_field(fields::NAME, TEXT);
+        schema.add_text_field(fields::NODE_TYPE, TEXT);
```
ljeub-pometry, line 121: these need the tokeniser tweaks

raphtory/src/search/mod.rs
```diff
@@ -106,256 +111,113 @@ impl<'graph, G: GraphViewOps<'graph>> IndexedGraph<G> {
...
+                        .set_tokenizer("custom_default")
+                        .set_index_option(
+                            tantivy::schema::IndexRecordOption::WithFreqsAndPositions,
+                        ),
+                ),
+        );
+        schema.add_json_field(
+            fields::TEMPORAL_PROPERTIES,
+            JsonObjectOptions::default()
+                .set_stored()
```
ljeub-pometry, line 137: revisit this when changing to exploded updates

ljeub-pometry: we should be able to get the actual value from the graph

raphtory/src/search/mod.rs
```diff
@@ -106,256 +111,113 @@ impl<'graph, G: GraphViewOps<'graph>> IndexedGraph<G> {
...
-    fn new_edge_schema_builder() -> SchemaBuilder {
+    fn edge_schema_builder() -> SchemaBuilder {
         let mut schema = Schema::builder();
-        // we first add GID time, ID and ID_REV
-        // ensure time is part of the index
         schema.add_i64_field(fields::TIME, INDEXED | STORED);
-        // ensure we add node_id as stored to get back the node id after the search
+        schema.add_u64_field(fields::EDGE_ID, FAST | STORED);
         schema.add_text_field(fields::SOURCE, TEXT);
         schema.add_text_field(fields::DESTINATION, TEXT);
```
ljeub-pometry, line 155: needs tokeniser tweaks

raphtory/src/search/mod.rs
```diff
@@ -906,7 +682,42 @@ impl<G: StaticGraphViewOps + InternalAdditionOps> InternalAdditionOps for Indexe
         props: &[(usize, Prop)],
         layer: usize,
     ) -> Result<MaybeNew<EID>, GraphError> {
-        self.graph.internal_add_edge(t, src, dst, props, layer)
+        let res = self.graph.internal_add_edge(t, src, dst, props, layer)?;
+
+        let edge_id = self
```
ljeub-pometry, line 687: you can get that from the res above

raphtory/src/search/mod.rs
```diff
@@ -920,6 +731,91 @@ impl<G: StaticGraphViewOps + InternalAdditionOps> InternalAdditionOps for Indexe
     }
 }
 
+fn collect_constant_properties(
+    properties: Box<dyn Iterator<Item = (ArcStr, Prop)> + '_>,
```
ljeub-pometry, line 735: `impl Iterator<Item = (ArcStr, Prop)> + '_`

ljeub-pometry: no reason to box?

raphtory/src/search/mod.rs
```diff
@@ -920,6 +731,91 @@ impl<G: StaticGraphViewOps + InternalAdditionOps> InternalAdditionOps for Indexe
...
+) -> serde_json::Map<String, serde_json::Value> {
+    properties.map(|(k, v)| (k.to_string(), v.into())).collect()
+}
+
+fn collect_temporal_properties(
+    properties: Box<dyn Iterator<Item = (i64, ArcStr, Prop)> + '_>,
```
ljeub-pometry, line 741: same, no real reason to box

raphtory/src/search/mod.rs
```diff
@@ -920,6 +731,91 @@ impl<G: StaticGraphViewOps + InternalAdditionOps> InternalAdditionOps for Indexe
...
+        temporal_properties_map
+            .entry(t)
+            .or_insert_with(serde_json::Map::new)
+            .insert(k.to_string(), v.into());
+    });
+
+    let temporal_properties = temporal_properties_map
+        .into_iter()
+        .map(|(t, mut props)| {
+            props.insert("time".to_string(), json!(t));
```
ljeub-pometry, line 757: should not add `time` to props, time should be a separate field in the schema?

ljeub-pometry: also, this can conflict if there is already a property called `time`

raphtory/src/search/mod.rs
```diff
@@ -1011,25 +951,41 @@ mod search_tests {
...
+            vec![
+                "0x1c5e2c8e97f34a5ca18dc7370e2bfc0da3baed5c",
+                "0x941900204497226bede1324742eb83af6b0b5eec",
+            ]
+        );
+
+        let mut results = ig
             .search_nodes(
-                "cluster_id:\"0x941900204497226bede1324742eb83af6b0b5eec\"",
+                "temporal_properties.cluster_id:\"0x941900204497226bede1324742eb83af6b0b5eec\"",
```
ljeub-pometry, line 975: are the quotes around the id necessary?

ljeub-pometry: also check you can find them by name

raphtory/src/search/mod.rs
```diff
@@ -1378,9 +1336,59 @@ mod search_tests {
     }
 
     #[test]
-    fn search_by_edge_props() {
-        let g = Graph::new();
+    fn search_by_edge_props_indexed_graph() {
+        let g = IndexedGraph::new(Graph::new());
```
ljeub-pometry, line 1340: don't think any of the `to_string` calls here are actually needed?

ljeub-pometry (changes requested):
- fix tokeniser for the other text fields
- don't add `time` as a property column (can conflict with existing property)
- no need to look up the edge again as internal_add_edge returns the edge id
- no reason to set the json fields to stored

raphtory-graphql/src/lib.rs
```diff
@@ -76,7 +76,7 @@ mod graphql_test {
         let query = r#"
         {
           graph(path: "lotr") {
-            searchNodes(query: "kind:wizard", limit: 10, offset: 0) {
+            searchNodes(query: "temporal_properties.kind:wizard", limit: 10, offset: 0) {
```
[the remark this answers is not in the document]
ljeub-pometry: maybe needs to be set as a default field: https://github.com/quickwit-oss/tantivy/blob/main/doc/src/json.md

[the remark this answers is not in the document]
ljeub-pometry: `search_nodes` is our function, we can do whatever we want inside, see my commit...

ljeub-pometry (approved):
This looks good, hopefully range queries will be supported soon 🤞

PR #1890, opened 2024-12-09
create python NodeState from Arc

ljeub-pometry:

Make it possible to construct the python NodeState wrappers from an `Arc<NodeState>`

Easier to work with in python


Cargo.lock
fabianmurariu, line 1: It doesn't feel like this file should change

fabianmurariu (approved):

PR #1894, opened 2024-12-10
Document template fixes

ricopinazo:

- [x] Add default document templates
- [x] Update vector API (on the server as well) to allow choosing between using the default template, a custom one, or nothing at all, for each of the three types of entities
- [x] Fix a bug causing subgraphs to allow containing the same node more than once
- [x] Fix a bug causing NaN float to panic when querying through GraphQL
- [x] Review public API to stick to temporal_props / constant_props naming convention

Having default templates is a first step towards a smart search view on the generic UI

Yes, and the docstrings were updated accordingly

- New tests were added to test the default templates in rust
- A new python test was added to double-check the default template is used if the user doesn't specify anything

No

### DIFF
--- a/raphtory/src/db/graph/views/node_subgraph.rs
+++ b/raphtory/src/db/graph/views/node_subgraph.rs
@@ -133,7 +133,7 @@ mod subgraph_tests {
         graph.add_node(2, 2, NO_PROPS, None).unwrap();
 
         test_storage!(&graph, |graph| {
-            let sg = graph.subgraph([1, 2]);
+            let sg = graph.subgraph([1, 2, 1]); // <- duplicated nodes should have no effect
 
             let actual = sg.materialize().unwrap().into_events().unwrap();
             assert_graph_equal(&actual, &sg);
--- a/raphtory/src/db/graph/views/node_subgraph.rs
+++ b/raphtory/src/db/graph/views/node_subgraph.rs
@@ -60,6 +60,7 @@ impl<'graph, G: GraphViewOps<'graph>> NodeSubgraph<G> {
             nodes.collect()
         };
         nodes.sort();
+        nodes.dedup();
         let nodes = Index::new(nodes, graph.unfiltered_num_nodes());
         Self { graph, nodes }
     }
--- a/raphtory/src/python/graph/properties/constant_props.rs
+++ b/raphtory/src/python/graph/properties/constant_props.rs
@@ -1,7 +1,7 @@
 use crate::{
     core::Prop,
     db::api::properties::{
-        dyn_props::DynConstProperties, internal::PropertiesOps, ConstProperties,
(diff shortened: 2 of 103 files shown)


raphtory/src/db/graph/views/node_subgraph.rs
```diff
@@ -60,6 +60,7 @@ impl<'graph, G: GraphViewOps<'graph>> NodeSubgraph<G> {
             nodes.collect()
         };
         nodes.sort();
+        nodes.dedup();
```
ricopinazo, line 63: add a test to confirm this

raphtory/src/python/graph/properties/constant_props.rs
```diff
@@ -1,7 +1,7 @@
 use crate::{
     core::Prop,
     db::api::properties::{
-        dyn_props::DynConstProperties, internal::PropertiesOps, ConstProperties,
+        dyn_props::DynConstProperties, internal::PropertiesOps, ConstantProperties,
```
ricopinazo, line 4: rename this file name (and search for others)

raphtory/src/vectors/template.rs
```diff
@@ -201,7 +204,10 @@ impl DocumentTemplate {
                 let mut env = Environment::new();
                 let template = build_template(&mut env, template);
                 match template.render(NodeTemplateContext::from(node)) {
-                    Ok(document) => Box::new(std::iter::once(document.into())),
+                    Ok(mut document) => {
+                        truncate(&mut document);
```
ricopinazo, line 208: mention this in the docs / try to make this configurable

ricopinazo: add helper funtion node.generate_document() to allow debugging

ricopinazo: ideally simply use multi documents for big documents

raphtory/src/vectors/template.rs
```diff
@@ -274,14 +290,102 @@ impl<'graph, G: GraphViewOps<'graph>> From<EdgeView<G>> for EdgeTemplateContext
...
+{% endfor %}";
+
+pub const DEFAULT_GRAPH_TEMPLATE: &str = "Graph with the following properties:
+{% for (key, value) in constant_properties|items %}
+{{ key }}: {{ value }}
+{% endfor %}
+{% for (key, values) in temporal_properties|items %}
+{{ key }}:
+{% for (time, value) in values %}
+ - changed to {{ value }} at {{ time|datetimeformat }}
```
ricopinazo, line 318: allow choosing between numbers or timestamps in the API

miratepuffin (approved):
Pair reviewed with Pedro - tickets being made for spotted issues, but LGTM for now

PR #1895, opened 2024-12-11
remove early-culling code from SCC

wyatt-joyner-pometry:
Remove inefficient early-culling code from SCC implementation.

The SCC implementation features a block of code in the beginning which exhaustively checks which nodes belong to a strongly connected component by performing a BFS search and checking if the source node is reachable from itself. In the way this is implemented, this is entirely redundant to the process of just executing Tarjan's SCC algorithm, which it already subsequently executes.

The Python function signature remains unaffected. The Rust method no longer contains a `threads` parameter.

All Rust unit-tests and Python tests still pass.

The Tarjan implementation can be parallelized. Additionally, I am working on a generalized validation and benchmarking framework for the algorithms library.

### DIFF
--- a/raphtory/src/algorithms/components/scc.rs
+++ b/raphtory/src/algorithms/components/scc.rs
@@ -1,14 +1,9 @@
-use std::{
-    collections::{HashMap, HashSet},
-};
+use std::collections::{HashMap, HashSet};
 
 use crate::{
     algorithms::algorithm_result::AlgorithmResult,
     core::entities::VID,
-    db::{
-        api::view::StaticGraphViewOps,
-        graph::node::NodeView,
-    },
+    db::{api::view::StaticGraphViewOps, graph::node::NodeView},
     prelude::*,
 };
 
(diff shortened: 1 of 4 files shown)


raphtory/src/algorithms/components/scc.rs
```diff
@@ -88,13 +74,12 @@ where
...
-pub fn strongly_connected_components<G>(
-    graph: &G,
-    threads: Option<usize>,
-) -> AlgorithmResult<G, usize>
+pub fn strongly_connected_components<G>(graph: &G) -> AlgorithmResult<G, usize>
 where
     G: StaticGraphViewOps,
 {
+    // TODO: evaluate/improve this early-culling code
+    /*
```
ljeub-pometry, line 82: maybe something like https://arxiv.org/pdf/2303.04934

ljeub-pometry: for now just Tarjan should be much better than this

ljeub-pometry (approved):

PR #1898, opened 2024-12-11
add integer support to balance algorithm; apply formatting

wyatt-joyner-pometry:
Add integer support to balance algorithm.

Previously, edge properties had to be converted to floats. Now ints and floats both work as expected.

No method signatures have been affected. 

I updated the Rust unit-test to test ints and it passes.

N/A

### DIFF
--- a/raphtory/src/algorithms/metrics/balance.rs
+++ b/raphtory/src/algorithms/metrics/balance.rs
@@ -20,7 +20,7 @@ use crate::{
             task_runner::TaskRunner,
         },
     },
-    prelude::{EdgeViewOps, GraphViewOps, NodeViewOps, PropUnwrap},
+    prelude::{EdgeViewOps, GraphViewOps, NodeViewOps},
 };
 use ordered_float::OrderedFloat;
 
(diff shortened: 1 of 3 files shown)


raphtory/src/algorithms/metrics/balance.rs
```diff
@@ -58,7 +58,7 @@ fn balance_per_node<'graph, G: GraphViewOps<'graph>, GH: GraphViewOps<'graph>, C
                 prop.temporal().get(name).map(|val| {
                     val.values()
                         .into_iter()
-                        .map(|valval| valval.into_f64().unwrap_or(0.0f64))
+                        .map(|valval| valval.as_f64().unwrap_or(0.0f64))
```
ljeub-pometry, line 61: Should this algorithm have a check for the type of the property instead of just returning all 0 if it happens to be a string or something else that can't be converted to float?

ljeub-pometry (approved):
This is definitely an improvement over what it was doing before so we can merge.

I wonder if it is worth writing this so it doesn't change the type of the output compared to the input property.

Also, maybe make this return a result instead of 0 balances if the property is not convertible to float?

raphtory/src/algorithms/metrics/balance.rs
wyatt-joyner-pometry: Another alternative is that edge weights default to 1. @miratepuffin @Alnaimi- any thoughts?

miratepuffin: Result makes more sense as we discussed before christmas

PR #1899, opened 2024-12-12
Impl index based filtering for Raphtory APIs alongside Search APIs

shivamka1:
### TODOs:
- [x] Support windowed graph for index based filtering/ Write tests
- [x] Support subgraph for index based filtering/ Write tests
- [x] Impl latest_time semantics for search apis
- [x] Property index addresses the default cprop use case i.e., ignore cprop matched results if tprop has a non-matched value
- [x] Property index/Composite filters needs to support Prop/Cprop/Tprop specific filtering
- [x] Support secondary indexes
- [x] Fix search APIs for python 
- [x] Fix search APIs for graphql
- [x] Fix limit/offset issue in index filtering for composite filters/ Write tests
- [x] Add filter builder apis
    ```rust
    TemporalProperty("p2").any().eq(2).   = PropertyFilter::eq(PropertyRef::TemporalProperty(Name, TemporalEnum::Any), value)
    TemporalProperty("p2").latest().eq(2) = PropertyFilter::eq(PropertyRef::TemporalProperty(Name, TemporalEnum::Latest), value)
    ConstantProperty("p2").eq(2)          = PropertyFilter::eq(PropertyRef::ConstantProperty(Name), value)
    Property("p2").eq(2).                 = PropertyFilter::eq(PropertyRef::Property(Name), value)
    
    CompositeNodeFilter::And(vec![Property("p2").eq(), Property("p2").eq()])
    
    Property("p1").eq().and(
         Property("p2").eq().or(Property("p2").eq())
    ).and(Property("p4").eq())
    
    CompositeNodeFilter::And(vec![
        Property("p2").eq(), 
        CompositeNodeFilter::Or(vec![Property("p2").eq(), Property("p3").eq()]),
        Property("p4").eq(), 
    ])
    ```


js-raphtory/src/graph/graph_view_impl.rs
fabianmurariu, line 1: js-raphtory is gone, you need to rebase

raphtory-graphql/src/lib.rs
```diff
@@ -44,29 +44,144 @@ mod graphql_test {
     use tempfile::tempdir;
 
     #[tokio::test]
-    async fn search_for_gandalf_query() {
-        let graph = PersistentGraph::new();
+    async fn test_search_nodes_gql() {
+        let graph = Graph::new();
```
fabianmurariu, line 48: This could be a vec of (time, name, prop) and a for loop

raphtory/src/search/mod.rs
```diff
@@ -1,1498 +1,367 @@
...
+    Ok(writers)
 }
 
-fn collect_constant_properties<'a>(
-    properties: impl Iterator<Item = (ArcStr, Prop)> + 'a,
-) -> serde_json::Map<String, serde_json::Value> {
-    properties.map(|(k, v)| (k.to_string(), v.into())).collect()
+fn initialize_node_const_property_indexes(
+    graph: &GraphStorage,
+    property_indexes: Arc<RwLock<Vec<Option<PropertyIndex>>>>,
```
fabianmurariu, line 130: This needs to take a guard not the actual lock, just call `node_index.constant_property_indexes.write()`
you can set have  `property_indexes: impl DerefMut<Target = Vec<Option<PropertyIndex>>>` as a type here

raphtory/src/search/mod.rs
```diff
@@ -1,1498 +1,367 @@
...
-        "temporal_properties": temporal_properties
-    });
-
-    let document = TantivyDocument::parse_json(schema, &doc.to_string())?;
-    // println!("doc as json = {}", &document.to_json(schema));
-
-    Ok(document)
+fn initialize_edge_const_property_indexes(
+    graph: &GraphStorage,
+    property_indexes: Arc<RwLock<Vec<Option<PropertyIndex>>>>,
```
fabianmurariu, line 164: Take guard, see above

raphtory/src/search/mod.rs
```diff
@@ -1,1498 +1,367 @@
...
-        "constant_properties": constant_properties,
-        "temporal_properties": temporal_properties
-    });
-
-    let document = TantivyDocument::parse_json(schema, &doc.to_string())?;
-    // println!("doc as json = {}", &document.to_json(schema));
-    Ok(document)
+fn initialize_edge_temporal_property_indexes(
+    graph: &GraphStorage,
+    property_indexes: Arc<RwLock<Vec<Option<PropertyIndex>>>>,
```
fabianmurariu, line 181: take guard, see above

raphtory/src/search/mod.rs
```diff
@@ -1,1498 +1,367 @@
...
+>(
+    node: NodeView<G, GH>,
+    properties: Vec<(i64, ArcStr, usize, Prop)>,
+    mut property_indexes: PI,
+    node_id: u64,
+    writers: &[Option<IndexWriter>],
+) -> tantivy::Result<()> {
+    for (time, prop_name, prop_id, prop_value) in properties {
+        if let Some(Some(prop_writer)) = writers.get(prop_id) {
+            let mut hist_entries =
```
fabianmurariu, line 237: not mut

raphtory/src/search/mod.rs
```diff
@@ -1,1498 +1,367 @@
...
+            let mut hist_entries = edge.graph.temporal_edge_prop_hist_window(
+                edge.edge,
+                prop_id,
+                time,
+                time + 1,
+                layer_ids,
+            );
+
+            if let Some(property_index) = &mut property_indexes[prop_id] {
+                for (tie, p) in hist_entries.filter(|(_, v)| *v == prop_value) {
```
fabianmurariu, line 312: p is not used

raphtory/src/search/edge_index.rs
```diff
@@ -0,0 +1,344 @@
...
+            let constant_properties = self.collect_constant_properties(&edge);
+            index_edge_const_properties(
+                constant_properties.iter().cloned(),
+                self.constant_property_indexes.write()?,
+                edge_id,
+                layer_id,
+                const_writers,
+            )?;
+
+            let temporal_properties = self.collect_temporal_properties(&edge);
```
fabianmurariu, line 204: If you rewrite this as below, you avoid some allocations
```rust
            let temporal = edge.properties().temporal();
            let temporal_properties = temporal.iter().flat_map(|(key, values)| {
                let pid = values.id;
                values
                    .into_iter()
                    .map(move |(t, v)| (t, key.clone(), pid, v))
            });
            index_edge_temporal_properties(
                edge.clone(),
                temporal_properties,
                self.temporal_property_indexes.write()?,
                edge_id,
                layer_ids,
                layer_id,
                temporal_writers,
            )?;
```

raphtory/src/search/node_index.rs
```diff
@@ -0,0 +1,325 @@
...
+
+        let const_properties = self.collect_constant_properties(&node);
+        index_node_const_properties(
+            const_properties.iter().cloned(),
+            self.constant_property_indexes.write()?,
+            node_id,
+            const_writers,
+        )?;
+
+        let temporal_properties = self.collect_temporal_properties(&node);
```
fabianmurariu, line 194: same here, if you drop the collect and use iter as in edge_index.rs you'll avoid some allocations

raphtory/src/search/mod.rs
```diff
@@ -1,1498 +1,367 @@
...
+    PI: DerefMut<Target = Vec<Option<PropertyIndex>>>,
+    G: GraphViewOps<'g>,
+    GH: GraphViewOps<'g>,
+>(
+    node: NodeView<G, GH>,
+    properties: Vec<(i64, ArcStr, usize, Prop)>,
+    mut property_indexes: PI,
+    node_id: u64,
+    writers: &[Option<IndexWriter>],
+) -> tantivy::Result<()> {
```
fabianmurariu, line 234: This is very strange to me, we seem to be indexing everything in a double for loop and checking the values equal values. When we call this we know what properties are being added and we changed the index structure so we don't index everything on every timestamp.

This happens for edges as well

raphtory/src/db/api/storage/storage.rs
```diff
@@ -223,6 +294,9 @@ impl InternalAdditionOps for Storage {
         #[cfg(feature = "proto")]
         self.if_cache(|cache| cache.add_node_update(t, v, props));
 
+        #[cfg(feature = "search")]
+        self.if_index(|index| index.add_node_update(&self.graph, t, v))?;
```
fabianmurariu, line 298: Why are we not passing props down here?

raphtory/src/search/edge_index.rs
```diff
@@ -0,0 +1,347 @@
...
+            self.constant_property_indexes.clone(),
+            const_property_keys,
+        )?;
+
+        let temporal_property_keys = edge.temporal_prop_keys();
+        let temporal_writers = initialize_edge_temporal_property_indexes(
+            graph,
+            self.temporal_property_indexes.clone(),
+            temporal_property_keys,
+        )?;
```
ljeub-pometry, line 334: shouldn't need to do this here every time, you can handle new property additions in the resolve step

raphtory/src/search/edge_index.rs
```diff
@@ -0,0 +1,347 @@
...
+        &self,
+        graph: &GraphStorage,
+        edge_id: EID,
+        t: TimeIndexEntry,
+        src: VID,
+        dst: VID,
+        layer: usize,
+        props: &[(usize, Prop)]
+    ) -> Result<(), GraphError> {
+        let edge = graph
```
ljeub-pometry, line 317: do we really ned an edge view?

raphtory/src/search/graph_index.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    pub(crate) fn edge_parser(&self) -> Result<QueryParser, GraphError> {
+        Ok(QueryParser::for_index(&self.edge_index.index, vec![]))
+    }
+
+    pub(crate) fn add_node_update(
+        &self,
+        graph: &GraphStorage,
+        t: TimeIndexEntry,
+        v: VID,
```
ljeub-pometry, line 72: this should also take the properties as input!

raphtory/src/search/graph_index.rs
```diff
@@ -0,0 +1,168 @@
...
+        edge_id: EID,
+        t: TimeIndexEntry,
+        src: VID,
+        dst: VID,
+        layer: usize,
+        props: &[(usize, Prop)]
+    ) -> Result<(), GraphError> {
+        self.edge_index
+            .add_edge_update(graph, edge_id, t, src, dst, layer, props)
+    }
```
ljeub-pometry, line 89: need separate methods for constant property updates for nodes and edges and likely also methods for creating new nodes and edges so you don't have to worry about duplicate documents in the index.

ljeub-pometry: Make sure to handle overwriting of constant property values correctly (the old value no longer exists and shouldn't show up in search results)!

ljeub-pometry (commented):
More things to tidy up!

fabianmurariu (approved):
LGTM

PR #1903, opened 2024-12-20
Add group_by support for NodeState

ljeub-pometry:

- Add `group_by` support to NodeState
- `NodeState` apis in python and in- and out-components in GraphQL now return `Nodes` instead of iterator/list of `Node`
- Fixed bug in stub parser such that properties now get type annotations as well

Yes:
 - in GraphQl, the in- and out-components now require an extra `list` to get the old output but now also support pagination.
 - renamed some methods in NodeStateOps rust apis to be explicit about returning iterators and avoid conflicts with methods on `NodeState` that return references directly

New APIs have python doc strings

- [x] add test for group_by in rust and python


fabianmurariu (approved):
LGTM

PR #1906, opened 2025-01-07
GqlEdges sorting

louisch:

Provide a `sorted` field to GqlEdges so that the client may query for a list of edges sorted by some criteria. Currently supports sorting by src ID, dst ID, latest time, earliest time, property value (via `properties`, i.e. the unified properties), passing in multiple fields to sort by in order of priority, and reversing the sort. 

Timestamp sorting was requested by client

Yes, GraphQL API is currently not documented

Some basic tests run in the GraphQL Playground to test the query, and unit tests were added to `python/tests/graphql/test_edge_sorting.py`.

Some improvements to the API to allow for sorting by constant properties and temporal properties specifically could be possible in the future, if need be.


ljeub-pometry (approved):
LGTM

PR #1912, opened 2025-01-13
Feature/gql sort nodes

ljeub-pometry:

add sorting support for nodes in GraphQl

Useful for paginated queries 

Only adds additional methods

added test

Add documentation for all the GraphQl methods


fabianmurariu (approved):
I understand we need these features but we are in `DIRE` need of a query language

PR #1917, opened 2025-01-15
Clean up python documentation by fixing the warnings from the stub parser

ljeub-pometry:

- use autosummary to generate the python api docs so they remain complete
- started fixing warnings from the stub parser

Docs are getting outdated

readthedocs is building

many more warnings to fix to get better stubs


fabianmurariu (approved):
LGTM

PR #1920, opened 2025-01-17
edges const prop iter do not aligh keys with values

fabianmurariu:
Fix this

There is a bug where the const properties for edges do not align with the values

No

tests

probably


ljeub-pometry (approved):
this looks good other than need to run fmt

PR #1921, opened 2025-01-20
Update GraphQl client documentation

ljeub-pometry:

- Cleaned up documentation
- Removed the `__new__` implementation on `RemoteEdge` and `RemoteNode` (these should be created via the `RemoteGraph`).

- Inconsistent and incomplete docs
- Publicly exposed internal interface
- 

Yes, removed some constructors. (They were not documented before.)

the tests

Other parts of the documentation also need updating.


ricopinazo (approved):
LGTM

PR #1924, opened 2025-01-20
add create_subgraph function to graphql

ricopinazo:
This PR adds a new GraphQL mutation `create_subgraph`.

We need this in order to be able to save graphs in the UI

Yes it adds a new GraphQL function and the new docstring was included

This is not being tested, although the function is quite small and all its components are heavily tested in similar functions.

No

### DIFF
--- a/raphtory-graphql/src/model/mod.rs
+++ b/raphtory-graphql/src/model/mod.rs
@@ -222,6 +222,27 @@ impl Mut {
         data.insert_graph(path, g).await?;
         Ok(path.to_owned())
     }
+
+    /// Create a subgraph out of some existing graph in the server
+    ///
+    /// Returns::
+    ///    name of the new graph
+    async fn create_subgraph<'a>(
+        ctx: &Context<'a>,
+        parent_path: &str,
+        nodes: Vec<String>,
+        new_path: String,
+        overwrite: bool,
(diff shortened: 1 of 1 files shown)


raphtory-graphql/src/model/mod.rs
```diff
@@ -222,6 +222,27 @@ impl Mut {
...
+    ///    name of the new graph
+    async fn create_subgraph<'a>(
+        ctx: &Context<'a>,
+        parent_path: &str,
+        nodes: Vec<String>,
+        new_path: String,
+        overwrite: bool,
+    ) -> Result<String> {
+        let data = ctx.data_unchecked::<Data>();
+        let parent_graph = data.get_graph(parent_path)?.0.graph.materialize()?;
```
ljeub-pometry, line 238: Shouldn't materialize the parent graph again? Materialising the subgraph is enough, no?

ljeub-pometry (changes requested):

raphtory-graphql/src/model/mod.rs
ricopinazo: oh you are right, I completely missed that one

ljeub-pometry (approved):
LGTM

PR #1929, opened 2025-01-21
fix GIL deadlock when setting up a custom embedding function in python

ricopinazo:

- [x] Fix deadlock when using Python functions as custom embedding functions
- [x] Re-enable vector tracing events to be printed
- [x] Add a utility function `vg.get_graph_documents()`

The bug this PR is fixing was making impossible to use custom Python functions

Yes, and they are.

I'm a bit confused because there is already a test with a custom Python function as embedding. I don't understand how was that working

No

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,9 @@
 # exclude .ipynb from stats
 *.ipynb linguist-documentation
 
+# exclude .html from stats
+*.html linguist-documentation
+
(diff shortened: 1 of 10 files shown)


ljeub-pometry (approved):
This looks fine as an immediate fix for the problem but does the server really need to finish embedding all the graphs before it reports as being online?

PR #1930, opened 2025-01-21
replace AlgorithmResult with NodeState in rust

ljeub-pometry:

- Replaces the old AlgorithmResult interface with NodeState 
- Tidies up some of the algorithms

Interface and efficiency improvements

Yes, the NodeState apis are somewhat different (and hopefully better) than the AlgorithmResult. All methods have documentation and the return type for all the algorithms is annotated in python.

Updated tests for the new apis.


fabianmurariu (approved):
LGTM

PR #1931, opened 2025-01-21
fix slow windowed views for disk storage

ljeub-pometry:

fixes the slowdown for windowed views on top of DiskGraph


fabianmurariu (approved):
LGTM

PR #1934, opened 2025-01-27
no manual variant management for proto enums

ljeub-pometry:

minor change to the new property type structure in proto so we don't manually have to maintain int -> variant mappings


fabianmurariu (approved):
LGTM

PR #1939, opened 2025-01-29
No constant edge properties on layers without updates

ljeub-pometry:

Unifies the logic between `update_constant_properties` and `add_constant_properties` on edges to make sure that the edge actually exists in the layer that the constant properties are being added to. 

`update_constant_properties` on an edge was missing the check to make sure that the edge actually exists in that layer

Returns a clear error in case the check fails.

Added a test for the both `add_constant_properties` and `update_constant_properties` to make sure they behave as expected.


miratepuffin (approved):
LGTM

PR #1940, opened 2025-01-30
Fix quirks in rolling and expanding APIs

ljeub-pometry:

`rolling` and `expanding` can now accept `Interval` directly instead of complaining about incompatible `Error` types in the conversion

Make it easier to reuse the parsed `Interval` inside an algorithm.

Not really, all existing code continues to work as before.

the tests

no


fabianmurariu (approved):
LGTM

PR #1943, opened 2025-02-06
Decimal support

fabianmurariu:
Add support for Decimal128

Datasets with very large numbers are failing to load

New type of prop

unit


raphtory-api/src/core/mod.rs
```diff
@@ -197,6 +202,14 @@ pub fn unify_types(l: &PropType, r: &PropType, unified: &mut bool) -> Result<Pro
             }
             Ok(PropType::Map(merged))
         }
+        (PropType::Decimal { scale: l_scale }, PropType::Decimal { scale: r_scale }) => {
+            if l_scale != r_scale {
+                *unified = true;
+            }
+            Ok(PropType::Decimal {
+                scale: *l_scale.max(r_scale),
```
ljeub-pometry, line 210: Is this correct?

raphtory/src/io/arrow/prop_handler.rs
```diff
@@ -309,6 +320,17 @@ impl PropCol for DTimeCol {
         StaticArray::get(&self.arr, i).map(self.map)
     }
 }
+
+struct DecimalTimeCol {
```
ljeub-pometry, line 324: Why is this not called DecimalPropCol?

raphtory/src/serialise/proto_ext.rs
```diff
@@ -727,15 +727,16 @@ fn as_proto_prop(prop: &Prop) -> proto::Prop {
...
-        }
-        Prop::Array(blob) => prop::Value::Array(Array {
+        Prop::DTime(dt) => Some(prop::Value::DTime(
+            dt.to_rfc3339_opts(chrono::SecondsFormat::AutoSi, true),
+        )),
+        Prop::Array(blob) => Some(prop::Value::Array(Array {
             data: blob.to_vec_u8(),
-        }),
+        })),
+        Prop::Decimal(_) => None,
```
ljeub-pometry, line 738: We probably should support these here?

ljeub-pometry (changes requested):
The unification of decimals with different scales looks like it will modify existing values (at least on disk graph)? Maybe we should just error instead.

Protobuf is silently removing decimal properties when writing to file, this seems bad.

ljeub-pometry (approved):
looks good

PR #1951, opened 2025-02-16
Apply multiple views in one graphql function

miratepuffin:

- Added default layer to GQLNodes, Node and Edge as they were missing.
- Added a new function `apply_views` to all Graphql entities which allows you to apply all views in one function by passing a vec of view objects.

Arbitrarily deep json responses when you want to apply different filters are annoying to parse. 

Yes, not currently
- [ ] Tests to be added
- [ ] Need to make some good examples
- [ ] need to refactor internal node which is called `vv` not `nn`
- [ ] Should GQLNodes/Edges be renamed or Node/Edge


ricopinazo (approved):
LGTM

PR #1970, opened 2025-03-07
Use DoubleEndedIterator for last value in node temporal properties

fabianmurariu:
Actually using the sorted timestamps we painfully build

Because last value on node properties is very slow

No

With ibuprofen

Oh so many!


ljeub-pometry (approved):
LGTM

PR #1972, opened 2025-03-10
try to optimise edges for subgraph

ljeub-pometry:

Calling `edges` on a subgraph should no longer iterate over all edges in the entire graph to apply the subgraph filter.

Subgraph `edges` was slow on large graphs.

No

The tests


miratepuffin (approved):
Tested this locally and seems to be significantly faster

PR #1975, opened 2025-03-14
Output nodes and edges in similarity search

rachchan:
create gqldocuments + output nodes and edges as well as gqldocument in that object

so i can get node types and props in semantic search


miratepuffin (approved):
LGTM

PR #1976, opened 2025-03-21
Basic creation/updating of metadata in .raph for MetaGraph

louisch:

Add functions hooking into graph creation and graph writes to create/update metadata within the .raph file, which can then be fetched from the MetaGraph resource from GraphQL.

This metadata currently consists of:
- node count
- edge count
- graph properties

To provide an endpoint to fetch this information without having to decode and fetch the full graph.

A new endpoint is provided from GQL, and the new behavior will modify the .raph file, but should not be a breaking change for users.

Tests added to raphtory-graphql to test that the metadata is created/updated and can be fetched via MetaGraph (as well as namespaces in general), and also tests were added to check that the right behavior happens if an invalid .raph file exists.

None unless more complex metadata is required in the future, or a more complex storage method is needed.

### DIFF
--- a/raphtory/src/serialise/serialise.rs
+++ b/raphtory/src/serialise/serialise.rs
@@ -1,5 +1,5 @@
 use super::{proto_ext::PropTypeExt, GraphFolder};
-use crate::prelude::GraphViewOps;
+use crate::db::api::view::StaticGraphViewOps;
 use crate::{
     core::{
         entities::{graph::tgraph::TemporalGraph, LayerIds},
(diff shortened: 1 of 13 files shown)


raphtory/src/serialise/serialise.rs
```diff
@@ -41,17 +42,15 @@ macro_rules! zip_tprop_updates {
     };
 }
 
-pub trait StableEncode {
+pub trait StableEncode<'graph>: GraphViewOps<'graph> {
```
fabianmurariu, line 45: Better use `StaticGraphViewOps` here and avoid `<'graph>` for StableEncode

fabianmurariu (changes requested):
All good overall, just one comment on `StaticGraphViewOps`

fabianmurariu (approved):
LGTM

PR #1979, opened 2025-03-25
Make rust filter builder apis consistent with python

shivamka1:
This PR attempts to make both rust and python APIs consistent


ljeub-pometry (approved):
Looks good

PR #1980, opened 2025-03-26
PyEdge access to attribute edge

BaCk7:

Make edge property pub in PyEdge 

So that we can use the PyEdge as inputs outside of the crate


ljeub-pometry (approved):
All good

PR #1989, opened 2025-03-27
add jwt bearer auth

ricopinazo:

This PR adds a JWT bearer auth layer on top of Raphtory. It does it by using an EdDSA public key, which makes the server responsibility boil down to only two things:
- Correctly validating JWTs.
- Allowing access only to those resources stated in the JWT.

The responsibility for preventing a secret leakage is out of the equation since Raphtory doesn't have access to the private key, responsible for encoding JWTs.

There were no docstrings setup for the functions that were modified by these changes

Multiple python tests were added


python/tests/test_graphql/test_auth.py
```diff
@@ -0,0 +1,194 @@
...
+import tempfile
+from datetime import datetime, timezone
+from raphtory import Graph
+from raphtory.graphql import GraphServer, RaphtoryClient
+
+# openssl genpkey -algorithm ed25519 -out raphtory-key.pem
+# openssl pkey -in raphtory-key.pem -pubout -outform DER | base64
+PUB_KEY = "MCowBQYDK2VwAyEADdrWr1kTLj+wSHlr45eneXmOjlHo3N1DjLIvDa2ozno="
+# cat raphtory-key.pem
+PRIVATE_KEY = """-----BEGIN PRIVATE KEY-----
```
fabianmurariu, line 16: I wonder if github will moan about this one

fabianmurariu (approved):
LGTM

PR #1992, opened 2025-04-04
fix dataframe loader bug

ljeub-pometry:

Fixes bug where loading from DataFrame would miss adding edges to the layer adjacency lists

confirmed it fixes the problem in the doc example, needs more tests

a proper prop test for the columnar loaders


fabianmurariu (approved):
LGTM

miratepuffin (approved):
Nice! LGTM

PR #1995, opened 2025-04-07
Fix ingestion from parquet with multiple layers

ljeub-pometry:

- Fixes a bug in the parquet/dataframe readers where multilayered edges were only inserted into the outbound adjacency list for one layer.
- Materialising and empty graph view now preserves the layer information

Improved the prop tests such that they now check that the graphs match also within each layer, not just globally.


miratepuffin (approved):
LGTM

PR #1997, opened 2025-04-07
Readme changes and final docs for 0.15

miratepuffin:
Fixed APIs in readme and generated new stubs

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,11 +224,6 @@ To get you up and running with Raphtory we provide a full set of tutorials on th
 
 If API documentation is more your thing, you can dive straight in [here](https://docs.raphtory.com/)!
 
-### Online notebook sandbox
-Want to give this a go, but can't install? Check out Raphtory in action with our interactive Jupyter Notebooks! Just click the badge below to launch a Raphtory sandbox online, no installation needed.
-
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -113,12 +113,12 @@ print(
 )
 
 # Run pagerank and ask for the top ranked node
-top_node = algo.pagerank(graph).top_k(1)
+top_node = algo.pagerank(graph).top_k(5).max_item()
```
ljeub-pometry, line 116: just max_item() should do?

python/python/raphtory/__init__.pyi
```diff
@@ -1260,23 +1347,15 @@ class Graph(GraphView):
...
             None:
         """
+
     def serialise(self) -> bytes:
         """
          Serialise Graph to bytes.
 
         Returns:
           bytes:
         """
```
ljeub-pometry, line 1357: tidy activates the private storage by default, we might need to rejig this slightly

ljeub-pometry (approved):

README.md
miratepuffin: It does - I just wanted to show off the other APIs

python/python/raphtory/__init__.pyi
miratepuffin: I didn't do tidy - I just run stubs and fmt, but indeed it would be nice to have a OS version of tidy

PR #2021, opened 2025-04-15
Update python dependencies and testing

ljeub-pometry:

- remove unneeded python dependencies and make those that are not needed for core functions optional
- add tox config to run tests with the different dependencies installed 
- clean up tests so they run without the private storage enabled by default
- relax the numpy version to 1.26


miratepuffin (approved):
Nice!

PR #2023, opened 2025-04-16
add option to output graphql schema

ricopinazo:

This adds a new option so that running the graphql binary with:
```bash
raphtory-graphql schema > schema.graphql
```

Will create the file in the given path containing the graphql schema. It won't run the server in this case and will ignore any other option.

No docstrings currently for the graphql binary

Tested manually

No

### DIFF
--- a/raphtory-graphql/src/main.rs
+++ b/raphtory-graphql/src/main.rs
@@ -1,4 +1,4 @@
-use clap::{command, Parser};
+use clap::{command, Parser, Subcommand};
 use raphtory_graphql::{
     config::{
         app_config::AppConfigBuilder,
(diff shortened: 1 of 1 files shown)


raphtory-graphql/src/main.rs
```diff
@@ -52,29 +53,46 @@ struct Args {
...
+}
+
+#[derive(Subcommand)]
+enum Commands {
+    #[command(about = "Writes the GraphQL schema into the given path")]
+    Schema {
+        #[arg(value_parser)]
+        path: String,
+    },
 }
```
ljeub-pometry, line 68: I think starting the server should also be a subcommand so you would do `raphtory_graphql server...` for starting the server and `raphtory_graphql schema ...` for the schema

ljeub-pometry (changes requested):

raphtory-graphql/src/main.rs
ricopinazo: I shall make a ticket for this as discussed in the standup to include this in 0.16.0

miratepuffin (approved):
LGTM!

ricopinazo:
I included all the suggestions by @ljeub-pometry into this issue: https://github.com/Pometry/Raphtory/issues/2027

PR #2025, opened 2025-04-16
Fix security and deps

miratepuffin:
Security patches for 15.1 release

### DIFF
--- a/python/python/raphtory/__init__.pyi
+++ b/python/python/raphtory/__init__.pyi
@@ -1319,6 +1319,17 @@ class Graph(GraphView):
           MutableNode: The node object with the specified id, or None if the node does not exist
         """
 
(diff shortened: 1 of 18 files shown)


python/python/raphtory/__init__.pyi
```diff
@@ -1319,6 +1319,17 @@ class Graph(GraphView):
...
+        """
+        save graph in disk_graph format and memory map the result
+
+        Arguments:
+            graph_dir (str | PathLike): folder where the graph will be saved
+
+        Returns:
+            DiskGraphStorage: the persisted disk graph storage
+        """
+
```
ljeub-pometry, line 1332: run stubs without the private storage?

ljeub-pometry (changes requested):
run stubs without storage enabled

ljeub-pometry (approved):

Issue #2026, opened 2025-04-16
Saving index on disk


shivamka1:
Todos:
- [ ] Configure saving indexes to disk or keep it in memory (at the time you create index)
- [ ] Saving a graph should save the in-memory index to disk (should save in the same graph directory under index/)
- [ ] Loading the graph should load the index from disk

---

Index (on disk/memory-mapped)
problem 1: graph gets updated in-memory but changes could be trashed; 
                     trash the changes to index as well
problem 2: graph gets updated in-memory but changes are saved to disk (but not necessarily to the same graph file)
                      save the changes to index as well (to a different graph directory if graph is saved to a different dir)

In both of these problem statements: graph/index in-memory are decoupled with graph/index on disk. Not allowed to modify data (graph/index) on disk.

Index (on disk/Memory Mapped)
      Cached Graph is coupled to the disk. Any update to graph in-memory could be pushed to disk using write_updates() api.
       Same approach should be impl for the indexes (memory mapped or not)

--

- [ ] Impl for cached graph view
- [ ] Impl for graph loaded as cache
- [ ] Persistent graph?

https://github.com/Pometry/Raphtory/pull/2030#discussion_r2063890491

https://github.com/Pometry/Raphtory/pull/2030#discussion_r2063911235

PR #2030, opened 2025-04-22
Features/index disk

shivamka1:
Impl persisting and loading non-cached index as part of graph encode and decode. #2026

### DIFF
--- a/python/tests/test_base_install/test_index.py
+++ b/python/tests/test_base_install/test_index.py
@@ -60,7 +60,7 @@ def init_graph(graph):
 
 
 def search_nodes(graph, filter_expr, limit=20, offset=0):
-    graph.create_index()
+    graph.create_index_in_ram()
     return sorted(
         [node.name for node in graph.search_nodes(filter_expr, limit, offset)]
     )
@@ -726,7 +726,7 @@ def init_edges_graph(graph):
 
 
 def search_edges(graph, filter_expr, limit=20, offset=0):
-    graph.create_index()
+    graph.create_index_in_ram()
     return sorted(
         [
             (edge.src.name, edge.dst.name)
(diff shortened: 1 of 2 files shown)


raphtory/src/serialise/serialise.rs
```diff
@@ -1456,4 +1475,148 @@ mod proto_test {
...
+            // Loaded index that was persisted
+            let graph = Graph::decode(path).unwrap();
+            let index = graph.get_storage().unwrap().index.get();
+            assert!(index.is_some());
+            assert_search_results(&graph, &filter1, vec!["Alice"]);
+            assert_search_results(&graph, &filter2, Vec::<&str>::new());
+        }
+
+        #[test]
+        fn test_zip_encode_decode_index() {
```
ljeub-pometry, line 1606: We should make a prop-test for this similar to the materialize tests

raphtory/src/serialise/serialise.rs
```diff
@@ -1456,4 +1475,148 @@ mod proto_test {
...
+        ) {
+            let res = graph
+                .search_nodes(filter.clone(), 2, 0)
+                .unwrap()
+                .into_iter()
+                .map(|n| n.name())
+                .collect::<Vec<_>>();
+            assert_eq!(res, expected);
+        }
+
```
ljeub-pometry, line 1530: make prop tests for the different cases (folder and zip, PersistentGraph, Graph, index, no index)

ljeub-pometry (changes requested):
Some minor things to clean up:
- don't expose the write on the index publicly (only write the index when writing the graph to avoid possibly confusing mismatches)
- need to validate the schema when reading index from file
- would be nice to add some proptests for encode/decode

Issue #2073, opened 2025-05-02
Impl filter negation


shivamka1:
Refer: https://github.com/Pometry/Raphtory/pull/1991#discussion_r2048890407

PR #2088, opened 2025-05-08
impl filter negation

shivamka1:
Impl filter negation #2073


raphtory/src/db/graph/views/filter/model/node_filter.rs
```diff
@@ -115,6 +117,14 @@ impl CompositeNodeFilter {
                     node,
                 )
             }
+            CompositeNodeFilter::Not(filter) => !filter.matches_node(
```
ljeub-pometry, line 120: is `matches_node` still used?

ljeub-pometry (approved):
Can probably remove matches_node/matches_edge from composite filters

PR #2090, opened 2025-05-09
Graphql refactor + rolling/expanding

miratepuffin:
#### Removed functions/API changes
- [x] Removed _id functions from graphql as they are no longer needed 
- [x] Removed GqlGraphs as we now have Namespace support
- [x] Added Namespaces and MetaGraphs so that large collections of graphs/namespaces can be paged/listed/counted

#### New functions
- [x] Added rolling and expanding to Graph, Node, Nodes, PathFromNode, Edge and Edges

#### Refactor
- [x] Renamed all graphql structs that started with GQL to make the user facing schema a bit cleaner
- [x] Started running queries in blocking threads to improve multi query execution and ui interactivity for large graphs

#### Bug fixes
- [x] Fixed a problem with namespaces returning loads of null paths and not returning root
- [x] Fixed issue with recursive writing of indexes causing the server to crash
- [x] changed caching policy to never kick out graphs after some timeout by default
- [x] Changed WindowSet to not allow zero size step 
- [x] Fixed issue in rolling where if the step was bigger than the window size the final window would be empty.

#### Testing
- [x] Added namespaces tests to check for correct output
- [x] Added better testing for namespaces and metagraphs
- [x] Need to add tests for GQL rollings/expanding

Improvements in the UI timeline and also useful for Graphql users.
Multiple queries being blocked as queries were being run on the main thread.

The spawn blocking is probably more of a patch than a final solution, but we need to get some benchmarks going first to know what works best.


raphtory/src/db/api/view/time.rs
```diff
@@ -390,12 +387,34 @@ impl<'graph, T: TimeOps<'graph> + Clone + 'graph> Iterator for WindowSet<'graph,
...
+            if let Some(start) = window_start {
+                if start >= self.end {
+                    break;
+                }
+            }
+            count += 1;
+            cursor = cursor + self.step;
+        }
+        count
+    }
```
ljeub-pometry, line 411: if we can't do better, we shouldn't implement this as it is no better than count on the iterator but I think we could do better (the months would need some special handling). At least implement the fast option when the interval is simple?

ljeub-pometry (approved):
Some things we can fix later, doesn't really introduce new issues.

PR #2094, opened 2025-05-13
add utf8view support for proptype conversion from arrow datatype

wyatt-joyner-pometry:
adding case for utf8view to proptype conversion from arrow datatype

this is fixing a bug with utf8view-encoded proptypes

N/A

this was applied in the artemis environment and resolved the issue where utf8view-encoded props were causing problems.

I think NodeTypes might need the same fix applied.


ljeub-pometry (approved):
Looks good

PR #2097, opened 2025-05-13
fix the benchmark permissions so they can be submitted to github pages

ljeub-pometry:

Permissions on the benchmark workflow were too restrictive


wyatt-joyner-pometry (approved):
👍

PR #2123, opened 2025-06-13
fixes filter_edges for layers is broken after add_constant_properties

shivamka1:
When we add constant props to edge layers, we compare map of constant props instead.

```rust
let edge = graph
    .add_edge(1, "shivam", "kapoor", [("p1", 100u64)], Some("fire_nation"))
    .unwrap();
edge.add_constant_properties([("z", true)], Some("fire_nation"))
    .unwrap();

let filter = PropertyFilter::property("z")
    .constant()
    .eq(Prop::map([("fire_nation", true)]));
assert_eq!(filter_edges(&graph, filter), vec!["shivam->kapoor"]);
```

We also have `num_layers` api allowing us to fetch the number of unique layers in graph
```rust
graph.num_layers()
```

### DIFF
--- a/raphtory/src/db/graph/views/filter/model/property_filter.rs
+++ b/raphtory/src/db/graph/views/filter/model/property_filter.rs
@@ -15,7 +15,7 @@ use itertools::Itertools;
 use raphtory_api::core::{
     entities::properties::{
         meta::Meta,
-        prop::{sort_comparable_props, Prop, PropType},
+        prop::{sort_comparable_props, Prop},
     },
     storage::arc_str::ArcStr,
 };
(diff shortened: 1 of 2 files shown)


raphtory/src/db/api/view/layer.rs
```diff
@@ -97,6 +101,10 @@ impl<'graph, V: OneHopFilter<'graph> + 'graph> LayerOps<'graph> for V {
         let ids = self.current_filter().valid_layer_ids_from_names(layers);
         self.one_hop_filtered(LayeredGraph::new(self.current_filter().clone(), ids))
     }
+
+    fn num_layers(&self) -> usize {
+        self.current_filter().unique_layers().count()
```
miratepuffin, line 106: Can you make a note of this function in the PR so I can add it to the patch notes

miratepuffin (approved):

raphtory/src/db/graph/views/filter/model/property_filter.rs
```diff
@@ -219,19 +219,37 @@ impl PropertyFilter {
...
+        resolve_to_map: bool,
+    ) -> Result<Option<usize>, GraphError> {
         let prop_name = self.prop_ref.name();
         if let PropertyFilterValue::Single(value) = &self.prop_value {
+            if resolve_to_map {
+                if let PropType::Map(map) = value.dtype() {
+                    if let Some((_k, v)) = map.iter().next() {
+                        return Ok(meta
+                            .const_prop_meta()
+                            .get_and_validate(prop_name, v.clone())?);
```
ljeub-pometry: it's not a full validation (all the types in the map should be the same) but maybe it is good enough?

ljeub-pometry (approved):
The map validation could be improved...

PR #2133, opened 2025-06-18
impl copy on write graph index

shivamka1:
Impl graph index as immutable and mutable version such that when graph is loaded it is loaded as immutable index and is converted into mutable index which involves copying the immutable index in a temp dir only if index needs to be updated.


pometry-storage-private
ljeub-pometry, line 1: this should not be here

raphtory/src/search/graph_index.rs
```diff
@@ -25,218 +26,64 @@ use walkdir::WalkDir;
...
-        self.edge_index.print()?;
-        Ok(())
-    }
-
     pub(crate) fn persist_to_disk(&self, path: &Path) -> Result<(), GraphError> {
-        let source_path = self.path.as_ref().ok_or(GraphError::GraphIndexIsMissing)?;
+        let source_path = self
+            .path
+            .as_ref()
+            .map(|p| p.path().join("index"))
```
ljeub-pometry: no reason to have "index" folder inside the temporary directory, it can just contain the node and edge indexes directly

ljeub-pometry: the input path here can just be a GraphFolder

ljeub-pometry (changes requested):
persisting to zip is currently broken

raphtory/src/search/graph_index.rs
```diff
@@ -360,11 +211,250 @@ impl GraphIndex {
...
+                edge_index,
+            },
+            path: Arc::new(path.clone()),
+            index_spec: Arc::new(index_spec),
+        }))
+    }
+
+    pub fn make_mutable_if_needed(&mut self) -> Result<(), GraphError> {
+        if let GraphIndex::Immutable(immutable) = self {
+            let temp_dir = TempDir::new_in(immutable.path.as_ref())?;
```
ljeub-pometry: make sure this doesn't end up inside of `index` by accident

raphtory/src/search/graph_index.rs
```diff
@@ -360,11 +211,250 @@ impl GraphIndex {
...
+        }
+    }
+
+    pub fn searcher(&self) -> Searcher {
+        Searcher::new(self.index())
+    }
+}
+
+fn get_node_index_path(path: &Option<Arc<TempDir>>) -> Option<PathBuf> {
+    path.as_ref().map(|p| p.path().join("index").join("nodes"))
```
ljeub-pometry: these probably shouldn't have `"index"`

raphtory/src/db/api/mutation/index_ops.rs
```diff
@@ -79,33 +77,32 @@ impl<G: AdditionOps> IndexMutationOps for G {
 
         self.get_storage()
             .map_or(Err(GraphError::IndexingNotSupported), |storage| {
-                if path.is_file() {
-                    if has_index(path)? {
-                        storage.get_or_load_index(path.clone())?;
+                if path.prefer_zip_format {
```
ljeub-pometry: You shouldn't rely on this on load? It is possible for the user to set this to true even when the path is a folder.

raphtory/src/db/api/storage/storage.rs
```diff
@@ -105,83 +109,99 @@ impl Storage {
...
+            }
+        };
+        Ok(())
     }
 }
 
 #[cfg(feature = "search")]
 impl Storage {
     pub(crate) fn get_index_spec(&self) -> Result<IndexSpec, GraphError> {
         let index = self.index.get().ok_or(GraphError::GraphIndexIsMissing)?;
```
ljeub-pometry: should this just return an empty IndexSpec if there is no index?

raphtory/src/db/api/view/graph.rs
```diff
@@ -969,7 +969,10 @@ impl<G: StaticGraphViewOps> SearchableGraphOps for G {
             .get_storage()
             .and_then(|s| s.get_index())
             .ok_or(GraphError::IndexNotCreated)?;
-        index.searcher().search_nodes(self, filter, limit, offset)
+        index
```
ljeub-pometry, line 968: Just use an empty index and rely on the fallback instead of failing? Index should no longer be required for this now.

raphtory/src/search/graph_index.rs
```diff
@@ -25,224 +27,72 @@ use walkdir::WalkDir;
...
-        Ok(())
-    }
-
-    pub(crate) fn persist_to_disk(&self, path: &Path) -> Result<(), GraphError> {
-        let source_path = self.path.as_ref().ok_or(GraphError::GraphIndexIsMissing)?;
+    pub(crate) fn persist_to_disk(&self, path: &GraphFolder) -> Result<(), GraphError> {
+        let source_path = self
+            .path
+            .as_ref()
+            .ok_or(GraphError::GraphIndexIsMissing)?
```
ljeub-pometry: This is the wrong error? This happens if you have an in-memory index...

raphtory/src/search/graph_index.rs
```diff
@@ -255,28 +105,27 @@ impl GraphIndex {
...
+            .read(true)
+            .write(true)
+            .open(path.get_base_path())?;
         let mut zip = ZipWriter::new_append(file)?;
 
-        for entry in WalkDir::new(source_path.path())
+        let source_path = self
+            .path
+            .as_ref()
+            .ok_or(GraphError::GraphIndexIsMissing)?
```
ljeub-pometry: wrong error message?

raphtory/src/search/graph_index.rs
```diff
@@ -360,11 +213,251 @@ impl GraphIndex {
...
+        }))
+    }
+
+    pub fn load_from_path(path: &GraphFolder) -> Result<GraphIndex, GraphError> {
+        let tmp_path = TempDir::new()?;
+        let mut index_path = path.get_index_path();
+        if path.prefer_zip_format {
+            index_path = tmp_path.keep();
+            unzip(&path.get_base_path(), index_path.as_path())?
+        };
```
ljeub-pometry: Several problems with this:
1. don't rely on `path.prefer_zip_format` for loading, there is no guarantee that it matches what is on disk
2. TempDir should only be created if we are actually loading from zip
3. If we are loading from zip and made a TempDir, we should return a mutable index as we have already copied the data!

raphtory/src/search/mod.rs
```diff
@@ -255,6 +295,21 @@ mod test_index {
             assert_search_results(&graph, &filter, vec!["Alice"]);
         }
 
+        #[test]
+        fn test_encoding_graph_zip_twice_overwrites_existing_graph_index_zip() {
```
ljeub-pometry: Is this actually what we would expect to happen? Seems inconsistent with what happens with the folder, I would expect this to fail as well (unless we add an overwrite flag).

raphtory/src/db/api/storage/storage.rs
```diff
@@ -105,83 +109,99 @@ impl Storage {
...
-    pub(crate) fn persist_index_to_disk(&self, path: &PathBuf) -> Result<(), GraphError> {
+    pub(crate) fn persist_index_to_disk(&self, path: &GraphFolder) -> Result<(), GraphError> {
         if let Some(index) = self.get_index() {
-            if index.path.is_none() {
+            if index.read().path().is_none() {
                 info!("{}", IN_MEMORY_INDEX_NOT_PERSISTED);
                 return Ok(());
             }
-            index.persist_to_disk(path)?
+            self.if_index_mut(|index| index.persist_to_disk(path))?;
```
ljeub-pometry: this doesn't need to make the index mutable, you are double-copying the files

raphtory/src/db/api/storage/storage.rs
```diff
@@ -105,83 +109,99 @@ impl Storage {
...
-    pub(crate) fn persist_index_to_disk_zip(&self, path: &PathBuf) -> Result<(), GraphError> {
+    pub(crate) fn persist_index_to_disk_zip(&self, path: &GraphFolder) -> Result<(), GraphError> {
         if let Some(index) = self.get_index() {
-            if index.path.is_none() {
+            if index.read().path().is_none() {
                 info!("{}", IN_MEMORY_INDEX_NOT_PERSISTED);
                 return Ok(());
             }
-            index.persist_to_disk_zip(path)?
+            self.if_index_mut(|index| index.persist_to_disk_zip(path))?;
```
ljeub-pometry: this doesn't need to make the index mutable, you are double-copying the files

shivamka1: ? It make it mutable only if it not already mutable

raphtory/src/db/api/storage/storage.rs
shivamka1: ? It make it mutable only if it not already mutable

raphtory/src/search/mod.rs
shivamka1: Yes, this is what graph_zip encode does

raphtory/src/search/graph_index.rs
shivamka1: If you're already attempting to persist graph, None is not index in RAM anymore

raphtory/src/search/graph_index.rs
shivamka1: If you're already attempting to persist graph, None is not index in RAM anymore

raphtory/src/search/graph_index.rs
shivamka1: 3. No, we don't want the copied graph to be mutable

raphtory/src/search/graph_index.rs
```diff
@@ -287,32 +209,97 @@ impl GraphIndex {
...
-                node_index,
-                edge_index,
-            },
-            path: Arc::new(path.clone()),
-            index_spec: Arc::new(index_spec),
-        }))
+    pub(crate) fn persist_to_disk(&self, path: &GraphFolder) -> Result<(), GraphError> {
+        let source_path = self.path().ok_or(GraphError::CannotPersistRamIndex)?;
+        let path = path.get_index_path();
+        let path = path.as_path();
```
ljeub-pometry, line 238: needed?

PR #2147, opened 2025-06-20
Apply views enums and tests

rachchan:
Add tests for apply views on graph, edge, edges, node and nodes
Change apply views input object to enum


raphtory-graphql/resources/old_index.html
miratepuffin, line 1: Why do we need this? Think it can just be deleted no?

PR #2149, opened 2025-06-23
Expose valid edge filter in python and graphql

ljeub-pometry:

- enable `valid` filter for event graph
- expose `valid` in python and graphql


miratepuffin (approved):
I will open a ticket for valid_graph filtering on edges etc, but otherwise looks great

PR #2152, opened 2025-06-24
New docs examples3

jbaross-pometry:

Adds tests to mkdocs codeblocks that do not require RTD resources

Speed of builds and because we keep hitting RTD limits

- Uses `pytest-markdown-docs` plugin, see  https://pypi.org/project/pytest-markdown-docs/ 
- Updates `pymdown-extensions ` tabbed codeblocks to newer syntax for compatibility with test parser. See [pymdown docs](https://facelessuser.github.io/pymdown-extensions/extensions/blocks/plugins/tab/) for details 
- Updates code examples to be runable stand-alone, this is checked by default when tests are run
- Where possible tests for correct output were added in hidden codeblocks using superfence syntax and hide-code extension. 
- Some examples need to be rewitten anyway and this will improve coverage. The GraphQL topics were to fragmented to add any tests.

Tested locally but github action will need to be tested on the main new_docs branch


jbaross-pometry:
RTD build fails beacuse `mkdocs get-deps` doesn't list https://github.com/staadecker/markdown-hide-code correctly despite being in the mkdocs.yml file

PR #2157, opened 2025-06-27
Add GIDResolver type alias

fabubaker:


db4-graph/src/lib.rs
```diff
@@ -224,9 +227,15 @@ impl<'a, EXT: PersistentStrategy<NS = NS<EXT>, ES = ES<EXT>>> WriteLockedGraph<'
...
     pub fn resolve_node(&self, gid: GidRef) -> Result<MaybeNew<VID>, InvalidNodeId> {
-        self.graph.logical_to_physical.get_or_init(gid, || {
+        let result = self.graph.logical_to_physical.get_or_init(gid, || {
             VID(self.num_nodes.fetch_add(1, atomic::Ordering::Relaxed))
-        })
+        });
+
+        match result {
+            Ok(vid) => Ok(vid),
+            Err(GIDResolverError::DBV4Error(e)) => panic!("Database error: {}", e),
```
fabubaker, line 258: I don't like panicking here, but this means having to change the error return type away from `InvalidNodeId`.

PR #2167, opened 2025-07-03
Fix index update

shivamka1:
Avoid indexing node/edge fields while index prop update.


ljeub-pometry (changes requested):
When you create a new index in the storage it is still calling update as well I think which now double-indexes everything. You need to update that as well.

ljeub-pometry (approved):
LGTM

PR #2179, opened 2025-07-09
Adding some tests for valid graph

miratepuffin:
Tests currently missing and want to make sure is correct for 16
Do we want to add these auto gens as they should only change if the docs change?

### DIFF
--- a/python/tests/test_base_install/test_graphdb/test_persistent_graph.py
+++ b/python/tests/test_base_install/test_graphdb/test_persistent_graph.py
@@ -197,6 +197,100 @@ def test_graph_type_swap():
     assert eg.event_graph().at(2).count_edges() == 1
 
 
+from raphtory import PersistentGraph
+from raphtory import filter
+from raphtory import Prop
+
+
+def test_basic_valid():
+    g = PersistentGraph()
+    g.add_edge(1, 1, 2)
+    g.add_edge(2, 1, 3)
+    g.delete_edge(10, 1, 3)
+
+    g.add_edge(30, 1, 4)
+
+    assert g.count_edges() == 3
+    assert g.valid().count_edges() == 2
+    assert g.at(9).valid().count_edges() == 2
+
+
+def test_properties_valid():
+    g = PersistentGraph()
+    g.add_edge(1, 1, 2, layer="red", properties={"weight": 1})
+    g.add_edge(2, 1, 2, layer="green", properties={"weight": 2})
+    g.add_edge(3, 1, 2, layer="blue", properties={"weight": 3})
+    g.delete_edge(4, 1, 2, layer="red")
+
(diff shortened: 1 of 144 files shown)


python/tests/test_base_install/test_graphdb/test_persistent_graph.py
```diff
@@ -197,6 +197,100 @@ def test_graph_type_swap():
...
+    assert list(g.edge(1, 2).properties.temporal.get("weight").values()) == [1, 2, 3]
+    assert list(g.valid().edge(1, 2).properties.temporal.get("weight").values()) == [
+        2,
+        3,
+    ]
+
+    assert g.edge(1, 2).is_deleted() == False
+    assert g.valid().count_edges() == 1
+    assert g.layer("red").edge(1, 2).is_deleted() == True
+    assert g.layer("red").valid().count_edges() == False
```
ljeub-pometry: 0 instead of False?

python/tests/test_base_install/test_graphdb/test_persistent_graph.py
```diff
@@ -197,6 +197,100 @@ def test_graph_type_swap():
...
+    f = filter.Property("weight").temporal().latest() < 3
+    e = g.filter_edges(f).edge(1, 2)
+    assert e == None
+    # assert list(e.properties.temporal.get("weight").values) == [1,2] -- this would be the case for filter_edge_layer?
+
+    # f = filter.Property("weight") < 3
+    f = Prop("weight") < 3  # this needs to be converted to the above
+    e = (
+        g.valid().filter_exploded_edges(f).edge(1, 2)
+    )  # this returns an edge which is invalid -- this would not be the case with Shivams changes
```
ljeub-pometry: this will still be the case after the changes, update this comment as `valid` is a filter, not a time semantic (i.e., valid only sees the graph as it was at the point when it is called, filters afterwards have no effect on the decision)

python/tests/test_base_install/test_graphdb/test_persistent_graph.py
```diff
@@ -197,6 +197,100 @@ def test_graph_type_swap():
...
+        2,
+    ]  # property values the same
+
+    f = Prop("weight") > 5  # this needs to be converted to the above
+    e = g.filter_exploded_edges(f).edge(
+        1, 2
+    )  # this should be probably be filtered out (along with edges that only have deletions) as we discussed
+    assert list(e.properties.temporal.get("weight").values()) == []
+
+
```
ljeub-pometry, line 293: What's going on with the old commented out test below? Is that fixed? Some of the asserts are wrong for the new semantics for deletions at the start of the window but I am not aware of any issues with hanging edges as such right now.

PR #2184, opened 2025-07-11
bring exploded_edge_filter input inline with other filters in python

ljeub-pometry:

- `filter_exploded_edges` now take `FilterExpr` as input in python
- the old `Prop("name")` api is gone, use `filter.Property("name")` instead

Consistency with other apis

yes, the input for `filter_exploded_edges` is changed in python.


python/tests/test_base_install/test_graphdb/test_property_filters.py
```diff
@@ -98,23 +98,26 @@ def test_filter_exploded_edges():
...
+        (filter.Property("test_int") == 2, [(3, 4)]),
+        (filter.Property("test_int") != 2, [(1, 2), (2, 3), (3, 4)]),
+        (filter.Property("test_int") > 2, [(3, 4)]),
+        (filter.Property("test_int") >= 2, [(3, 4)]),
+        (filter.Property("test_int") < 3, [(1, 2), (2, 3), (3, 4)]),
+        (filter.Property("test_int") <= 1, [(1, 2), (2, 3)]),
+        (
+            filter.Property("test_bool") == True,
+            [(2, 3)],
+        ),  # worth adding special support for this?
```
miratepuffin, line 120: This comment seems not needed any more?

raphtory/src/python/types/wrappers/prop.rs
```diff
@@ -229,7 +178,25 @@ impl CreateEdgeFilter for PyFilterExpr {
         match self.0 {
             PyInnerFilterExpr::Edge(i) => i.create_edge_filter(graph),
             PyInnerFilterExpr::Property(i) => i.create_edge_filter(graph),
-            PyInnerFilterExpr::Node(_) => Err(GraphError::ParsingError),
+            PyInnerFilterExpr::Node(_) => Err(GraphError::NodeFilterIsNotEdgeFilter),
+        }
+    }
+}
+
+impl CreateExplodedEdgeFilter for PyFilterExpr {
```
miratepuffin: Do you not thing this would make more sense at a top level alongside the Node and Edge filter creators?

python/tests/test_base_install/test_graphdb/test_property_filters.py
ljeub-pometry: I think this is talking about having this work without the `== True` bit for booleans

raphtory/src/python/types/wrappers/prop.rs
ljeub-pometry: Yes it would

miratepuffin (approved):

PR #2187, opened 2025-07-14
Fixed the exploded edge or

miratepuffin:

### DIFF
--- a/python/tests/test_base_install/test_filters/test_exploded_edge_filter.py
+++ b/python/tests/test_base_install/test_filters/test_exploded_edge_filter.py
@@ -1,0 +1,489 @@
+from raphtory import Graph
+from raphtory import filter
+import pytest
+from datetime import datetime
+
+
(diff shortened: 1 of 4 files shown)


python/tests/test_base_install/test_filters/test_exploded_edge_filter.py
```diff
@@ -0,0 +1,898 @@
...
+        assert e1.deletions() == []
+        assert e2.deletions() == []
+    else:
+        assert e1.deletions() == [1, 2]
+        assert e2.deletions() == [1, 2]
+
+    assert list(e1.history()) == [3]
+    assert list(e2.history()) == [3]
+
+    # assert e2.layer_names == ["red"] returning red blue for PersistentGraph which feels wrong?
```
ljeub-pometry, line 37: that is consistent with how we are handling edges that are only deleted right now (so probably ok of this version).

PR #2189, opened 2025-07-15
Expose K-Core to Python API

wyatt-joyner-pometry:
Add K-Core to exposed algorithms

The more the merrier.

Python signature contains relevant docstring.

Tested with the test_algorithms suite.

No.

### DIFF
--- a/python/python/raphtory/__init__.pyi
+++ b/python/python/raphtory/__init__.pyi
@@ -47,6 +47,7 @@ __all__ = [
     "WindowSet",
     "IndexSpecBuilder",
     "IndexSpec",
+    "DiskGraphStorage",
     "graphql",
     "algorithms",
     "graph_loader",
@@ -1385,6 +1386,17 @@ class Graph(GraphView):
           MutableNode: The node object with the specified id, or None if the node does not exist
         """
 
+    def persist_as_disk_graph(self, graph_dir: str | PathLike) -> DiskGraphStorage:
+        """
+        save graph in disk_graph format and memory map the result
+
+        Arguments:
+            graph_dir (str | PathLike): folder where the graph will be saved
+
+        Returns:
+            DiskGraphStorage: the persisted graph storage
+        """
+
     def persistent_graph(self) -> PersistentGraph:
         """
         View graph with persistent semantics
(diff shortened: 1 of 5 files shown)


python/python/raphtory/__init__.pyi
```diff
@@ -1385,6 +1386,17 @@ class Graph(GraphView):
           MutableNode: The node object with the specified id, or None if the node does not exist
         """
 
+    def persist_as_disk_graph(self, graph_dir: str | PathLike) -> DiskGraphStorage:
```
ljeub-pometry: this shouldn't be here, run `tidy-public`

python/python/raphtory/algorithms/__init__.pyi
```diff
@@ -628,6 +630,23 @@ def label_propagation(
 
     """
 
+def k_core(
+    graph: GraphView, k: int, iter_count: int, threads: Optional[int] = None
+) -> list[Node]:
```
ljeub-pometry, line 634: return Nodes instead?

raphtory/src/python/packages/algorithms.rs
```diff
@@ -747,6 +748,31 @@ pub fn label_propagation(
...
+///     list[Node]: A list of nodes in the k core
+///
+#[pyfunction]
+#[pyo3[signature = (graph, k, iter_count, threads=None)]]
+pub fn k_core(
+    graph: &PyGraphView,
+    k: usize,
+    iter_count: usize,
+    threads: Option<usize>,
+) -> Vec<NodeView<'static, DynamicGraph>> {
```
ljeub-pometry: this should return Nodes instead of Vec<NodeView>

wyatt-joyner-pometry: Can do.

wyatt-joyner-pometry:
> Change the return type to `Nodes`

Should be returning `Nodes<'static, NodeSubGraph<DynamicGraph>>` now.

ljeub-pometry:
> > Change the return type to `Nodes`
> 
> Should be returning `Nodes<'static, NodeSubGraph<DynamicGraph>>` now.

Actually, given that type, shouldn’t we just return the NodeSubgraph?

wyatt-joyner-pometry:
> > > Change the return type to `Nodes`
> > 
> > 
> > Should be returning `Nodes<'static, NodeSubGraph<DynamicGraph>>` now.
> 
> Actually, given that type, shouldn’t we just return the NodeSubgraph?

Sure!

PR #2201, opened 2025-07-29
Batch generate embeddings in `add_nodes` / `add_edges`

fabubaker:
Currently, `GqlMutableGraph.add_nodes` / `GqlMutableGraph.add_edges` repeatedly calls the embedding function for each node/edge. This PR modifies it to use the function just once for an entire batch of nodes/edges.

Optimizes how embeddings are created, for eg, by using a batch call to OpenAI.

No.

Unit tests were added to `mutable_graph.rs`.

No.

### DIFF
--- a/raphtory-graphql/schema.graphql
+++ b/raphtory-graphql/schema.graphql
@@ -344,8 +344,8 @@ type Graph {
 }
 
 type GraphAlgorithmPlugin {
-	pagerank(iterCount: Int!, threads: Int, tol: Float): [PagerankOutput!]!
 	shortest_path(source: String!, targets: [String!]!, direction: String): [ShortestPathOutput!]!
+	pagerank(iterCount: Int!, threads: Int, tol: Float): [PagerankOutput!]!
 }
 
 type GraphSchema {
--- a/raphtory-graphql/schema.graphql
+++ b/raphtory-graphql/schema.graphql
@@ -344,8 +344,8 @@ type Graph {
 }
 
 type GraphAlgorithmPlugin {
-	shortest_path(source: String!, targets: [String!]!, direction: String): [ShortestPathOutput!]!
 	pagerank(iterCount: Int!, threads: Int, tol: Float): [PagerankOutput!]!
+	shortest_path(source: String!, targets: [String!]!, direction: String): [ShortestPathOutput!]!
 }
 
 type GraphSchema {
--- a/raphtory-graphql/schema.graphql
+++ b/raphtory-graphql/schema.graphql
@@ -1,0 +1,1069 @@
+enum AllPropertySpec {
+	ALL
+	ALL_METADATA
+	ALL_PROPERTIES
+}
+
+
+type CollectionOfMetaGraph {
+	list: [MetaGraph!]!
+	"""
+	Fetch one "page" of items, optionally offset by a specified amount.
+	
+	* `limit` - The size of the page (number of items to fetch).
+	* `offset` - The number of items to skip (defaults to 0).
+	* `page_index` - The number of pages (of size `limit`) to skip (defaults to 0).
+	
+	e.g. if page(5, 2, 1) is called, a page with 5 items, offset by 11 items (2 pages of 5 + 1),
+	will be returned.
+	"""
+	page(limit: Int!, offset: Int, pageIndex: Int): [MetaGraph!]!
+	count: Int!
+}
+
+type CollectionOfNamespace {
+	list: [Namespace!]!
+	"""
+	Fetch one "page" of items, optionally offset by a specified amount.
+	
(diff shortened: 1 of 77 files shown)


raphtory-graphql/src/model/graph/mutable_graph.rs
```diff
@@ -544,3 +542,178 @@ impl GqlMutableEdge {
...
+                node_type: Some("test_node_type".to_string()),
+                metadata: None,
+                updates: None,
+            },
+        ];
+
+        let result = mutable_graph.add_nodes(nodes).await;
+
+        assert!(result.is_ok());
+        assert!(result.unwrap());
```
ricopinazo, line 651: Ideally, here we would also do some search against the vectorised graph and confirm the vectors were ingested. You should be able to do `mutable_graph.graph.vectors.unwrap().nodes_by_similarity()` or any other method available for VectorisedGraph

raphtory-graphql/src/model/graph/mutable_graph.rs
```diff
@@ -544,3 +542,178 @@ impl GqlMutableEdge {
...
+    #[tokio::test]
+    async fn test_add_nodes_simple() {
+        let (mutable_graph, _tmp_dir) = create_mutable_graph().await;
+
+        let nodes = vec![
+            NodeAddition {
+                name: "node1".to_string(),
+                node_type: Some("test_node_type".to_string()),
+                metadata: None,
+                updates: None,
```
ricopinazo: I think for this to make sense, there has to be at least one update, but it doesn't need to hold any property value. You can set one update with time 0 and properties None. Otherwise, this does nothing by looking at the implementation. Not sure if this has to do with the error you were getting.

fabubaker: Done.

raphtory-graphql/src/model/graph/mutable_graph.rs
fabubaker: Done.

raphtory/src/vectors/vectorised_graph.rs
```diff
@@ -56,6 +78,28 @@ impl<G: StaticGraphViewOps> VectorisedGraph<G> {
...
+
+                        self.template.edge(edge).map(|doc| (id, doc))
+                    })
+            })
+            .unzip();
+
+        let vectors = self.cache.get_embeddings(docs).await?;
+
+        for (id, vector) in ids.iter().zip(vectors) {
+            self.edge_db.insert_vector(*id, &vector)?;
```
ricopinazo: Every time you call `insert_vector` this is going to update the index. We should probably change that function to accept a vector of embeddings instead, so we can write everything to arroy and only in the end update the index just once. The same applies to nodes

ricopinazo (changes requested):

raphtory/src/vectors/vectorised_graph.rs
fabubaker: Done

raphtory-graphql/src/model/graph/mutable_graph.rs
```diff
@@ -247,18 +245,24 @@ impl GqlMutableGraph {
                     .collect()
             })
             .await;
-        for edge in edges {
-            let _ = edge?.update_embeddings().await; // FIXME: ideally this should call the embedding function just once!!
-        }
+
+        // Generate embeddings
+        let edges: Vec<_> = edges.into_iter().collect::<Result<Vec<_>, _>>()?;
```
ricopinazo: I'm not sure if might be posible to collect here just once

raphtory-graphql/schema.graphql
```diff
@@ -344,8 +344,8 @@ type Graph {
 }
 
 type GraphAlgorithmPlugin {
-	shortest_path(source: String!, targets: [String!]!, direction: String): [ShortestPathOutput!]!
 	pagerank(iterCount: Int!, threads: Int, tol: Float): [PagerankOutput!]!
+	shortest_path(source: String!, targets: [String!]!, direction: String): [ShortestPathOutput!]!
```
ricopinazo: Just letting a comment here to remember we should make sure whatever generates de GraphQL schema file should do it in a predictable way to avoid these kinds of conflicts

ricopinazo (approved):
LGTM, just left a small comment that you might want to have a look at, but it's not the end of the world

raphtory-graphql/src/model/graph/mutable_graph.rs
fabubaker: The first collect uses `?` to return the first error it encounters in `edges`, so both `collect` are needed.
That said I did collapse both into one chain and got rid of the intermediate variable.

PR #2221, opened 2025-08-07
fix deadlock on filtered_edges_iter

ricopinazo:


raphtory/src/db/api/view/internal/time_semantics/filtered_node.rs
fabianmurariu, line 1: Fix `filtered_par_iter` as well please

fabianmurariu (commented):
LGTM

PR #2249, opened 2025-08-28
embedding api improvements

ricopinazo:

- [ ] make sure we actually saving the templates on disk in a test
- [x] maybe embeddings should not be enabled by default, which simply ignores any vectors on disk
- [x] make sure we have the appropriate blocking_io calls, especially around read/write operations on arroy
- [x] there might be some corner case where some old embedding sample is read from the cache but the actual model changed
- [x] rename scores to distances in rust/python/graphql apis
- [ ] forbid graph names starting with a dot
- [x] make sure to add a graphql function to vectorise graphs on demand
- [x] remove arroy dependency
- [ ] make sure we align our policy to ignore or not embedding errors so they are consistent across all APIs. Ideally, we would ignore them always, and simply retry embeddings later on so that the vectorised graph doesn't get out of sync.
- [x] enable Ctrl+C for the embedding server
- [x] Revisit execute_async_task and block_on in src/python/utils/mod.rs, should just need on of them, probably don't need a thread
- [x] Make sure a failing python embedding doesn't kill the embedding server or handle those appropriately
- [x] Stop eagerly writing out the vector cache folder
- [x] expose re-index api


.github/workflows/bench-graphql.yml
```diff
@@ -27,14 +27,16 @@ jobs:
       - name: Install maturin
         run: pip install maturin==1.8.3
       - name: Build raphtory
-        run: make install-python
+        run: cd python && maturin build --release --out /tmp/wheels # If this fails, try adding --strip back again
```
fabianmurariu: Why are we moving out of make? at best we would put this in a bash script

.github/workflows/stress-test.yml
```diff
@@ -24,14 +24,16 @@ jobs:
       - name: Install maturin
         run: pip install maturin==1.8.3
       - name: Build raphtory
-        run: make install-python
+        run: cd python && maturin build --strip --out /tmp/wheels # double-check --strip acutally makes sense
```
fabianmurariu: can we put this back into make or in a bash script?

raphtory-graphql/src/model/graph/vector_selection.rs
```diff
@@ -100,60 +91,58 @@ impl GqlVectorSelection {
...
-        let vector = ctx.embed_query(query).await?;
+        let vector = self.embed_text(query).await?;
         let window = window.into_window_tuple();
         let mut selection = self.cloned();
-        blocking_compute(move || {
-            selection.expand_edges_by_similarity(&vector, limit, window)?;
-            Ok(selection.into())
-        })
-        .await
+        selection
```
fabianmurariu, line 133: make sure we don't interact with raphtory otherwise we get deadlocks

raphtory-graphql/src/model/graph/vectorised_graph.rs
```diff
@@ -45,42 +49,42 @@ impl GqlVectorisedGraph {
...
         limit: usize,
         window: Option<VectorisedGraphWindow>,
     ) -> GraphResult<GqlVectorSelection> {
-        let vector = ctx.embed_query(query).await?;
+        let vector = self.0.embed_text(query).await?;
         let w = window.into_window_tuple();
         let cloned = self.0.clone();
-        blocking_io(move || Ok(cloned.entities_by_similarity(&vector, limit, w)?.into())).await
+        Ok(cloned
+            .entities_by_similarity(&vector, limit, w)
```
fabianmurariu: make sure we don't interact with raphtory otherwise we get deadlocks

raphtory-graphql/src/model/graph/vectorised_graph.rs
```diff
@@ -45,42 +49,42 @@ impl GqlVectorisedGraph {
...
         query: String,
         limit: usize,
         window: Option<VectorisedGraphWindow>,
     ) -> GraphResult<GqlVectorSelection> {
-        let vector = ctx.embed_query(query).await?;
+        let vector = self.0.embed_text(query).await?;
         let w = window.into_window_tuple();
         let cloned = self.0.clone();
-        blocking_io(move || Ok(cloned.nodes_by_similarity(&vector, limit, w)?.into())).await
+        Ok(cloned.nodes_by_similarity(&vector, limit, w).await?.into())
```
fabianmurariu: make sure we don't interact with raphtory otherwise we get deadlocks

raphtory-graphql/src/graph.rs
```diff
@@ -87,9 +89,11 @@ impl GraphWithVectors {
...
         };
-        let vectors = cache.and_then(|cache| {
-            VectorisedGraph::read_from_path(&folder.get_vectors_path(), graph.clone(), cache).ok()
-        });
+        let vectors =
+            VectorisedGraph::read_from_path(&folder.get_vectors_path(), graph.clone(), cache)
+                .await
+                .ok();
+
         println!("Graph loaded = {}", folder.get_original_path_str());
```
fabianmurariu, line 97: We really should sort out our logging

raphtory/src/vectors/vector_collection/lancedb.rs
```diff
@@ -0,0 +1,385 @@
...
+        let stream = filtered.execute().await?;
+        let result = stream.try_collect::<Vec<_>>().await?;
+
+        let downcasted = result
+            .into_iter()
+            .map(|record| {
+                let ids = primitive_column_iter::<UInt64Type>(&record, "id")?;
+                let scores = primitive_column_iter::<Float32Type>(&record, "_distance")?;
+                // TODO: try using record.columns() instead of getting them independently?
+                Some(ids.zip(scores).collect::<Vec<_>>())
```
fabianmurariu: there are ways in which you don't need to collect here, 
```rust
0..array.len().filter_map(move |i| array.get(i)) 
```

raphtory/src/vectors/vector_collection/lancedb.rs
```diff
@@ -0,0 +1,385 @@
...
+            })
+            .collect::<Option<Vec<_>>>()
```
fabianmurariu, line 157: likely you don't need to collect here either

raphtory/src/vectors/vector_collection/lancedb.rs
```diff
@@ -0,0 +1,385 @@
...
+        vectors: impl IntoIterator<Item = Embedding>,
+    ) -> crate::errors::GraphResult<()> {
+        let mut builder = FixedSizeListBuilder::new(Float32Builder::new(), self.dim as i32);
+        for vector in vectors {
+            builder.values().append_slice(&vector);
+        }
+        let batches = RecordBatchIterator::new(
+            vec![RecordBatch::try_new(
+                self.schema(),
+                vec![Arc::new(UInt64Array::from(ids)), Arc::new(builder.finish())],
```
fabianmurariu, line 96: maybe missing `?` here?

raphtory/src/vectors/vector_selection.rs
```diff
@@ -115,25 +122,29 @@ impl<G: StaticGraphViewOps> VectorSelection<G> {
...
-            .iter()
-            .map(|(entity, score)| self.regenerate_doc(*entity).map(|doc| (doc, *score)))
-            .collect()
+    /// Return the documents alongside their distances present in the current selection
+    pub async fn get_documents_with_distances(&self) -> GraphResult<Vec<(Document<G>, f32)>> {
+        let futures = self.selected.iter().map(|(entity, distance)| async {
+            self.regenerate_doc(*entity)
+                .await
+                .map(|doc| (doc, *distance))
+                .unwrap() // TODO: REMOVE UNWRAP
```
fabianmurariu: There's quite a few TODO's

raphtory/src/vectors/vector_collection/milvus.rs
fabianmurariu, line 1: Do we need this?

raphtory/src/vectors/vector_collection/lancedb.rs
ricopinazo: couldn't remove the first collect. Claude just gives up and simply moves the collect somewhere else and I can't think of any alternative. The second collect is necessary since we want to return an error if some column is missing in any of the records, so we need to know all of them beforehand. I left some comments to clarify all this

python/tests/test_base_install/test_vectors.py
```diff
@@ -77,7 +117,9 @@ def test_selection():
     nodes = selection.nodes()
     node_names_returned = [node.name for node in nodes]
     assert node_names_returned == nodes_to_select
+    print("before get documents")
```
ricopinazo: should remove this

.github/workflows/bench-graphql.yml
```diff
@@ -27,14 +27,19 @@ jobs:
       - name: Install maturin
         run: pip install maturin==1.8.3
       - name: Build raphtory
-        run: make install-python
+        run: |
```
fabianmurariu, line 30: why did we move this out of make?

.github/workflows/stress-test.yml
```diff
@@ -24,14 +24,19 @@ jobs:
       - name: Install maturin
         run: pip install maturin==1.8.3
       - name: Build raphtory
-        run: make install-python
+        run: |
```
fabianmurariu, line 27: can we have this in make or in a script?

python/tests/test_base_install/test_vectors.py
```diff
@@ -12,19 +17,46 @@
...
+def post_json(url, payload):
+    data = json.dumps(payload).encode()
+    req = Request(
+        url,
+        data=data,
+        headers={"Content-Type": "application/json"},
+        method="POST",
+    )
+    try:
+        with urlopen(req, timeout=10) as r:
```
fabianmurariu: On a slow github day this might fail, maybe bump it up to 50?

python/tests/test_base_install/test_vectors.py
```diff
@@ -12,19 +17,46 @@
...
+        payload = {"model": "whatever", "input": ["Hello world"]}
+        status, _ = post_json("http://localhost:7342/embeddings", payload)
+        assert status == 500
+        status, _ = post_json("http://localhost:7342/embeddings", payload)
+        assert status == 500
 
 
 def floats_are_equals(float1: float, float2: float) -> bool:
-    return float1 + 0.001 > float2 and float1 - 0.001 < float2
+    return float1 + 0.00001 > float2 and float1 - 0.01 < float2
```
fabianmurariu: https://stackoverflow.com/a/16925316

```python
numpy.isclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False)
```

raphtory-graphql/src/model/graph/mutable_graph.rs
```diff
@@ -598,75 +598,88 @@ impl GqlMutableEdge {
...
         let mutable_graph = GqlMutableGraph::new(path, graph_with_vectors);
 
-        (mutable_graph, tmp_dir)
+        (mutable_graph, tmp_dir, embedding_server)
     }
 
     #[tokio::test]
     async fn test_add_nodes_empty_list() {
-        let (mutable_graph, _tmp_dir) = create_mutable_graph().await;
+        let (mutable_graph, _tmp_dir, _embedding_server) = create_mutable_graph(1745).await;
```
fabianmurariu: _embedding_server is used, so maybe call it embedding_server, this would be fine as (mutable_graph, _, embedding_server)

raphtory/src/python/packages/vectors.rs
```diff
@@ -3,33 +3,165 @@ use crate::{
...
+            project_id: value.project_id.clone(),
+        }
+    }
+}
+
+impl EmbeddingFunction for Arc<Py<PyFunction>> {
+    fn call(&self, text: &str) -> Vec<f32> {
+        Python::with_gil(|py| {
+            // TODO: remove unwraps?
+            let any = self.call1(py, (text,)).unwrap();
```
fabianmurariu, line 110: you can return Result<T, PyErr>

raphtory/src/vectors/entity_db.rs
```diff
@@ -0,0 +1,137 @@
...
+        &self.0
+    }
+
+    fn into_entity_ref(id: u64) -> EntityRef {
+        EntityRef::Edge(id)
+    }
+
+    fn view_has_entity<G: StaticGraphViewOps>(entity: &EntityRef, view: &G) -> bool {
+        let (src, dst) = entity.as_edge_gids(view).unwrap();
+        view.has_edge(src, dst) // TODO: there should be a quicker way of chking of some edge exist by pid
```
fabianmurariu, line 69: if you have the edge_id you can make EdgeView::new_filtered(id, view) then call is_active

raphtory/src/python/utils/mod.rs
```diff
@@ -411,24 +412,28 @@ impl<'py> IntoPyObject<'py> for NumpyArray {
...
-                    .unwrap()
-                    .block_on(task())
-            })
-            .join()
-            .expect("error when waiting for async task to complete")
-        })
+    Python::with_gil(|py| py.allow_threads(move || get_runtime().block_on(task())))
+}
+
+static RUNTIME: OnceLock<Runtime> = OnceLock::new();
```
ricopinazo, line 424: probably now that this block_on function is being used not only for python we should me it somewhere else alongside the global RUNTIME

PR #2253, opened 2025-09-04
implemented erdos renyl model generation

DanielLacina:


raphtory/src/graphgen/erdos_renyi.rs
```diff
@@ -0,0 +1,112 @@
...
+/// - Adds `n_nodes` nodes to the graph.
+/// - For each pair of distinct nodes, adds a directed edge with probability `p`.
+/// - Uses the provided seed for reproducibility if given.
+///
+/// # Example
+/// ```
+/// let graph = Graph::new();
+/// erdos_renyi(&graph, 10, 0.2, None);
+/// ```
+pub fn erdos_renyi(graph: &Graph, n_nodes: usize, p: f64, seed: Option<[u8; 32]>) {
```
miratepuffin: This is kind of a strange seed, not possible to use just a u64?

raphtory/src/graphgen/erdos_renyi.rs
```diff
@@ -0,0 +1,112 @@
...
+    let mut max_id = next_id(graph, ids.iter().max().cloned());
+    for _ in 0..n_nodes {
+        max_id = next_id(graph, Some(max_id));
+        latest_time += 1;
+        graph
+            .add_node(latest_time, &max_id, NO_PROPS, None)
+            .map_err(|err| error!("{:?}", err))
+            .ok();
+    }
+    let all_ids = graph.nodes().id().iter_values().collect::<Vec<_>>();
```
miratepuffin: You already have this above - this could be quite a big vec - would be better to reuse

raphtory/src/python/packages/graph_gen.rs
```diff
@@ -67,3 +67,23 @@ pub fn ba_preferential_attachment(
 ) {
     pa(&g.graph, nodes_to_add, edges_per_step, seed);
 }
+
+/// Generates a graph using the Erdos-Renyi random graph model.
```
miratepuffin: Looks great! As above seed probably should be a singular number

raphtory/src/graphgen/erdos_renyi.rs
```diff
@@ -0,0 +1,112 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::prelude::*;
+
+    #[test]
+    fn test_erdos_renyi_small_graph() {
```
miratepuffin: Wanna add some tests where you add 100,000 nodes or somethign else reasonably big.
Also a good idea to look at adding some tests for graphs which already have nodes and edges present.

raphtory/src/graphgen/erdos_renyi.rs
```diff
@@ -0,0 +1,112 @@
...
+    for _ in 0..n_nodes {
+        max_id = next_id(graph, Some(max_id));
+        latest_time += 1;
+        graph
+            .add_node(latest_time, &max_id, NO_PROPS, None)
+            .map_err(|err| error!("{:?}", err))
+            .ok();
+    }
+    let all_ids = graph.nodes().id().iter_values().collect::<Vec<_>>();
+    for id in all_ids.iter() {
```
miratepuffin: You should be able to do this in parallel over these iterators

raphtory/src/graphgen/erdos_renyi.rs
```diff
@@ -0,0 +1,112 @@
...
+        for other_id in all_ids.iter() {
+            if id != other_id && rng.gen::<f64>() < p {
+                latest_time += 1;
+                graph
```
miratepuffin: Do we consider that an edge may already be present?

raphtory/src/graphgen/erdos_renyi.rs
```diff
@@ -0,0 +1,112 @@
...
+    } else {
+        rng = StdRng::from_entropy();
+    }
+    let mut latest_time = graph.latest_time().unwrap_or(0);
+    let ids = graph.nodes().id().iter_values().collect::<Vec<_>>();
+    let mut max_id = next_id(graph, ids.iter().max().cloned());
+    for _ in 0..n_nodes {
+        max_id = next_id(graph, Some(max_id));
+        latest_time += 1;
+        graph
```
miratepuffin, line 59: ID's in raphtory can be either strings or ints - this is assuming that the graph either has no nodes or that those present are indexed by int - I think this will explode if you already have some string Ids present - should add a test for this

raphtory/src/graphgen/erdos_renyi.rs
```diff
@@ -0,0 +1,112 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::prelude::*;
+
+    #[test]
+    fn test_erdos_renyi_small_graph() {
+        let graph = Graph::new();
+        let n_nodes = 5;
```
miratepuffin: Could also add some tests to see if the graphs have expected stats e.g. density

raphtory/src/graphgen/erdos_renyi.rs
```diff
@@ -0,0 +1,112 @@
...
+        max_id = next_id(graph, Some(max_id));
+        latest_time += 1;
+        graph
+            .add_node(latest_time, &max_id, NO_PROPS, None)
+            .map_err(|err| error!("{:?}", err))
+            .ok();
+    }
+    let all_ids = graph.nodes().id().iter_values().collect::<Vec<_>>();
+    for id in all_ids.iter() {
+        for other_id in all_ids.iter() {
```
miratepuffin: Erdos Renyi graphs are typically undirected - here you are going to loop over each edge both ways so each edge gets checked twice meaning you may get double the density you are expecting. 

A better solution could be to order the id's and then only check edges for src.id < dst.id  - this would also remove self loops (<= if you want to include these)

raphtory/src/graphgen/erdos_renyi.rs
DanielLacina: We need all the ids to be stored in a vector so the algorithm knows about all the other node ids ahead of time.

raphtory/src/graphgen/erdos_renyi.rs
DanielLacina: what if the ids are strings?

DanielLacina: networkx has a directed parameter

raphtory/src/graphgen/erdos_renyi.rs
DanielLacina: Could I delete an edge if its present?

raphtory/src/graphgen/erdos_renyi.rs
ljeub-pometry: have a look at this: https://docs.rs/rand_core/latest/rand_core/trait.SeedableRng.html#method.seed_from_u64

DanielLacina: yeah I'm refactoring everything atm

raphtory/src/graphgen/erdos_renyi.rs
ljeub-pometry: you generate the vector twice, you already have it as `ids` above

DanielLacina:
@ljeub-pometry

raphtory/src/graphgen/erdos_renyi.rs
DanielLacina: ig ignore it

raphtory/src/graphgen/erdos_renyi.rs
DanielLacina: like a hypothesis test?

raphtory/src/graphgen/erdos_renyi.rs
```diff
@@ -0,0 +1,116 @@
...
+    }
+    let mut ids = graph.nodes().id().par_iter_values().collect::<Vec<GID>>();
+    let mut latest_time = graph.latest_time().map_or(0, |t| t.t());
+    let mut max_id = next_id(graph, graph.nodes().id().iter_values().max());
+    for _ in 0..nodes_to_add {
+        max_id = next_id(graph, Some(max_id));
+        latest_time += 1;
+        graph
+            .add_node(latest_time, &max_id, NO_PROPS, None)
+            .map_err(|err| error!("{:?}", err))
```
ljeub-pometry: You should return a `Result<(), GraphError>` from this function so you can handle the errors properly

raphtory/src/graphgen/erdos_renyi.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    #[test]
+    fn test_erdos_renyi_zero_probability() {
+        let graph = Graph::new();
+        let n_nodes = 20;
+        let p = 0.0;
+        let seed = Some(42); 
+        erdos_renyi(&graph, n_nodes, p, seed);
+        let edge_count = graph.edges().into_iter().count();
+        assert_eq!(edge_count, 0);
```
ljeub-pometry, line 104: you should assert that the graph actually has 20 nodes as well

raphtory/src/graphgen/erdos_renyi.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    #[test]
+    fn test_erdos_renyi_full_probability() {
+        let graph = Graph::new();
+        let n_nodes = 20;
+        let p = 1.0;
+        let seed = Some(42);
+        erdos_renyi(&graph, n_nodes, p, seed);
+        let edge_count = graph.edges().into_iter().count();
+        assert_eq!(edge_count, (n_nodes * (n_nodes - 1))/2);
```
ljeub-pometry: you should assert that the graph actually has 20 nodes as well

raphtory/src/graphgen/erdos_renyi.rs
```diff
@@ -0,0 +1,116 @@
...
+/// - Adds `n_nodes` nodes to the graph.
+/// - For each pair of distinct nodes, adds a directed edge with probability `p`.
+/// - Uses the provided seed for reproducibility if given.
+///
+/// # Example
+/// ```
+/// let graph = Graph::new();
+/// erdos_renyi(&graph, 10, 0.2, None);
+/// ```
+pub fn erdos_renyi(graph: &Graph, nodes_to_add: usize, p: f64, seed: Option<u64>) {
```
ljeub-pometry: Maybe it would be better if this doesn't take a graph as input and makes a new graph internally? That way we don't have to worry about weird things like other threads modifying the graph at the same time or string ids.

DanielLacina: Yeah that would make things easier

raphtory/src/graphgen/erdos_renyi.rs
ljeub-pometry: you need to actually return the error, this should be `add_node(...)?;`

raphtory/src/graphgen/erdos_renyi.rs
```diff
@@ -0,0 +1,119 @@
...
+            .ok();
+    }
+    for i in 0..nodes_to_add {
+        let source_id = GID::U64(i as u64);
+        for j in (i + 1)..nodes_to_add {
+            let dst_id = GID::U64(j as u64);
+            let create_edge = rng.gen_bool(p);  
+            if create_edge {
+                latest_time += 1;
+                graph.add_edge(latest_time, &source_id, &dst_id, NO_PROPS, None).expect("Not able to add edge");
```
ljeub-pometry: Same here, this should now be `add_edge(...)?;`

ljeub-pometry (changes requested):
You are basically there, just need to add some `?` to actually return the errors.

raphtory/src/graphgen/erdos_renyi.rs
```diff
@@ -0,0 +1,117 @@
...
+            .add_node(latest_time, &id, NO_PROPS, None)?;
+    }
+    for i in 0..nodes_to_add {
+        let source_id = GID::U64(i as u64);
+        for j in (i + 1)..nodes_to_add {
+            let dst_id = GID::U64(j as u64);
+            let create_edge = rng.gen_bool(p);  
+            if create_edge {
+                latest_time += 1;
+                graph.add_edge(latest_time, &source_id, &dst_id, NO_PROPS, None)?;
```
ljeub-pometry, line 69: I'm sorry I missed this before. As graphs in raphtory are currently always directed, this is actually only making half the graph. If you actually want this to be undirected, you need to explicitly add the reverse edge as well.

PR #2271, opened 2025-09-09
Fix debug symbols

fabianmurariu:
macro to automatically box some iterators

bring back debug symbols


db4-storage/src/lib.rs
```diff
@@ -160,3 +160,20 @@ pub fn collect_tree_paths(path: &Path) -> Vec<PathBuf> {
...
+        {
+            println!("Debug symbols are present.");
+        }
+
+        #[cfg(not(has_debug_symbols))]
+        {
+            println!("Debug symbols are not present.");
+        }
+    }
+}
```
ljeub-pometry: probably don't need this any more?

raphtory-api-macros/tests/user_examples_test.rs
```diff
@@ -0,0 +1,137 @@
+use raphtory_api_macros::box_on_debug_lifetime;
```
ljeub-pometry: these AI-generated tests aren't very good, most of the stuff isn't even used

raphtory/build.rs
```diff
@@ -3,10 +3,19 @@ use std::io::Result;
 fn main() -> Result<()> {
     prost_build::compile_protos(&["src/serialise/graph.proto"], &["src/serialise"])?;
     println!("cargo::rerun-if-changed=src/serialise/graph.proto");
+
+    println!("cargo::rustc-check-cfg=cfg(has_debug_symbols)");
+    if let Ok("true") = std::env::var("DEBUG").as_deref() {
```
ljeub-pometry: Have a look at the docs for the DEBUG flag, there are more options than true, you also want to turn this on if it is set to 2 I think.

ljeub-pometry (changes requested):
Tidy up the AI-generated tests and enable the cfg when `DEBUG=2`, the macro itself looks good.

Issue #2268, opened 2025-09-08
Community detection is not well documented or searchable


BaCk7:
If we search for community detection, we don't find label propagation and fast rp. We may want to update their docs.

<img width="1816" height="1532" alt="Image" src="https://github.com/user-attachments/assets/64f5af53-2116-46af-bafa-295458e8950d" />

PR #2276, opened 2025-09-11
community detection

jbaross-pometry:

Covers changes for ticket https://github.com/Pometry/Raphtory/issues/2268 and adds an introduction to community detection with a tutorial using our existing algorithms and UI. This should apear first in the search results.

Current search results were uninformative and a key usecase was missing from the userguide

Inlione code examples have basic tests.

There is a lot of room for updates here as the product evolves and if we want to cover a more elaborate use case (syntheticr was discussed in this context). 

### For reviewers

- Correctness of language in the introduction wrt to community norms. I wrote this based on wikipedia and some lit review papers I found.
- Do the descriptions for Louvain and Label propogation need more detail?

### DIFF
--- a/docs/reference/graphql/graphql_API.md
+++ b/docs/reference/graphql/graphql_API.md
@@ -2547,26 +2547,6 @@ This allows you to specify multiple operations together.
 </thead>
 <tbody>
 <tr>
-<td colspan="2" valign="top"><strong id="graphalgorithmplugin.pagerank">pagerank</strong></td>
-<td valign="top">[<a href="#pagerankoutput">PagerankOutput</a>!]!</td>
-<td></td>
-</tr>
-<tr>
-<td colspan="2" align="right" valign="top">iterCount</td>
-<td valign="top"><a href="#int">Int</a>!</td>
-<td></td>
-</tr>
-<tr>
-<td colspan="2" align="right" valign="top">threads</td>
-<td valign="top"><a href="#int">Int</a></td>
-<td></td>
-</tr>
-<tr>
-<td colspan="2" align="right" valign="top">tol</td>
-<td valign="top"><a href="#float">Float</a></td>
-<td></td>
-</tr>
-<tr>
 <td colspan="2" valign="top"><strong id="graphalgorithmplugin.shortest_path">shortest_path</strong></td>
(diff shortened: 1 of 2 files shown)


docs/user-guide/algorithms/5_community_detection.md
```diff
@@ -0,0 +1,144 @@
...
+```python
+from raphtory import Graph
+import pandas as pd
+import networkx as nx
+from raphtory import graphql
+from raphtory import algorithms as rp
+import matplotlib.pyplot as plt
+
+# Load the GML file using NetworkX
+karate = nx.read_gml("../data/karate.gml", label=None)
```
miratepuffin: Don't need this any more?

docs/user-guide/algorithms/5_community_detection.md
```diff
@@ -0,0 +1,144 @@
...
+You can explore the results of our cluster detection algorithm in greater detail using the Raphtory UI.
+
+To do this assign a type to nodes of each cluster and start the Raphtory server. Each unique node type will be assigned a colour in the **Graph canvas** so that you can distinguish them visually.
+
+/// tab | :fontawesome-brands-python: Python
+```{.python continuation}
+
+# Check value of cluster for each node and add to corresponding cluster list
+for node, cluster in clustering.items():
+    if cluster == 0:
```
miratepuffin: labels = {0:'Cobra Kai',1:'Miyagi-Do',2:'Polarslaget',3:'Redentores'}
G.node(node).set_node_type(labels[cluster])

miratepuffin (approved):

docs/user-guide/algorithms/5_community_detection.md
jbaross-pometry: we do because that's the simplist way to read the .gml file afaik

docs/user-guide/algorithms/5_community_detection.md
jbaross-pometry: doesn't seem to assign all the node types, if you run `G.get_all_node_types()` only Polarslaget is assigned

jbaross-pometry: ah no got it, inside the for loop

PR #2277, opened 2025-09-15
Rolling and expanding window alignment based on the user's time interval input

arienandalibi:
Added rolling_aligned() and expanding_aligned() functions which behave like rolling() and expanding() windows except that the start of each window is aligned with the smallest unit of time passed by the user. For example, if the interval is "1 month and 1 day", the first window will begin at the start of the day of the first time event. The non-aligned versions of these functions begin the window at the first time event. When used in Python, aligning is now the default and has to be disabled using align_start = False if desired.

It's more intuitive for windows to be aligned cleanly. If someone wants a rolling window over "1 year", they likely want it to start at the beginning of the calendar year and end at the end of the year.

The functions expanding() and rolling() now have a new option parameter called align_start, which allows users to disable the alignment (use the same functionality as before).

### DIFF
--- a/python/tests/test_base_install/test_graphdb/test_rolling_weird_alignment.py
+++ b/python/tests/test_base_install/test_graphdb/test_rolling_weird_alignment.py
@@ -3,6 +3,7 @@ import pytest
 from raphtory import Graph, GraphView
 from utils import run_group_graphql_test
 
+
 # previously, month boundaries could offset all following rolling windows.
 # eg. january 31st + "1 month" = february 28th (to not spill into next month)
 # these tests ensure the following windows aren't offset, eg. here on march 28th, april 28th, ...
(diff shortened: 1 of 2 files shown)


python/tests/test_base_install/test_graphql/test_rolling_expanding.py
```diff
@@ -2041,3 +2041,242 @@ def test_wrong_window():
     """
     queries_and_exceptions.append((query, mismatch_exception))
     run_group_graphql_error_test(queries_and_exceptions, graph)
+
+def test_alignment():
+    g = Graph()
```
miratepuffin, line 3203: These tests all look great! Can you just add a couple weird ones like "2 weeks" "3 days"  "2 weeks, 3 days" etc so we have some durations that non singular units 

Sorry meant to put this on the file above when you were doing the rolling/expanding in python tests

miratepuffin: The implementation all looks good for this, just so we have the tests available in case we change the implementation later

python/tests/test_base_install/test_graphql/test_rolling_expanding.py
```diff
@@ -2041,3 +2041,242 @@ def test_wrong_window():
...
+    dt1 = datetime(2025, 3, 15, 14, 37, 52)     # March 15
+    dt2 = datetime(2025, 7, 8, 9, 12, 5)        # July 8
+    dt3 = datetime(2025, 11, 22, 21, 45, 30)    # November 22
+
+    g.add_node(dt1, 1)
+    g.add_node(dt2, 1)
+    g.add_node(dt3, 1)
+
+    query = """
+    {
```
miratepuffin, line 3215: Can you just chuck in a rolling/expanding test for Node, nodes, edge, edges, pathFromNode, PathFromGraph. I know its a little extra, but just want to make sure we have them called at least once

miratepuffin (approved):

python/tests/test_base_install/test_graphdb/test_rolling_expanding_alignment.py
```diff
@@ -0,0 +1,348 @@
...
+
+    g.add_edge(datetime(2025, 3, 15, 15, 0, 0), 1, 2)
+    g.add_edge(datetime(2025, 3, 16, 1, 0, 0), 2, 3)
+
+    return g
+
+
+def test_rolling_month_alignment_default_true(example_graph):
+    g = example_graph
+    windows = list(islice(g.rolling("1 month"), 3))  # align_start defaults to True
```
ljeub-pometry: you really should also validate the last window (most likely to have bugs on the boundaries). Something like `windows = list(g.rolling("1 month"))`. Your existing test should still pass with that since you are indexing into windows anyway. Then also look at `windows[-1]` and make sure it looks correct.

ljeub-pometry: Same for all the other tests

python/tests/test_base_install/test_graphql/test_rolling_expanding.py
```diff
@@ -1871,173 +1953,1337 @@ def test_wrong_window():
...
     {
-  graph(path: "g") {
-    node(name: "1") {
-      rolling(window: {duration: "1 day"}, step: {epoch: 11}) {
-        list {
-          earliestTime
+      graph(path: "g") {
+        node(name: "1") {
+          rolling(window: {duration: "1 day"}, step: {epoch: 11}) {
+            page(limit: 3) {
```
ljeub-pometry, line 2016: again, would be good to also test the end of the rolling to make sure the boundaries are sensible

ljeub-pometry: you can use the other arguments to page to get the last window as a page (make sure to set limit>1 to actually test that it stops iterating correctly)

raphtory-core/src/utils/time.rs
```diff
@@ -205,16 +213,93 @@ impl From<Duration> for IntervalSize {
...
+    Hour,
+    Day,
+    Week,
+    Month,
+    Year,
+}
+
+impl AlignmentUnit {
+    /// Floors a UTC timestamp in milliseconds since Unix epoch to the nearest alignment unit.
+    pub fn align_timestamp(timestamp: i64, unit: AlignmentUnit) -> i64 {
```
ljeub-pometry: why does this not take `&self` instead of `unit: AlignmentUnit`?

raphtory-core/src/utils/time.rs
```diff
@@ -205,16 +213,93 @@ impl From<Duration> for IntervalSize {
...
+        let second_ms: i64 = 1000;
```
ljeub-pometry: these can be global constants?

raphtory-core/src/utils/time.rs
```diff
@@ -481,6 +566,30 @@ impl Add<Interval> for i64 {
     }
 }
 
+impl Add<Interval> for TimeIndexEntry {
+    type Output = TimeIndexEntry;
+    fn add(self, rhs: Interval) -> Self::Output {
+        match rhs.size {
+            IntervalSize::Discrete(number) => TimeIndexEntry::from(self.0 + (number as i64)),
```
ljeub-pometry: this one should also preserve the secondary index?

raphtory/src/db/api/view/time.rs
```diff
@@ -146,6 +146,18 @@ pub trait TimeOps<'graph>:
         I: TryInto<Interval>,
         ParseTimeError: From<<I as TryInto<Interval>>::Error>;
 
+    /// Creates a `WindowSet` with the given `step` size
+    /// using an expanding window. The last window may fall partially outside the range of the data/view.
+    /// The window will be aligned with the smallest unit of time passed. For example, if the interval
+    /// is "1 month and 1 day", the first window will begin at the start of the day of the first time event.
+    ///
+    /// An expanding window is a window that grows by `step` size at each iteration.
+    fn expanding_aligned<I>(&self, step: I) -> Result<WindowSet<'graph, Self>, ParseTimeError>
```
ljeub-pometry: I would keep the base methods as `rolling` and `expanding` which should do the automatic alignment. The `expanding_aligned` and `rolling_aligned` can take an extra argument that is `Option<AlignmentUnit>` so you can get custom alignment.

raphtory/src/db/api/view/time.rs
```diff
@@ -280,6 +330,35 @@ impl<'graph, V: OneHopFilter<'graph> + 'graph + InternalTimeOps<'graph>> TimeOps
...
+        window: I,
+        step: Option<I>,
+    ) -> Result<WindowSet<'graph, Self>, ParseTimeError>
+    where
+        Self: Sized + Clone + 'graph,
+        I: TryInto<Interval>,
+        ParseTimeError: From<<I as TryInto<Interval>>::Error>,
+    {
+        let parent = self.clone();
+        match (self.timeline_start(), self.timeline_end()) {
```
ljeub-pometry, line 367: If the graph is explicitly windowed, we probably don't want to do alignment if we rename this back to `rolling`. The easiest way to achieve that would be to pass the alignment unit as input to `timeline_start` and `timeline_end` which can handle that.

python/tests/test_base_install/test_graphql/test_rolling_expanding.py
arienandalibi: It looks like PathFromGraph isn't available in GraphQL, but it was tested in Python

PR #2284, opened 2025-09-22
impl qualifier filters for list properties and metadata

shivamka1:
Fixes https://github.com/Pometry/Raphtory/issues/2255

### DIFF
--- a/docs/reference/graphql/graphql_API.md
+++ b/docs/reference/graphql/graphql_API.md
@@ -2329,26 +2329,6 @@ This allows you to specify multiple operations together.
 </thead>
 <tbody>
 <tr>
-<td colspan="2" valign="top"><strong id="graphalgorithmplugin.pagerank">pagerank</strong></td>
-<td valign="top">[<a href="#pagerankoutput">PagerankOutput</a>!]!</td>
-<td></td>
-</tr>
-<tr>
-<td colspan="2" align="right" valign="top">iterCount</td>
-<td valign="top"><a href="#int">Int</a>!</td>
-<td></td>
-</tr>
-<tr>
-<td colspan="2" align="right" valign="top">threads</td>
-<td valign="top"><a href="#int">Int</a></td>
-<td></td>
-</tr>
-<tr>
-<td colspan="2" align="right" valign="top">tol</td>
-<td valign="top"><a href="#float">Float</a></td>
-<td></td>
-</tr>
-<tr>
 <td colspan="2" valign="top"><strong id="graphalgorithmplugin.shortest_path">shortest_path</strong></td>
 <td valign="top">[<a href="#shortestpathoutput">ShortestPathOutput</a>!]!</td>
 <td></td>
@@ -2366,6 +2346,26 @@ This allows you to specify multiple operations together.
 <tr>
 <td colspan="2" align="right" valign="top">direction</td>
 <td valign="top"><a href="#string">String</a></td>
+<td></td>
+</tr>
+<tr>
+<td colspan="2" valign="top"><strong id="graphalgorithmplugin.pagerank">pagerank</strong></td>
+<td valign="top">[<a href="#pagerankoutput">PagerankOutput</a>!]!</td>
+<td></td>
+</tr>
+<tr>
+<td colspan="2" align="right" valign="top">iterCount</td>
+<td valign="top"><a href="#int">Int</a>!</td>
+<td></td>
+</tr>
+<tr>
+<td colspan="2" align="right" valign="top">threads</td>
+<td valign="top"><a href="#int">Int</a></td>
+<td></td>
+</tr>
+<tr>
+<td colspan="2" align="right" valign="top">tol</td>
+<td valign="top"><a href="#float">Float</a></td>
 <td></td>
 </tr>
 </tbody>
--- a/raphtory-graphql/schema.graphql
+++ b/raphtory-graphql/schema.graphql
@@ -876,8 +876,8 @@ type Graph {
 }
 
 type GraphAlgorithmPlugin {
-	pagerank(iterCount: Int!, threads: Int, tol: Float): [PagerankOutput!]!
 	shortest_path(source: String!, targets: [String!]!, direction: String): [ShortestPathOutput!]!
+	pagerank(iterCount: Int!, threads: Int, tol: Float): [PagerankOutput!]!
 }
 
 type GraphSchema {


raphtory-graphql/src/model/graph/filtering.rs
```diff
@@ -407,6 +409,42 @@ pub enum EdgeFilter {
     Not(Wrapped<EdgeFilter>),
 }
 
+#[derive(Enum, Debug, Clone, Copy, PartialEq, Eq)]
+pub enum GqlListAgg {
```
ljeub-pometry, line 414: these should have `#[graphql(name="ListAgg")]` to remove the Gql on the actual graphql type

raphtory-graphql/src/model/graph/filtering.rs
```diff
@@ -407,6 +409,42 @@ pub enum EdgeFilter {
...
+    Len,
+    Sum,
+    Avg,
+    Min,
+    Max,
+}
+
+#[derive(Enum, Debug, Clone, Copy, PartialEq, Eq)]
+pub enum GqlListElemQualifier {
```
ljeub-pometry, line 424: this should have #[graphql(name="ListElemQualifier")] to remove the Gql on the actual graphql type

raphtory/src/db/graph/views/filter/model/property_filter.rs
```diff
@@ -502,8 +578,17 @@ impl<M> PropertyFilter<M> {
         match meta.get_prop_id_and_type(name, is_static) {
             None => Ok(None),
```
ljeub-pometry: We should make this case an error and return Result<usize, GraphError> (see #2285)

raphtory/src/db/graph/views/filter/model/property_filter.rs
```diff
@@ -1105,6 +1283,22 @@ impl<M: Send + Sync + Clone + 'static> InternalPropertyFilterOps
     }
 }
 
+impl<M> AnyTemporalPropertyFilterBuilder<M> {
+    pub fn any(self) -> AnyElemFilterBuilder<M> {
```
ljeub-pometry: I think you can make this a trait and blanket-implement it for anything that is InternalPropertyFilterOps?

raphtory/src/python/filter/property_filter_builders.rs
```diff
@@ -231,74 +243,165 @@ impl PyPropertyFilterOps {
...
+    fn all(&self) -> PyResult<PyPropertyFilterOps> {
+        Err(PyTypeError::new_err("Element qualifiers (any/all) cannot be used after a list aggregation (len/sum/avg/min/max)."))
+    }
+}
+
+#[derive(Clone)]
+struct NoListAggOps;
+
+impl DynListAggOps for NoListAggOps {
+    fn len(&self) -> PyResult<PyPropertyFilterOps> {
```
ljeub-pometry, line 310: We need to have a think about what happens with lists of lists and further nesting

raphtory/src/python/filter/property_filter_builders.rs
```diff
@@ -307,40 +410,69 @@ where
     M: Clone + Send + Sync + 'static,
     PropertyFilter<M>: CreateFilter + TryAsCompositeFilter,
 {
-    fn len(&self) -> PyPropertyFilterOps {
+    fn len(&self) -> PyResult<PyPropertyFilterOps> {
```
ljeub-pometry: you should make a trait for the aggregation methods on the rust side so you don't have to repeat this implementation for every builder

ljeub-pometry (changes requested):
We should have a look if we can implement this using some traits so we don't have quite so much repetition

python/tests/test_base_install/test_filters/semantics/test_edge_property_filter_semantics.py
```diff
@@ -226,9 +227,11 @@ def check(graph):
 def test_property_semantics_only_metadata2():
     def check(graph):
         filter_expr = filter.Edge.metadata("p1") == 1
-        result_ids = sorted(graph.filter(filter_expr).edges.id)
-        expected_ids = []
-        assert result_ids == expected_ids
+        with pytest.raises(
+            Exception,
+            match=r"Property p1 does not exist",
```
ljeub-pometry: The error message should probably say `Metadata` here and `Property` only if you did `filter.Edge.property("p1")`?

ljeub-pometry: Also note the issue in the description of the main `filter` PR such that it gets marked as resolved when that is merged.

raphtory/src/db/graph/views/filter/model/property_filter.rs
```diff
@@ -565,18 +565,14 @@ impl<M> PropertyFilter<M> {
...
-    ) -> Result<Option<usize>, GraphError> {
+    pub fn resolve_prop_id(&self, meta: &Meta, expect_map: bool) -> Result<usize, GraphError> {
         let (name, is_static) = match &self.prop_ref {
             PropertyRef::Metadata(n) => (n.as_str(), true),
             PropertyRef::Property(n) | PropertyRef::TemporalProperty(n, _) => (n.as_str(), false),
         };
 
         match meta.get_prop_id_and_type(name, is_static) {
-            None => Ok(None),
+            None => Err(GraphError::PropertyMissingError(name.to_string())),
```
ljeub-pometry: return different errors depending on the value of is_static

raphtory/src/db/graph/views/filter/exploded_edge_property_filter.rs
```diff
@@ -237,9 +237,18 @@ mod test_exploded_edge_property_filtered_graph {
     ) -> (PersistentGraph, PersistentGraph) {
         let g = PersistentGraph::new();
         let g_filtered = PersistentGraph::new();
-        if !edges.iter().all(|(_, v)| v.is_empty()) {
-            g_filtered.resolve_layer(None).unwrap();
-        }
```
ljeub-pometry, line 242: I am pretty sure this should still be here for the same reason that you added the property resolves

ljeub-pometry: all your resolves should be inside this if statement as these will only actually exist if there is an edge update (i.e., not on an empty graph)

raphtory/src/lib.rs
```diff
@@ -206,7 +206,7 @@ mod test_utils {
                 any::<String>(),
                 any::<i64>(),
             ),
-            0..=len,
+            1..=len,
```
ljeub-pometry: this does need to generate the empty graph

raphtory/src/lib.rs
```diff
@@ -607,10 +607,21 @@ mod test_utils {
     pub(crate) fn build_node_props(
         max_num_nodes: u64,
     ) -> impl Strategy<Value = Vec<(u64, Option<String>, Option<i64>)>> {
-        (0..max_num_nodes).prop_flat_map(|num_nodes| {
-            (0..num_nodes)
-                .map(|node| (Just(node), any::<Option<String>>(), any::<Option<i64>>()))
-                .collect_vec()
+        (1..=max_num_nodes).prop_flat_map(|num_nodes| {
```
ljeub-pometry: this does need to generate the empty graph

raphtory/src/db/graph/assertions.rs
```diff
@@ -440,3 +440,40 @@ pub fn search_edges(graph: &Graph, filter: impl TryAsCompositeFilter) -> Vec<Str
...
+
+pub fn assert_ok_or_missing_nodes<T>(
+    nodes: &[(u64, Option<String>, Option<i64>)],
+    res: Result<T, GraphError>,
+    on_ok: impl FnOnce(T),
+) {
+    match res {
+        Ok(v) => on_ok(v),
+        Err(GraphError::PropertyMissingError(name)) => {
+            let any_int_prop = nodes.iter().any(|(_, _, iv)| iv.is_some());
```
ljeub-pometry: this should match on `name` and check the correct property has no values accordingly (i.e., look at the strings if it is `str_prop` and look at the ints if it is `int_prop`.

PR #2287, opened 2025-09-24
add u8, u16, u32, f32, i32 to the graphql types

### LINKED ISSUE #2243: Graphql needs to support adding to raphtory graph
Graph that supports adding different rust data types like u8, etc

shivamka1:
Fixes  #2243


ljeub-pometry (approved):
can be merged but we should have a think about how to add dates and decimals

PR #2294, opened 2025-09-29
impl aggs for temporal properties, add tests

### LINKED ISSUE #2242: Aggs on the temporals
If the prop itself is not a list.
 - If it is numeric: temporal.values().avg(), sum(), min(), max(), median() (anything that works for numeric list should be supported here)
 - Always supported: len()

shivamka1:
Fixes #2242

### DIFF
--- a/raphtory/src/db/graph/views/filter/model/property_filter.rs
+++ b/raphtory/src/db/graph/views/filter/model/property_filter.rs
@@ -113,18 +113,6 @@ impl<M> Display for PropertyFilter<M> {
         let base = match &self.prop_ref {
             PropertyRef::Property(name) => name.to_string(),
             PropertyRef::Metadata(name) => format!("const({})", name),
-            PropertyRef::TemporalProperty(name, Temporal::Any) => {
-                format!("temporal_any({})", name)
-            }
-            PropertyRef::TemporalProperty(name, Temporal::Latest) => {
-                format!("temporal_latest({})", name)
-            }
-            PropertyRef::TemporalProperty(name, Temporal::First) => {
-                format!("temporal_first({})", name)
-            }
-            PropertyRef::TemporalProperty(name, Temporal::All) => {
-                format!("temporal_all({})", name)
-            }
             PropertyRef::TemporalProperty(name, t) => {
                 format!("temporal_{:?}({})", t, name)
             }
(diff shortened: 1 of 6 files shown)


raphtory/src/db/graph/views/filter/model/property_filter.rs
```diff
@@ -122,23 +125,35 @@ impl<M> Display for PropertyFilter<M> {
             PropertyRef::TemporalProperty(name, Temporal::All) => {
                 format!("temporal_all({})", name)
             }
+            PropertyRef::TemporalProperty(name, t) => {
```
ljeub-pometry, line 116: this covers the other `Temporal::` possibilities as well

raphtory/src/db/graph/views/filter/model/property_filter.rs
```diff
@@ -122,23 +125,35 @@ impl<M> Display for PropertyFilter<M> {
...
+                format!("temporal_{:?}({})", t, name)
+            }
+        };
+
+        let qualified = match self.list_elem_qualifier {
```
ljeub-pometry, line 121: Should `any` and `all` also be list aggregations instead of something special. I.e., if property "x" is e.g. `List(I64)`, the syntax would be something like

```python
(filter.Node.property("x") == 1).any()
```

instead of

```python
filter.Node.property("x").any() == 1
```

and `any` and `all` should be supported for other `List(Bool)` properties, e.g., 
```python
filter.Node.property("y").all()
```
where property "y" is of type `List(Bool)`.

raphtory/src/db/graph/views/filter/model/property_filter.rs
```diff
@@ -571,32 +593,67 @@ impl<M> PropertyFilter<M> {
...
+            match temporal {
+                Temporal::Values => {
+                    if self.list_agg.is_none() {
+                        return Err(GraphError::InvalidFilter(
+                            "temporal() must be followed by an aggregation (len/sum/avg/min/max)"
+                                .into(),
+                        ));
+                    }
+                    if self.list_elem_qualifier.is_some() {
+                        return Err(GraphError::InvalidFilter(
```
ljeub-pometry: this could be used to handle `Any` and `All` instead of them being special-cased as `Temporal::Any` and `Temporal::All`

raphtory/src/db/graph/views/filter/model/property_filter.rs
```diff
@@ -571,32 +593,67 @@ impl<M> PropertyFilter<M> {
...
+                Temporal::Latest | Temporal::First | Temporal::Any | Temporal::All => {
+                    if self.list_agg.is_some() {
+                        self.validate_agg(&dtype)?;
+                    } else if self.list_elem_qualifier.is_some() {
+                        self.validate_list_elem_and_operator(&dtype)?;
+                    } else {
+                        self.validate(&dtype, is_static && expect_map)?;
+                    }
+                    return Ok(id);
                 }
```
ljeub-pometry: this bit isn't needed, it is identical to the fallback code below

raphtory/src/db/graph/views/filter/model/property_filter.rs
ljeub-pometry: Probably for when we generalise the nested lists

raphtory/src/db/graph/views/filter/model/property_filter.rs
```diff
@@ -870,6 +941,22 @@ impl<M> PropertyFilter<M> {
                     has_any && all_ok
                 })
                 .is_some(),
+            PropertyRef::TemporalProperty(_, _) => {
```
ljeub-pometry: put `Temporal::Values` explicitly here instead of a catch-all

raphtory/src/python/filter/property_filter_builders.rs
```diff
@@ -438,24 +436,47 @@ where
...
+        self.agg.avg()
+    }
+
+    pub fn min(&self) -> PyResult<PyPropertyFilterOps> {
+        self.agg.min()
+    }
+
+    pub fn max(&self) -> PyResult<PyPropertyFilterOps> {
+        self.agg.max()
     }
```
ljeub-pometry, line 492: `__eq__` and `__ne__` would make sense here for checking for equal lists

PR #2298, opened 2025-09-30
bump rust version for release action

ricopinazo:

### DIFF
--- a/.github/workflows/release_bump_versions.yml
+++ b/.github/workflows/release_bump_versions.yml
@@ -50,8 +50,10 @@ jobs:
       - run: |
           git config --global user.email "ben.steer@pometry.com"
           git config --global user.name "Pometry-Team"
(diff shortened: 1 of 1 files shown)


.github/workflows/release_bump_versions.yml
```diff
@@ -50,8 +50,10 @@ jobs:
       - run: |
           git config --global user.email "ben.steer@pometry.com"
           git config --global user.name "Pometry-Team"
-      - uses: ./.github/actions/setup_rust
-        name: Setup Rust
+      - name: Setup rust
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: 1.87
```
ljeub-pometry: you can just set this to "toolchain: stable" so it stays up to date

ricopinazo: done

miratepuffin (approved):

PR #2313, opened 2025-10-03
semaphore-plus-final-fix-plus-rwlock-plus-loop

ricopinazo:
- **Fix deadlock (#2292)**
- **Release v0.16.2 (#2299)**
- **make all the rw_lock.read be read_recursive**
- **fixes deadlocks**
- **fixes Python deadlock**
- **add a sempahore in front of the graphql query execution**
- **sort pr and add env variable to be able to set concurrency limit**
- **add rwlock in the middleware**

### DIFF
--- a/raphtory-graphql/src/auth.rs
+++ b/raphtory-graphql/src/auth.rs
@@ -30,6 +30,7 @@ pub(crate) struct TokenClaims {
     pub(crate) a: Access,
 }
 
+// TODO: maybe this should be renamed as it doens't only take care of auth anymore
 pub struct AuthenticatedGraphQL<E> {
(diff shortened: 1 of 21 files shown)


raphtory-graphql/src/auth.rs
```diff
@@ -106,11 +163,43 @@ where
...
+                }
+            }
         }
     }
 }
 
+fn is_query_heavy(query: &str) -> bool {
+    query.contains("outComponent")
+        || query.contains("inComponent")
+        || query.contains("edges")
```
fabianmurariu, line 195: since when are these heavy?

PR #2316, opened 2025-10-06
Migrate polars to arrow-rs

ljeub-pometry:

migrate all interactions with arrow from polars-arrow to arrow-rs

easier integration with the wider ecosystem

Only in the very low-level apis

the tests


raphtory/src/io/arrow/node_col.rs
```diff
@@ -1,133 +1,217 @@
...
+    }
+    fn len(&self) -> usize {
+        ArrowArray::len(self)
+    }
 }
 
-impl NodeColOps for PrimitiveArray<i64> {
-    fn get(&self, i: usize) -> Option<GidRef> {
-        StaticArray::get(self, i).map(|v| GidRef::U64(v as u64))
+impl NodeColOps for arrow_array::StringArray {
```
fabianmurariu: fully qualified is a bit inconsistent we should import instead `use arrow_array::StringArray`

raphtory/src/io/arrow/node_col.rs
```diff
@@ -1,133 +1,217 @@
...
+    }
+    fn len(&self) -> usize {
+        ArrowArray::len(self)
     }
 }
 
-impl NodeColOps for PrimitiveArray<i32> {
-    fn get(&self, i: usize) -> Option<GidRef> {
-        StaticArray::get(self, i).map(|v| GidRef::U64(v as u64))
+impl NodeColOps for arrow_array::LargeStringArray {
```
fabianmurariu: `use arrow_array::LargeStringArray`

raphtory/src/io/arrow/node_col.rs
```diff
@@ -1,133 +1,217 @@
...
+
+    fn len(&self) -> usize {
+        ArrowArray::len(self)
     }
 }
 
-impl<O: Offset> NodeColOps for Utf8Array<O> {
-    fn get(&self, i: usize) -> Option<GidRef> {
-        if i >= self.len() {
+impl NodeColOps for arrow_array::StringViewArray {
```
fabianmurariu: same, use import

raphtory/src/io/arrow/node_col.rs
fabianmurariu, line 1: `use arrow_array::..` instead of using them fully qualified

raphtory-storage/Cargo.toml
```diff
@@ -22,14 +22,15 @@ serde = { workspace = true, features = ["derive"] }
 itertools = { workspace = true }
 thiserror = { workspace = true }
 pometry-storage = { workspace = true, optional = true }
-polars-arrow = { workspace = true, optional = true }
 bigdecimal = { workspace = true, optional = true }
 num-traits = { workspace = true, optional = true }
 parking_lot = { workspace = true }
+arrow = { workspace = true }
```
fabianmurariu: Would just importing arrow work? they re-export everything from what I recall

ljeub-pometry: actually only needed arrow-schema...

PR #2317, opened 2025-10-07
Stress tests

ricopinazo:

This just adds a stress test to run on every push and and a long set of stress tests to run nightly

- [x] put all utility functions used across bench.ts and stress-test in a common file
- [ ] still need to add queries that delete/add graphs
- [ ] still need to do things like including an update and a read on the same query
- [x] add edges into abc layers

### DIFF
--- a/docs/reference/graphql/graphql_API.md
+++ b/docs/reference/graphql/graphql_API.md
@@ -2417,26 +2417,6 @@ This allows you to specify multiple operations together.
 </thead>
 <tbody>
 <tr>
-<td colspan="2" valign="top"><strong id="graphalgorithmplugin.shortest_path">shortest_path</strong></td>
-<td valign="top">[<a href="#shortestpathoutput">ShortestPathOutput</a>!]!</td>
-<td></td>
-</tr>
-<tr>
-<td colspan="2" align="right" valign="top">source</td>
-<td valign="top"><a href="#string">String</a>!</td>
-<td></td>
-</tr>
-<tr>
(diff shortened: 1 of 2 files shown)


graphql-bench/Makefile
```diff
@@ -23,6 +25,50 @@ bench: deploy push-graphs build-raphtory
...
+
+stress-test-long: build
+	for _ in $(seq 1 12); do \
+		K6_WEB_DASHBOARD=true K6_WEB_DASHBOARD_EXPORT=stress-test-report.html k6 run --duration=2h dist/stress-test.js
+	done
+
+stress-until-error-with-docker-and-samply:
+	date
+	# cd .. && docker build -t raphtory:debug-base --build-arg RAPHTORY_PROFILE=release-with-debug .
+	# docker build -t raphtory:debug . -f debug.Dockerfile
```
miratepuffin: Do we want a function that builds these as well

graphql-bench/src/stress-test.ts
```diff
@@ -0,0 +1,572 @@
...
+  return {
+    numNodes: response.data.graph.countNodes as number,
+    numEdges: response.data.graph.countEdges as number
+  }
+}
+
+function deleteEdge() {
+  const { numEdges } = queryGraphSize("empty");
+
+  const edgeIndex = Math.floor(numEdges / 2 * Math.random()); // just in case, to avoid races, only target the first half
```
miratepuffin: Can we also delete some random edges?

PR #2330, opened 2025-10-13
Explicitly add filter to return types and misc filter stub fixes

jbaross-pometry:
- Explicitly add filter to return types
- Add names to argument docstrings
- usize -> int

### DIFF
--- a/python/python/raphtory/filter/__init__.pyi
+++ b/python/python/raphtory/filter/__init__.pyi
@@ -79,9 +79,9 @@ class PropertyFilterOps(object):
         Uses a specified Levenshtein distance and optional prefix matching.
 
         Arguments:
-            prop_value (str):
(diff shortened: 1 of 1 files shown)


raphtory/src/db/graph/views/filter/model/mod.rs
```diff
@@ -112,10 +117,11 @@ impl Filter {
...
     ///
     /// Arguments:
-    ///     prop_value (str):
-    ///     levenshtein_distance (usize):
+    ///     levenshtein_distance (int):
     ///     prefix_match (bool):
-    ///  
+    ///
+    /// Returns:
+    ///     PropValue (str):
```
miratepuffin, line 124: Does the field_name/value get inferred here?

PR #2340, opened 2025-10-15
make all the main write locks loopy

fabianmurariu:
Deadlocks
Deadlocks
Deadlocks
Deadlocks
Deadlocks


ricopinazo (commented):
can confirm the deadlock seems to be gone with these changes

PR #2341, opened 2025-10-15
James/ingestion options

jbaross-pometry:

- broadens discussion of ingestion to incclude parquet in addition to dataframes


docs/user-guide/ingestion/3_dataframes.md
```diff
@@ -136,6 +136,33 @@ assert str(g) == "Graph(number_of_nodes=5, number_of_edges=7, number_of_temporal
     Edge(source=ServerA, target=ServerB, earliest_time=1693555200000, latest_time=1693555200000, properties={data_size_MB: 5.6, datasource: {Critical System Request: docs/data/network_traffic_edges.csv}, is_encrypted: {Critical System Request: true}}, layer(s)=[Critical System Request])
     ```
 
+## Creating a graph from a Parquet file
+
+Similarly for Parquet you can use `from_parquet()`, or `load_edges_from_parquet()` and `load_nodes_from_parquet()` to load data from files in the common [Apache Parquet](https://parquet.apache.org/) format.
```
miratepuffin, line 141: Do we want to be explicit that this parquet is from calling to_parquet?

miratepuffin (approved):

docs/user-guide/ingestion/3_dataframes.md
jbaross-pometry: No we don't. 

We just want to show an import and not imply anything about it's origin since this should work with any parquet not just our outputs afaik.

jbaross-pometry: via slack - turns out parquets from `to_parquet` are special so it's better to show nodes + edges here.

miratepuffin (approved):

PR #2348, opened 2025-10-21
Fix master tests

miratepuffin:


stub_gen/build/lib/stub_gen.py
```diff
@@ -0,0 +1,488 @@
+import ast
```
miratepuffin: @jbaross-pometry any idea why this has popped up as new?

jbaross-pometry: Delete this. Stubs_gen/build should be empty.

PR #2350, opened 2025-10-21
Features/gql apis

shivamka1:
Fixes https://github.com/Pometry/Raphtory/issues/2353


raphtory-graphql/src/model/graph/nodes.rs
```diff
@@ -235,7 +224,7 @@ impl GqlNodes {
                 NodesViewCollection::ShrinkStart(time) => return_view.shrink_start(time).await,
                 NodesViewCollection::ShrinkEnd(time) => return_view.shrink_end(time).await,
                 NodesViewCollection::NodeFilter(node_filter) => {
-                    return_view.node_filter(node_filter).await?
```
ljeub-pometry, line 238: this seems weird, shouldn't it use `filter` to be consistent with the other cases?

ljeub-pometry: The `apply_views` function should just go away and be replaced by `filter` once the filter apis are feature-complete and cover all these cases!

raphtory-graphql/src/model/graph/nodes.rs
```diff
@@ -352,4 +341,26 @@ impl GqlNodes {
...
+        })
+        .await
+    }
+
+    /// Returns filtered list of nodes
+    async fn select(&self, expr: NodeFilter) -> Result<Self, GraphError> {
+        let self_clone = self.clone();
+        blocking_compute(move || {
+            let filter: CompositeNodeFilter = expr.try_into()?;
+            let filtered = self_clone.nn.filter_iter(filter)?;
```
ljeub-pometry: let's rename the rust method to `select` as well

raphtory-graphql/src/model/graph/graph.rs
```diff
@@ -502,10 +526,10 @@ impl GqlGraph {
         .await
     }
 
-    async fn node_filter(&self, filter: NodeFilter) -> Result<Self, GraphError> {
+    async fn filter_nodes(&self, expr: NodeFilter) -> Result<Self, GraphError> {
```
ljeub-pometry: the `filter_nodes` and `filter_edges` functions should be combined (same as in rust)

ljeub-pometry: make an issue for this (not part of this PR)

raphtory-graphql/src/model/graph/graph.rs
```diff
@@ -374,12 +374,28 @@ impl GqlGraph {
...
 
     /// Gets (optionally a subset of) the nodes in the graph.
-    async fn nodes(&self, ids: Option<Vec<String>>) -> GqlNodes {
-        let nodes = self.graph.nodes();
-        match ids {
-            None => GqlNodes::new(nodes),
-            Some(ids) => GqlNodes::new(blocking_compute(move || nodes.id_filter(ids)).await),
+    async fn nodes(
+        &self,
+        ids: Option<Vec<String>>,
```
ljeub-pometry: `ids` is redundant with the select (also we need to remember to fix the select filter for nodes (right now it applies a subgraph filter which can remove too many nodes if they were only added via edges, can you make sure there is a ticket for this?)

ljeub-pometry: add the optional `select` input argument to all methods that return an object that has `select` for consistency.

PR #2351, opened 2025-10-22
Update UI to v0.1.17

miratepuffin:
Filter out non valid edges in direct connections
Added Node type loading indicator to search page


louisch:
Need to make the action update the ui-tests as well

PR #2362, opened 2025-10-27
Adding node filtering on src and dst endpoints for edges

arienandalibi:
Adding all node filtering options on src and dst endpoints for edges. This includes property and metadata filtering, as well as operations that can be chained to them.

Edge src and dst filtering is currently limited to name and id, which is restrictive and inconsistent with the rest of the filtering system.

The same APIs available for node filtering will now be available for filter.Edge.src() and filter.Edge.dst()

Tests to be added

### DIFF
--- a/raphtory-graphql/src/model/graph/filtering.rs
+++ b/raphtory-graphql/src/model/graph/filtering.rs
@@ -8,7 +8,7 @@ use dynamic_graphql::{
 };
 use raphtory::{
     db::graph::views::filter::model::{
-        edge_filter::CompositeEdgeFilter,
+        edge_filter::{CompositeEdgeFilter, EdgeDstEndpoint, EdgeSrcEndpoint},
         filter_operator::FilterOperator,
         node_filter::CompositeNodeFilter,
         property_filter::{Op, PropertyFilter, PropertyFilterValue, PropertyRef},
@@ -421,6 +421,18 @@ pub enum EdgeFilter {
     Src(NodeFieldFilterNew),
     /// Destination node filter.
     Dst(NodeFieldFilterNew),
+    /// Source node property filter.
+    SrcProperty(PropertyFilterNew),
+    /// Destination node property filter.
+    DstProperty(PropertyFilterNew),
+    /// Source node metadata filter.
+    SrcMetadata(PropertyFilterNew),
+    /// Destination node metadata filter.
+    DstMetadata(PropertyFilterNew),
+    /// Source node temporal property filter.
+    SrcTemporalProperty(PropertyFilterNew),
+    /// Destination node temporal property filter.
+    DstTemporalProperty(PropertyFilterNew),
     /// Property filter.
     Property(PropertyFilterNew),
     /// Metadata filter.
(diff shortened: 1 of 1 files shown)


raphtory/src/db/graph/views/filter/model/edge_filter.rs
```diff
@@ -28,6 +30,8 @@ impl Display for EdgeFieldFilter {
 pub enum CompositeEdgeFilter {
     Edge(Filter),
     Property(PropertyFilter<EdgeFilter>),
+    SrcEndpointProperty(PropertyFilter<EdgeSrcEndpoint>),
+    DstEndpointProperty(PropertyFilter<EdgeDstEndpoint>),
```
shivamka1, line 34: It is not just property filtering that we intend to add to edge src and dst but anything and everything that a node filter supports. 

Probably we can utilise NodeFilter by wrapping around it? Something like:
```rust 
SrcNode(NodeFilter),
DstNode(NodeFilter),
```

This would allow to support any node filtering (current and future) out of the box without having to add new variants and wrappers.

This would also mean we can get rid of existing edge id and name endpoint builders.

shivamka1: Similarly we can introduce an enum variant `enum EdgeEndpoint { Src, Dst }`. Allows us to have a single builder.

raphtory/src/db/graph/views/filter/edge_endpoint_property_filtered_graph.rs
```diff
@@ -0,0 +1,239 @@
...
+    prop_id: usize,
+    filter: PropertyFilter<EdgeSrcEndpoint>,
+}
+
+#[derive(Debug, Clone)]
+pub struct EdgeDstEndpointPropertyFilteredGraph<G> {
+    graph: G,
+    prop_id: usize,
+    filter: PropertyFilter<EdgeDstEndpoint>,
+}
```
shivamka1, line 39: Since we can utilise NodeFilters, I don't think we would need these specialised edge src/dst endpoint property filtered graphs either.

arienandalibi: I was having trouble understanding all the types at play here so correct me if I'm wrong, but it seems like the M in PropertyFilter\<M\> is simply a marker trait, and that is the only thing differentiating the different types of filtered graphs. 

Also correct me if I'm wrong but, if that's the case, aren't these filtered graphs different from NodePropertyFilteredGraph in that there is a custom implementation of InternalEdgeFilterOps (to select the right edge endpoint) instead of the node filtered graph's custom implementation of InternalNodeFilterOps? I thought this graph filtered edges while the node one filtered nodes.

I'm not too sure how to get rid of these filtered graph wrappers and use the node ones.

raphtory/src/db/graph/views/filter/model/edge_filter.rs
arienandalibi: I'm having trouble seeing how to easily use SrcNode/DstNode(NodeFilter) to add NodeFilter operations to filter edges based on endpoints. Even in the node filtering functions, I only see PropertyFilter<NodeFilter>, where we call i.create_filter() to return a NodePropertyFilteredGraph (nodes are filtered, not edges). Might be worth going over this in case I'm missing something.

shivamka1: Resolved in https://github.com/Pometry/Raphtory/pull/2364

raphtory/src/db/graph/views/filter/edge_endpoint_property_filtered_graph.rs
shivamka1: Resolved in https://github.com/Pometry/Raphtory/pull/2364

PR #2363, opened 2025-10-28
James/vectors

jbaross-pometry:

- Updated docstrings for raphtory/src/python/packages/vectors.rs and raphtory/src/python/types/wrappers/document.rs for clarity and grammar.
- Added a userguide page to describe the vertor tools at a conceptual level
- Added a simple example of retriving results and feeding into an LLM
    - This example isn't great becuse I didn't have rich data to demo the templates with or ask more interesting queries.
    - Template strings also seem to break mkdocs when included in code fences

---

It looks like some things have been renamed or reworked in the https://github.com/Pometry/Raphtory/pull/2249/files PR so any feedback to align these would be helpful

### DIFF
--- a/docs/user-guide/algorithms/6_vectorisation.md
+++ b/docs/user-guide/algorithms/6_vectorisation.md
@@ -22,9 +22,9 @@ When you call [Vectorise()][raphtory.GraphView.vectorise] Raphtory automatically
 
 ### Document templates
 
-The templates for entity documents follow a subset of [Jinja](https://jinja.palletsprojects.com/en/stable/templates/) and graph attributes and properties are exposed so that you can use them in template expressions.
+The templates for entity documents follow a subset of [Jinja](https://jinja.palletsprojects.com/en/stable/templates/) using [Mini Jinja](https://docs.rs/minijinja/latest/minijinja/). 
 
-Most attributes of graph entities are exposed and can be used in Jinja expressions. The nesting of attributes reflects the Python interface and the final result of any chain such as `properties.prop_name` or `src.name` should be a standard python type. For datetime values, Raphtory converts them into milliseconds since the Unix epoch by default and provides an optional `datetimeformat` function to convert this to a human readable format.
+Additionally, graph attributes and properties are exposed so that you can use them in template expressions. The nesting of attributes reflects the Python interface and you can perform chains such as `properties.prop_name` or `src.name` which will follow the same typing as in Python. For `datetime` values, by default Raphtory converts these into milliseconds since the Unix epoch but provides an optional `datetimeformat` function to convert this to a human readable format.
 
 ## Retrieve documents
 


docs/user-guide/algorithms/6_vectorisation.md
```diff
@@ -0,0 +1,142 @@
...
+```
+///
+
+When you call [Vectorise()][raphtory.GraphView.vectorise] Raphtory automatically creates documents for each node and edge entity in your graph, optionally you can provide template strings to format documents and pass these to `vectorise()`. This is useful when you know which properties are semantically relevant or want to present information in a specific format when retrieved by a human or machine user. Additionally, you can cache the embedded graph to disk to avoid having to recompute the vectors when nothing has changed.
+
+### Document templates
+
+The templates for entity documents follow a subset of [Jinja](https://jinja.palletsprojects.com/en/stable/templates/) and graph attributes and properties are exposed so that you can use them in template expressions.
+
+Most attributes of graph entities are exposed and can be used in Jinja expressions. The nesting of attributes reflects the Python interface and the final result of any chain such as `properties.prop_name` or `src.name` should be a string.
```
ricopinazo: this is not entirely true, for properties the types are what you would expect, they can be strings, numbers, booleans, lists, and objects. For datetime values, we convert them into milliseconds since epoch, but we do provide a function to format those 

```
propreties.some_time_property|datetimeformat
```

raphtory/src/python/types/wrappers/document.rs
```diff
@@ -5,13 +5,13 @@ use crate::{
 };
 use pyo3::{prelude::*, IntoPyObjectExt};
 
-/// A Document
+/// A document corresponding to a graph entity. Used to generate embeddings.
 ///
 /// Args:
-///     content (str): the document content
-///     life (int | Tuple[int, int], optional): the optional lifespan for the document (single value
+///     content (str): The document content.
```
ricopinazo: I think Document objects can't be created anymore (and there are no plans for it in the future), so the Args section here can be simply removed instead

raphtory/src/python/packages/vectors.rs
```diff
@@ -261,15 +262,15 @@ impl PyVectorisedGraph {
             .nodes_by_similarity(&embedding, limit, translate_window(window))?)
     }
 
-    /// Search the top scoring edges according to `query` with no more than `limit` edges
+    /// Search the top similarity scoring edges according to matching a specified `query` with no more than `limit` edges in the result.
```
ricopinazo: I've seen these phrasing in multiple places and it sounds weird to me, but might be just because of my non-native brain. I see why you are changing these though, the old verions doesn't explain very well what's happening. I wonder if we could write something along the lines of: 
Retrieves up to `limit` edges whose associated text documents are most similar to the given `query`, ranked by their similarity scores in descending order.
Or something like:
Searches for edges whose associated text content best matches the given `query`. Returns up to `limit` results, ranked in descending order of similarity score.

ricopinazo (changes requested):
Everything looks good overall, I like the vectorisation page, just left a few comments that might be worth addressing now since they refer to things that are not going to change in a foreseeable future

ricopinazo (approved):
LGTM

PR #2368, opened 2025-11-04
Added edge node filtering options in Python and GraphQL

### LINKED ISSUE #1984: Add filter support for edge src and dst filters
Edge.src.property("p1") == 2
Edge.dst.property("p1").temporal().latest() > 2
Edge.src().id()
Edge.src().degree(), etc

filter.Edge.node().degree > 5 ----- filter.Edge.src().degree > 5 | filter.Edge.dst().degree > 5 (maybe, AND or OR) ??

All filter expressions supported for nodes should be supported for edge src and dst

arienandalibi:
Fixes #1984

### DIFF
--- a/raphtory/src/python/filter/node_filter_builders.rs
+++ b/raphtory/src/python/filter/node_filter_builders.rs
@@ -7,7 +7,7 @@ use crate::{
             NodeNameFilterBuilder, NodeTypeFilterBuilder,
         },
         property_filter::{MetadataFilterBuilder, PropertyFilterBuilder},
-        PropertyFilterFactory, Windowed,
+        PropertyFilterFactory, TryAsCompositeFilter, Windowed,
     },
     python::{
         filter::{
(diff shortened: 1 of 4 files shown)


raphtory/src/python/filter/node_filter_builders.rs
```diff
@@ -270,3 +296,260 @@ where
         )))
     }
 }
+
+impl DynNodeFilterBuilderOps for EndpointWrapper<NodeNameFilterBuilder> {
```
ljeub-pometry: I would not expect this trait to be implemented directly, should be via the blanket implementation above, something is weird here.

raphtory/src/python/filter/window_filter.rs
```diff
@@ -16,17 +17,46 @@ pub fn py_into_millis(obj: &Bound<PyAny>) -> PyResult<i64> {
...
-        let wpr: WindowedPropertyRef<NodeFilter> = self.0.clone().property(name);
+        self.0.property(name)
+    }
+
+    fn metadata(&self, name: String) -> PyPropertyFilterOps {
+        self.0.metadata(name)
+    }
+}
+
+pub(crate) trait DynWindowedNodeFilter: Send + Sync {
```
ljeub-pometry: We need to tidy this up a bit and change the name. This has nothing to do with window as such but just implements `property` and `metadata`.

ljeub-pometry: It should be possible to blanket-implement this for anything that implements `PropertyFilterFactory`

raphtory/src/search/edge_filter_executor.rs
```diff
@@ -242,12 +218,13 @@ impl<'a> EdgeFilterExecutor<'a> {
         offset: usize,
     ) -> Result<Vec<EdgeView<G>>, GraphError> {
         match filter {
+            CompositeEdgeFilter::Src(_) | CompositeEdgeFilter::Dst(_) => {
+                // TODO: Can we use an index here to speed up search?
+                fallback_filter_edges(graph, filter, limit, offset)
```
ljeub-pometry: We should be able to find the nodes first and then get their out/in-edges as needed. This can use an index and is also much more efficient even when indexes are not available.

```
graph.nodes[filter].out_edges.explode()
```
versus 
```
graph.edges.explode()[filter.ExplodedEdge.src(filter)]
```

Issue #2001, opened 2025-04-09
Exception: Utf8 not supported for time column


CiaranRobinson:
Data loaded in pandas from csv with `engine='c'` like so `df = pd.read_csv(filepath, usecols=cols,  engine='c')`
Result errors with the following exception:
```
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)

 g2 = rp.Graph()
 g2.load_edges_from_pandas(
     df=pandas_df,
     src='StartID',
     dst='EndID',
     time='DepartureDateTime',
     properties=['SomeValue']
     )

Exception: Utf8 not supported for time column
Caused by:
  -> Utf8 not supported for time column
```

Set up is:
Windows 11 
OS Version - 10.0.26100 Build 26100
Python - 3.12.9
Pandas - 2.2.3
raphtory - 0.14.0

Loads fine with python /  standard engine / pyarrow and string dtype. crashes with any string dtype if loaded with a c engine. :shrug:

PR #2370, opened 2025-11-06
Add support for date32 types in Docbrownv4, previously used to crash

arienandalibi:
Fixes:
Exception: Failed to load graph: Failed to load graph LoadError { source: InvalidPropertyType(Date32) }

Includes some tests


ljeub-pometry:
This should fix #2001 and should be based of master instead

PR #2382, opened 2025-11-11
Add support for date32, utf8, and different Timestamp(TimeUnit) types when loading from pandas or parquet

### LINKED ISSUE #2001: Exception: Utf8 not supported for time column

arienandalibi:
* Fixes:
Exception: Failed to load graph: Failed to load graph LoadError { source: InvalidPropertyType(Date32) }

* Fixes #2001

### DIFF
--- a/.github/workflows/_release_python.yml
+++ b/.github/workflows/_release_python.yml
@@ -44,18 +44,17 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: |
-            3.9
-            3.10
             3.11
             3.12
             3.13
+            3.14
       - name: Build wheels
         uses: PyO3/maturin-action@v1
         with:
           working-directory: ./python
           command: build
           target: ${{ matrix.target }}
-          args: --release --out dist -i python3.9 -i python3.10 -i python3.11 -i python3.12 -i python3.13
+          args: --release --out dist -i python3.11 -i python3.12 -i python3.13 -i python3.14
           manylinux: 2_28
           before-script-linux: |
             if [[ -f /etc/os-release ]]; then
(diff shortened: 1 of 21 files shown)


python/tests/test_base_install/test_loaders/test_load_from_pandas.py
```diff
@@ -1375,3 +1375,335 @@ def test_node_both_option_failures_pandas():
...
+    return datetime.datetime(dt.year, dt.month, dt.day, tzinfo=datetime.timezone.utc)
+
+
+def test_load_edges_from_pandas_time_date32_arrowdtype():
+    try:
+        import pyarrow as pa
+
+        arrow_dtype = pd.ArrowDtype(pa.date32())
+    except Exception:
+        pytest.skip("pandas ArrowDtype(pa.date32()) not available")
```
miratepuffin: Does this happen on some specific OS?

PR #2390, opened 2025-11-20
Implement graph properties for v4

fabubaker:


db4-storage/src/pages/graph_prop_page/writer.rs
```diff
@@ -0,0 +1,58 @@
...
+    pub fn update_metadata(&mut self, props: impl IntoIterator<Item = (usize, Prop)>, lsn: u64) {
+        let add = self.mem_segment.update_metadata(props);
+        self.mem_segment.layers_mut()[MemGraphPropSegment::DEFAULT_LAYER].set_lsn(lsn);
+
+        self.graph_props.increment_est_size(add);
+        self.graph_props.mark_dirty();
+    }
+}
+
+impl<GS: GraphPropSegmentOps> Drop for GraphPropWriter<'_, GS> {
```
ljeub-pometry, line 53: We might need to make this an actual method so we can report errors properly instead of panicking (panicking in drop is very ugly).

db4-storage/src/pages/locked/graph_props.rs
```diff
@@ -0,0 +1,79 @@
...
+        let add = self.lock.update_metadata(props);
+        self.lock.layers_mut()[MemGraphPropSegment::DEFAULT_LAYER].set_lsn(lsn);
+
+        self.page.increment_est_size(add);
+        self.page.mark_dirty();
+    }
+}
+
+impl<GS: GraphPropSegmentOps> Drop for LockedGraphPropPage<'_, GS> {
+    fn drop(&mut self) {
```
ljeub-pometry, line 52: I've made issue #2411 (we need to get rid of these drop implementations)

db4-storage/src/segments/graph_prop/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+            _persistent: ext,
+        }
+    }
+
+    fn load(
+        _meta: Arc<Meta>,
+        _path: impl AsRef<Path>,
+        _ext: Self::Extension,
+    ) -> Result<Self, StorageError> {
+        todo!()
```
ljeub-pometry: shouldn't this just return an error (I am assuming this is never going to get implemented?)

db4-storage/src/segments/graph_prop/segment.rs
```diff
@@ -0,0 +1,137 @@
...
+        layer.reserve_local_row(Self::DEFAULT_ROW.into());
+        let mut prop_mut_entry = layer.properties_mut().get_mut_entry(Self::DEFAULT_ROW);
+        prop_mut_entry.append_t_props(ts, props);
+
+        let layer_est_size = layer.est_size();
+        layer_est_size - est_size
+    }
+
+    pub fn add_metadata(&mut self, props: impl IntoIterator<Item = (usize, Prop)>) {
+        self.update_metadata(props);
```
ljeub-pometry: why do both add and update exist if we don't have the check here that the properties are new

raphtory/src/db/api/view/graph.rs
```diff
@@ -249,12 +249,25 @@ impl<'graph, G: GraphView + 'graph> GraphViewOps<'graph> for G {
...
         node_meta.set_temporal_prop_mapper(self.node_meta().temporal_prop_mapper().deep_clone());
         edge_meta.set_metadata_mapper(self.edge_meta().metadata_mapper().deep_clone());
         edge_meta.set_temporal_prop_mapper(self.edge_meta().temporal_prop_mapper().deep_clone());
-        let layer_meta = edge_meta.layer_meta();
+        graph_props_meta
+            .set_metadata_mapper(self.graph_props_meta().metadata_mapper().deep_clone());
+        graph_props_meta
+            .set_temporal_prop_mapper(self.graph_props_meta().temporal_prop_mapper().deep_clone());
+
+        let temporal_graph = TemporalGraph::new_with_meta(
```
ljeub-pometry: we are making the storage too early which will break on empty graphs

raphtory-api/src/core/entities/properties/meta.rs
```diff
@@ -104,6 +104,20 @@ impl Meta {
         }
     }
 
+    pub fn new_for_graph_props() -> Self {
+        let meta_layer = DictMapper::new_layer_mapper();
+        let meta_node_type = DictMapper::default();
+        meta_node_type.get_or_create_id("_default");
```
ljeub-pometry: no reason to add this?

db4-storage/src/pages/graph_prop_page/writer.rs
fabubaker: I'm fully aligned, I'm not a huge fan of having this in drop. In addition to the panicking, we should be very explicit about where we call `persist` instead of running it implicitly in drop.

db4-storage/src/segments/graph_prop/mod.rs
fabubaker: I assume so too. Patched things to return an error, also did the same for Node/EdgeSegmentView.

db4-storage/src/segments/graph_prop/segment.rs
fabubaker: I'm not sure either. I ported this over directly from `PropertyAdditionOps` which has both `add_metadata` and `update_metadata`, but they are literally the same lines of code. 

If `update_metadata` is supposed to check that the prop first exists I can make that change.

raphtory-api/src/core/entities/properties/meta.rs
fabubaker: Removed, I suspect we don't need it for `new_for_edges` either.

db4-storage/src/segments/graph_prop/segment.rs
fabubaker: As discussed, changed `add_metadata` to check if key exists and throw an error if it does.

raphtory/src/db/api/view/graph.rs
fabubaker: Fixed.

PR #2392, opened 2025-11-21
call blocking_compute in health check

ricopinazo:
Make the GraphQL health check go through the read and write rayon pools before returning, so deadlocks don't go undetected when using it

### DIFF
--- a/raphtory-graphql/src/rayon.rs
+++ b/raphtory-graphql/src/rayon.rs
@@ -31,14 +31,14 @@ pub async fn blocking_write<R: Send + 'static, F: FnOnce() -> R + Send + 'static
 #[cfg(test)]
 mod deadlock_tests {
     use std::{
-        path::PathBuf,
         sync::{Arc, Mutex},
         time::Duration,
     };
 
-    use reqwest::Client;
+    use reqwest::{Client, StatusCode};
+    use tempfile::TempDir;
 
-    use crate::{rayon::WRITE_POOL, GraphServer};
+    use crate::{rayon::WRITE_POOL, routes::Health, GraphServer};
 
     #[tokio::test]
     async fn test_deadlock_in_read_pool() {
@@ -61,7 +61,8 @@ mod deadlock_tests {
     }
 
     async fn test_pool_lock(port: u16, pool_lock: impl FnOnce(Arc<Mutex<()>>)) {
-        let server = GraphServer::new(PathBuf::from("/tmp/health-check-test"), None, None).unwrap();
+        let tempdir = TempDir::new().unwrap();
+        let server = GraphServer::new(tempdir.path().to_path_buf(), None, None).unwrap();
         let _running = server.start_with_port(port).await.unwrap();
         tokio::time::sleep(Duration::from_secs(1)).await; // this is to wait for the server to be up
         let lock = Arc::new(Mutex::new(()));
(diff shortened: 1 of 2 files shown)


raphtory-graphql/src/rayon.rs
```diff
@@ -27,3 +27,56 @@ pub async fn blocking_write<R: Send + 'static, F: FnOnce() -> R + Send + 'static
...
+
+    #[tokio::test]
+    async fn test_deadlock_in_read_pool() {
+        let port = 43871;
+        let _running = start_server(port).await;
+        for _ in 0..rayon::current_num_threads() {
+            rayon::spawn(move || loop {
+                std::hint::spin_loop();
+            });
+        }
```
ljeub-pometry: have a look at `spawn_broadcast` (https://docs.rs/rayon/latest/rayon/fn.spawn_broadcast.html) to reliably run a task on all threads (no guarantees with this code that one thread doesn't end up with more than one task and the other remains idle, though it should probably then steal a task)

raphtory-graphql/src/routes.rs
```diff
@@ -21,8 +23,10 @@ struct Version {
 
 #[handler]
 pub(crate) async fn health() -> impl IntoResponse {
-    let health = Health { healthy: true };
-    (StatusCode::OK, Json(health))
+    // using blocking_compute and blocking_write to identify deadlocks on any of the two rayon pools
+    blocking_compute(|| {}).await;
+    blocking_write(|| {}).await;
+    (StatusCode::OK, Json(Health { healthy: true }));
```
ljeub-pometry: Maybe we should have both a quick check (which just does what we had before) and this one? So you can tell if there is a network problem or an execution problem.

raphtory-graphql/src/rayon.rs
ljeub-pometry: also instead of busy-looping you can push a locked lock into the task
(something like this, not tested)
 
```
let lock = Arc::new(Mutex::new(()));
let guard = lock.lock();

let lock_clone = lock.clone();
rayon::spawn_broadcast(move || lock_clone().lock())
```

ricopinazo: Both changes worked, amazing, thanks!

raphtory-graphql/src/routes.rs
ricopinazo: I think some users might use the old check thinking that's enough and might not understand the difference. I'd prefer keeping it simple and provide only one way if that sounds good

raphtory-graphql/src/rayon.rs
```diff
@@ -27,3 +27,56 @@ pub async fn blocking_write<R: Send + 'static, F: FnOnce() -> R + Send + 'static
...
+        test_pool_lock(43872, |lock| {
+            WRITE_POOL.spawn_broadcast(move |_| {
+                let _guard = lock.lock().unwrap();
+            });
+        })
+        .await;
+    }
+
+    async fn test_pool_lock(port: u16, pool_lock: impl FnOnce(Arc<Mutex<()>>)) {
+        let server = GraphServer::new(PathBuf::from("/tmp/health-check-test"), None, None).unwrap();
```
ljeub-pometry: use TempDir::new() instead to avoid issues on windows though it seems you are getting away with it because you never actually try to write to the path?

raphtory-graphql/src/routes.rs
ljeub-pometry: You might be right, @miratepuffin ?

miratepuffin: Yeah I think one check is best really - 100% someone will get the wrong end of the stick and we get an angry message from someone :D

PR #2400, opened 2025-11-28
Added new tracing levels

miratepuffin:
Full tracing for complex queries generates quite large spans so we have added some different tracing levels. Available options are:
* COMPLETE: Provides full traces for each query.
* ESSENTIAL: Tracks key functions — addEdge, addEdges, deleteEdge, graph, updateGraph, addNode, node, nodes, edge, edges.
* MINIMAL: Provides only summary execution times.

Fixed issue where the server would only fail after timeout completes.

All fully documented 

Added into config tests, but need more CLI tests


raphtory-graphql/src/config/otlp_config.rs
```diff
@@ -9,13 +9,19 @@ use serde::Deserialize;
...
+pub const DEFAULT_TRACING_LEVEL: &'static str = "Complete";
+
 pub const DEFAULT_OTLP_AGENT_HOST: &'static str = "http://localhost";
 pub const DEFAULT_OTLP_AGENT_PORT: &'static str = "4317";
 pub const DEFAULT_OTLP_TRACING_SERVICE_NAME: &'static str = "Raphtory";
 
 #[derive(Clone, Deserialize, Debug, PartialEq, serde::Serialize)]
 pub struct TracingConfig {
     pub tracing_enabled: bool,
+    pub tracing_level: String,
```
ricopinazo: I think this could be an enum that gets deserialised using serde_json, but we can do that change later if the number of options increases in a way that cna be transparent to the user

PR #2427, opened 2025-12-18
Make graphql schema output stable

ljeub-pometry:

Register all plugins in a deterministic order such that the generated schema output remains stable

The old mechanism was using `HashMap` which generates a different order every time, causing the docs to change.

no

the tests


miratepuffin (approved):
Looks good to me

PR #2441, opened 2026-01-15
impl windowing filter

shivamka1:
fixes https://github.com/Pometry/Raphtory/issues/2403

### DIFF
--- a/python/python/raphtory/plottingutils.py
+++ b/python/python/raphtory/plottingutils.py
@@ -5,7 +5,6 @@ Useful code snippets for making commonly used plots in Raphtory.
 from pathlib import Path
 import numpy as np
 
-
 motif_im_dir = Path(__file__).parents[1].absolute().as_posix() + "/motif-images/"
 
 
--- a/python/tests/test_base_install/test_graphdb/test_node_state.py
+++ b/python/tests/test_base_install/test_graphdb/test_node_state.py
@@ -49,6 +49,6 @@ def test_group_by():
     assert len(groups_from_lazy) == len(expected)
 
     for i, (v, nodes) in enumerate(groups_from_lazy):
-        (v2, nodes2) = groups_from_lazy[i]
+        v2, nodes2 = groups_from_lazy[i]
         assert v == v2
         assert nodes.id == nodes2.id
--- a/python/tests/test_base_install/test_graphql/misc/test_snapshot.py
+++ b/python/tests/test_base_install/test_graphql/misc/test_snapshot.py
@@ -11,8 +11,7 @@ def test_snapshot():
         client = RaphtoryClient("http://localhost:1736")
 
         def query(graph: str, window: str):
-            return client.query(
-                f"""{{
+            return client.query(f"""{{
                 graph(path: "{graph}") {{
                     window: {window} {{
                         edges {{
@@ -27,8 +26,7 @@ def test_snapshot():
                         }}
                     }}
                 }}
-            }}"""
-            )
+            }}""")
 
         client.new_graph("event", "EVENT")
         g = client.remote_graph("event")


python/tests/test_base_install/test_graphql/test_filters/test_graph_edges_property_filter.py
```diff
@@ -743,8 +747,7 @@ def test_edge_temporal_property_filter_empty_layers_is_error(graph):
     run_graphql_error_test_contains(
         query,
         [
-            "EdgeFilter.temporalProperty",
-            "'layers' must be non-empty",
+            "EdgeFilter.layers.names must be non-empty",
```
ljeub-pometry: Do we actually want this error? In rust and python this is allowed.

raphtory/src/db/graph/views/filter/model/mod.rs
```diff
@@ -196,3 +400,115 @@ impl<T: TryAsCompositeFilter + ?Sized> TryAsCompositeFilter for Arc<T> {
...
+    ) -> Arc<dyn DynInternalViewWrapOps> {
+        Arc::new(self.clone().build_window(start, end))
+    }
+}
+
+impl InternalViewWrapOps for Arc<dyn DynInternalViewWrapOps> {
+    type Window = Arc<dyn DynInternalViewWrapOps>;
+
+    fn build_window(self, start: EventTime, end: EventTime) -> Self::Window {
+        self.dyn_build_window(start, end)
```
ljeub-pometry: this seems to be missing a `deref()`

raphtory/src/db/graph/views/filter/model/windowed_filter.rs
```diff
@@ -66,6 +65,18 @@ impl<M> Windowed<M> {
...
 
+impl<T: InternalViewWrapOps> InternalViewWrapOps for Windowed<T> {
+    type Window = T::Window;
+
+    fn bounds(&self) -> (EventTime, EventTime) {
+        (self.start, self.end)
+    }
+
+    fn build_window(self, start: EventTime, end: EventTime) -> Self::Window {
+        self.inner.build_window(start, end)
```
ljeub-pometry, line 76: we probably should move the `min`/`max` logic from the `window` method here instead

raphtory/src/db/graph/nodes.rs
```diff
@@ -185,7 +184,7 @@ where
         })
     }
 
-    fn iter_vids(&self, g: GraphStorage) -> impl Iterator<Item = VID> + Send + Sync + 'graph {
+    fn iter_vids(&self) -> impl Iterator<Item = VID> + Send + Sync + 'graph {
         let g = self.base_graph.core_graph().clone();
```
ljeub-pometry: This is broken, the GraphStorage input was used to implement locked and unlocked iterators. The correct fix is to delete the line above and use the input `g`!

raphtory/src/db/graph/nodes.rs
```diff
@@ -197,16 +196,17 @@ where
 
     #[inline]
     pub(crate) fn iter_refs_unlocked(&self) -> impl Iterator<Item = VID> + Send + Sync + 'graph {
-        let g = self.graph.core_graph().clone();
-        self.iter_vids(g)
+        self.iter_vids()
```
ljeub-pometry: revert this change (see above)!

raphtory/src/db/graph/nodes.rs
ljeub-pometry: Also, `iter_refs` above should call this method with a locked `GraphStorage` instead of re-implementing the filtering!

raphtory/src/python/filter/property_filter_builders.rs
```diff
@@ -6,207 +6,30 @@ use crate::{
...
-pub trait DynPropertyExprOps: DynPropertyExprBuilderOps + DynPropertyFilterOps {}
-
-impl<T: DynPropertyExprBuilderOps + DynPropertyFilterOps + ?Sized> DynPropertyExprOps for T {}
-
 #[pyclass(frozen, name = "FilterOps", module = "raphtory.filter", subclass)]
 #[derive(Clone)]
-pub struct PyPropertyExprBuilder {
-    inner: Arc<dyn DynPropertyExprOps>,
-}
+pub struct PyPropertyExprBuilder(pub(crate) Arc<dyn DynPropertyFilterBuilder>);
```
ljeub-pometry: you should make this `pub`, might be useful when extending the python interface

raphtory/src/python/filter/property_filter_builders.rs
```diff
@@ -6,207 +6,30 @@ use crate::{
...
 
 impl PyPropertyExprBuilder {
-    pub fn wrap<T: DynPropertyExprOps + 'static>(t: T) -> Self {
-        Self { inner: Arc::new(t) }
+    pub(crate) fn wrap<T: DynPropertyFilterBuilder + 'static>(t: T) -> Self {
```
ljeub-pometry: same, can keep this `pub` I think

raphtory/src/python/filter/property_filter_builders.rs
```diff
@@ -541,48 +353,68 @@ impl<'py> IntoPyObject<'py> for PyPropertyFilterBuilder {
...
+        PyPropertyFilterBuilder(self.0.property(name))
     }
 
     fn metadata(&self, name: String) -> PyPropertyExprBuilder {
-        self.0.metadata(name)
+        PyPropertyExprBuilder(self.0.metadata(name))
+    }
+
+    fn window(&self, start: EventTime, end: EventTime) -> PyViewFilterBuilder {
+        PyViewFilterBuilder(self.0.clone().window(start, end))
```
ljeub-pometry, line 386: there doesn't seem to be any tests for these (nested window/layers tests) as these probably hang currently due to the missing `deref`

ljeub-pometry (approved):

PR #2446, opened 2026-01-21
Features/graph filter

shivamka1:
Fixes https://github.com/Pometry/Raphtory/issues/2263


raphtory-graphql/src/model/graph/filtering.rs
```diff
@@ -1252,3 +1293,93 @@ impl TryFrom<GqlEdgeFilter> for CompositeEdgeFilter {
...
+    fn try_from(f: GqlGraphFilter) -> Result<Self, Self::Error> {
+        let default_inner: DynView = Arc::new(GraphFilter);
+
+        Ok(match f {
+            GqlGraphFilter::Window(w) => {
+                let inner: DynView = match w.expr {
+                    Some(e) => e.deref().clone().try_into()?,
+                    None => default_inner,
+                };
+                Arc::new(Windowed::new(w.start.0, w.end.0, inner))
```
ljeub-pometry: this should just call `inner.window` to make sure the implementation remains consistent (same for the other methods)

PR #2454, opened 2026-01-26
trigger graphql cache eviction automatically

ricopinazo:


raphtory-graphql/src/server.rs
```diff
@@ -200,6 +200,16 @@ impl GraphServer {
         self.data.vectorise_all_graphs_that_are_not().await?;
         let work_dir = self.data.work_dir.clone();
 
+        // Otherwise evictions are only triggered when the cache is actively touchedss
```
ljeub-pometry: typo

ricopinazo: It seems I was unsuccessfully trying to save the file 😆

PR #2461, opened 2026-01-29
Fast distributed dominating set

DanielLacina:


raphtory/src/algorithms/covering/dominating_set.rs
```diff
@@ -0,0 +1,338 @@
...
+                }
+            }
+        } else {
+            if actual_uncovered_count > 0 {
+                queue.insert(index, actual_uncovered_count);
+            }
+        }
+    }
+    dominating_set
+}
```
DanielLacina, line 124: https://kk.up45.ac.id/scholarhub/Lazy%20and%20Eager%20Approaches%20for%20the%20Set%20Cover%20Problem.pdf
Algorithm 3
Used a bucket queue instead of a heap to decrease the time complexity of maximum() and insert() from O(logn) to O(1).

raphtory/src/algorithms/covering/dominating_set.rs
```diff
@@ -0,0 +1,338 @@
...
+                linked_nodes[existing_index].prev = Some(index);
+            } 
+            uncovered_count_map[uncovered_count] = Some(index);
+        }
+        Self {
+            linked_nodes,
+            uncovered_count_map,
+            max_uncovered_count: n_nodes,
+        }
+    }
```
DanielLacina, line 50: Each raphtory node corresponds to the linked node - linked_nodes[node index] in the Dominating Set Queue. 
A node can cover 0 - n uncovered nodes. Therefore, the uncovered_count_map contains n + 1 linked lists where the ith linked list in the map corresponds to all the nodes who cover i uncovered nodes. 
The max_uncovered_count will stay at n but will be decremented later on request.
When each node is being processed in the for loop, their corresponding linked_node is updated and is set as the head of the ith linked list of the uncovered_count_map where i is the number of uncovered nodes it covers.

raphtory/src/algorithms/covering/dominating_set.rs
```diff
@@ -0,0 +1,338 @@
...
+                } else {
+                    self.uncovered_count_map[self.max_uncovered_count] = None;
+                }
+                self.linked_nodes[index].next = None;
+                return Some(index);
+            } 
+            self.max_uncovered_count -= 1;
+        }
+        None
+    }
```
DanielLacina, line 68: Decrements self.max_uncovered_count until it finds a non empty linked list. Then it removes the head of the linked list and makes its successor (if any) the head of the linked list. The node index corresponding to the head of the linked list is returned. 
Note: Any future value inserted into the Queue will have a value less than self.max_uncovered_count.

raphtory/src/algorithms/covering/dominating_set.rs
```diff
@@ -0,0 +1,338 @@
...
+    // uncovered count should be less than the max uncovered count
+    pub fn insert(&mut self, index: usize, uncovered_count: usize) {
+        self.linked_nodes[index].uncovered_count = uncovered_count;
+        if let Some(existing_index) = self.uncovered_count_map[uncovered_count] {
+            self.linked_nodes[index].next = Some(existing_index);
+            self.linked_nodes[existing_index].prev = Some(index);
+        }
+        self.uncovered_count_map[uncovered_count] = Some(index);
+    }
```
DanielLacina, line 77: Modifies the raphtory node corresponding linked node and makes it the head of the ith linked list. 
Note: Probably not a good practice to assume that the uncovered count should be less than the max uncovered count (assuming inputs probably is not ideal). Makes the code non reusable across modules. Also, should probably somehow integrate the insert method/function into the from_graph constructor to abstract upon the insertion logic.

raphtory/src/algorithms/covering/dominating_set.rs
```diff
@@ -0,0 +1,338 @@
...
+
+    pub fn node_details(&self, index: usize) -> (VID, usize) {
```
DanielLacina, line 79: straightforward

raphtory/src/algorithms/covering/dominating_set.rs
```diff
@@ -0,0 +1,338 @@
...
+        let node = g.node(vid).unwrap();
+        for neighbor in node.neighbours() {
+            if !covered_nodes[neighbor.node.index()] {
+                covered_nodes[neighbor.node.index()] = true;
+                covered_count += 1;
+            }
+        }
+    }
+    covered_count == n_nodes
+}
```
DanielLacina, line 144: Obtains the count of all unique nodes covered by all the nodes in the dominating set. By definition of dominating set, the count should be equal to n_nodes.

raphtory/src/algorithms/covering/fast_distributed_dominating_set.rs
```diff
@@ -0,0 +1,397 @@
...
+                    let neighbor_config = &mut current_node_configs[*neighbor_index];
+                    if !neighbor_config.is_covered {
+                        neighbor_config.is_covered = true;
+                        covered_count += 1;
+                    }
+                }
+            }
+        }
+    }
+    dominating_set
```
DanielLacina, line 214: https://ac.informatik.uni-freiburg.de/teaching/ss_12/netalg/lectures/chapter7.pdf
p. 5
Uses a double buffer method to prioritize minimal memory allocation
For each par_iter -> next_node_config is synchronized with current_node_config
and next_node_config is updated which becomes current_node_config. The two variables swap pointers. Logic is repeated for next par_iter.

PR #2474, opened 2026-02-05
Rust GraphQL Client

arienandalibi:
Adding a Rust client for managing and sending GraphQL requests to a server. Previously, the client was only available in Python as PyRaphtoryClient, along with types such as PyRemoteGraph, PyRemoteNode, etc... Now, these will also be available in Rust, which will be useful in our Rust tests.

### DIFF
--- a/raphtory-graphql/src/client/mod.rs
+++ b/raphtory-graphql/src/client/mod.rs
@@ -54,7 +54,7 @@ pub(crate) fn inner_collection(value: &Prop) -> String {
         }
         Prop::DTime(value) => format!("{{ str: {} }}", serde_json::to_string(value).unwrap()),
         Prop::NDTime(value) => format!("{{ str: {} }}", serde_json::to_string(value).unwrap()),
-        Prop::Decimal(value) => format!("{{ decimal: {} }}", value),
+        Prop::Decimal(value) => format!("{{ decimal: \"{}\" }}", value),
     }
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4497,6 +4497,7 @@ dependencies = [
  "async-graphql-poem",
  "base64 0.22.1",
  "base64-compat",
+ "bigdecimal",
  "chrono",
  "clap",
  "config",
--- a/raphtory-graphql/Cargo.toml
+++ b/raphtory-graphql/Cargo.toml
@@ -53,6 +53,7 @@ rayon = { workspace = true }
 ahash = { workspace = true }
 strum = {workspace = true}
 strum_macros = {workspace = true}
+bigdecimal = {workspace = true, features = ["serde"]}
 
 # python binding optional dependencies
 pyo3 = { workspace = true, optional = true }
--- a/raphtory-graphql/schema.graphql
+++ b/raphtory-graphql/schema.graphql
@@ -3319,13 +3319,17 @@ input Value @oneOf {
 	"""
 	object: [ObjectEntry!]
 	"""
-	Timezone-aware datetime
+	Timezone-aware datetime.
 	"""
 	dtime: String
 	"""
-	Naive datetime (no timezone)
+	Naive datetime (no timezone).
 	"""
 	ndtime: String
+	"""
+	Big Decimal numbers.
+	"""
+	decimal: String
 }
 
 type VectorSelection {
--- a/raphtory-graphql/src/model/graph/property.rs
+++ b/raphtory-graphql/src/model/graph/property.rs
@@ -6,6 +6,7 @@ use crate::{
     rayon::blocking_compute,
 };
 use async_graphql::{Error, Name, Value as GqlValue};
+use bigdecimal::BigDecimal;
 use dynamic_graphql::{
     InputObject, OneOfInput, ResolvedObject, ResolvedObjectFields, Scalar, ScalarValue,
 };
@@ -33,6 +34,7 @@ use std::{
     convert::TryFrom,
     fmt,
     fmt::{Display, Formatter},
+    str::FromStr,
     sync::Arc,
 };
 
@@ -70,10 +72,12 @@ pub enum Value {
     List(Vec<Value>),
     /// Object.
     Object(Vec<ObjectEntry>),
-    /// Timezone-aware datetime
+    /// Timezone-aware datetime.
     DTime(String),
-    /// Naive datetime (no timezone)
+    /// Naive datetime (no timezone).
     NDTime(String),
+    /// BigDecimal number (string representation, e.g. "3.14159" or "123e-5").
+    Decimal(String),
 }
 
 impl Display for Value {
@@ -102,6 +106,7 @@ impl Display for Value {
             }
             Value::DTime(v) => write!(f, "DTime({})", v),
             Value::NDTime(v) => write!(f, "NDTime({})", v),
+            Value::Decimal(v) => write!(f, "Decimal({})", v),
         }
     }
 }
@@ -149,6 +154,14 @@ fn value_to_prop(value: Value) -> Result<Prop, GraphError> {
             t.dt()
                 .map(|dt| Prop::NDTime(dt.naive_utc()))
                 .map_err(GraphError::from)
+        }
+        Value::Decimal(s) => {
+            let bd = BigDecimal::from_str(&s).map_err(|e| GraphError::InvalidProperty {
+                reason: format!("Invalid Decimal: {e}"),
+            })?;
+            Prop::try_from_bd(bd).map_err(|e| GraphError::InvalidProperty {
+                reason: format!("Decimal too large: {e}"),
+            })
         }
     }
 }


raphtory-graphql/src/client/mod.rs
```diff
@@ -0,0 +1,507 @@
+//! Pure Rust GraphQL client for Raphtory GraphQL server.
+//!
+//! No Python dependency; usable from Rust and from PyO3 bindings.
```
fabubaker: Maybe remove this line, the "python dependency" part might be weird for someone reading this in the future...

raphtory-graphql/src/client/mod.rs
```diff
@@ -0,0 +1,507 @@
...
+        let mut buffer = Vec::new();
+        folder.zip_from_folder(Cursor::new(&mut buffer))?;
+
+        let variables = format!(
+            r#""path": "{}", "overwrite": {}, "graph": null"#,
+            path, overwrite
+        );
+        let operations = format!(
+            r#"{{
+            "query": "mutation UploadGraph($path: String!, $graph: Upload!, $overwrite: Boolean!) {{ uploadGraph(path: $path, graph: $graph, overwrite: $overwrite) }}",
```
fabubaker: This line is too long, can we break it up?

raphtory-graphql/src/client/mod.rs
```diff
@@ -0,0 +1,507 @@
...
+        Ok(Self {
+            url,
+            token,
+            client: Client::new(),
+        })
+    }
+
+    /// Execute a GraphQL query asynchronously.
+    /// Returns the `data` object as a map; errors if the response contains GraphQL `errors`.
+    pub async fn query_async(
```
fabubaker: You can remove the `_async` suffix on all these methods, seems redundant since the client only supports async calls anyways.

raphtory-graphql/src/python/client/remote_graph.rs
```diff
@@ -11,7 +14,6 @@ use raphtory_api::core::{
     utils::time::IntoTime,
 };
 use std::collections::HashMap;
-
 #[derive(Clone)]
 #[pyclass(name = "RemoteGraph", module = "raphtory.graphql")]
 pub struct PyRemoteGraph {
```
fabubaker, line 23: `PyRemoteGraph` should internally hold a `GraphQLRemoteGraph` and call methods on it instead

raphtory-graphql/src/client/mod.rs
arienandalibi: I actually can't :/ tests start failing if I do: "control character (\u0000-\u001F) found while parsing a string"

raphtory-graphql/src/python/client/remote_graph.rs
```diff
@@ -58,7 +81,7 @@ impl PyRemoteGraph {
     /// Returns:
     ///     None:
     #[pyo3(signature = (updates))]
-    pub fn add_nodes(&self, py: Python, updates: Vec<PyNodeAddition>) -> Result<(), GraphError> {
```
fabubaker, line 61: Might want to leave a TODO for implementing `add_nodes` and `add_edges` in the Rust client in case someone was expecting it to be there...

raphtory-graphql/src/python/client/raphtory_client.rs
```diff
@@ -31,88 +27,32 @@ use tracing::debug;
...
-                return Err(exception);
-            }
-        }
-        match graphql_result.remove("data") {
-            Some(JsonValue::Object(data)) => Ok(data.into_iter().collect()),
-            _ => Err(PyException::new_err(format!(
-                "Error while reading server response for query:\n\t{graphql_query}"
-            ))),
-        }
+        self.run_async(move |client| async move { client.query(&query, variables).await })
```
fabubaker, line 48: We're still raising `PyException` correctly here right?

raphtory-graphql/src/python/client/raphtory_client.rs
```diff
@@ -1,28 +1,24 @@
...
-use raphtory_api::python::error::adapt_err_value;
-use reqwest::{multipart, multipart::Part, Client};
-use serde_json::{json, Value as JsonValue};
-use std::{collections::HashMap, future::Future, io::Cursor, sync::Arc};
+use pyo3::{exceptions::PyException, prelude::*, types::PyDict};
+use raphtory::db::api::view::MaterializedGraph;
+use serde_json::Value as JsonValue;
+use std::{collections::HashMap, future::Future, sync::Arc};
 use tokio::runtime::Runtime;
 use tracing::debug;
```
fabubaker, line 17: Can you leave a small comment mentioning this converts the rust client error to a `PyException`? Was easy to gloss over it otherwise.

raphtory-graphql/src/client/mod.rs
fabubaker: Fine for now assuming the queries will be rewritten using `graphql_client` at some point.

raphtory-graphql/src/python/client/remote_edge.rs
```diff
@@ -52,32 +48,15 @@ impl PyRemoteEdge {
...
-        let query_context = context! {
-            path => self.path,
-            src => self.src,
-            dst => self.dst,
-            t => t.into_time().t(),
-            properties =>  properties.map(|p| build_property_string(p)),
-            layer => layer
-        };
+    ) -> Result<(), ClientError> {
+        let edge = self.edge.clone();
```
fabubaker: `self.edge` here should be behind an `Arc`, otherwise we're cloning a bunch of fields unnecessarily every time we perform an operation on a `PyRemoteEdge`.

raphtory-graphql/src/python/client/remote_graph.rs
```diff
@@ -1,22 +1,39 @@
...
-use std::collections::HashMap;
+use std::{collections::HashMap, future::Future, sync::Arc};
+use tokio::runtime::Runtime;
 
 #[derive(Clone)]
 #[pyclass(name = "RemoteGraph", module = "raphtory.graphql")]
 pub struct PyRemoteGraph {
-    pub(crate) path: String,
-    pub(crate) client: PyRaphtoryClient,
+    pub(crate) graph: GraphQLRemoteGraph,
```
fabubaker: Need to wrap `graph` in an `Arc`.

raphtory-graphql/src/python/client/remote_node.rs
```diff
@@ -1,22 +1,14 @@
...
+use std::{collections::HashMap, future::Future, sync::Arc};
+use tokio::runtime::Runtime;
 
 #[derive(Clone)]
 #[pyclass(name = "RemoteNode", module = "raphtory.graphql")]
 pub struct PyRemoteNode {
-    pub(crate) path: String,
-    pub(crate) client: PyRaphtoryClient,
-    pub(crate) id: String,
+    pub(crate) node: GraphQLRemoteNode,
```
fabubaker: Wrap `node` in an `Arc`.

raphtory-graphql/src/python/client/raphtory_client.rs
arienandalibi: Yup, query() returns ClientError which gets turned into PyException using the From<ClientError> impl in src/python/client/mod.rs:337

raphtory-graphql/src/python/client/raphtory_client.rs
arienandalibi: I got rid of this and used a proper From<ClientError> for PyErr impl in src/python/client/mod.rs:337 (with a comment explaining)

raphtory-graphql/src/python/client/remote_edge.rs
```diff
@@ -52,32 +51,15 @@ impl PyRemoteEdge {
     #[pyo3(signature = (t, properties=None, layer=None))]
     fn add_updates(
         &self,
-        py: Python,
```
fabubaker, line 55: We're are no longer passing in the GIL into this function. I'm not very familiar with pyo3 so one of @fabianmurariu or @ljeub-pometry will have to confirm this is ok.

fabianmurariu: I think execute_async_task should take care of that, once I merge pedro's PR

raphtory-graphql/src/python/client/remote_edge.rs
```diff
@@ -19,22 +12,28 @@ use std::collections::HashMap;
...
         }
     }
+
+    fn execute_async_task<T, F, O>(&self, task: T) -> O
+    where
+        T: FnOnce() -> F + Send + 'static,
+        F: Future<Output = O> + 'static,
+        O: Send + 'static,
+    {
+        pyo3::Python::attach(|py| py.detach(|| self.runtime.block_on(task())))
```
fabianmurariu: see [this](https://github.com/Pometry/Raphtory/pull/2249/changes#diff-c7f5926d91dfd68186100096a331e0c60467136bfef9ca1d1dc01592a3437ff9R424) PR
I don't think we need to carry around an Arc<Runtime>, we can have a static lazy runtime

raphtory-graphql/src/python/client/remote_edge.rs
arienandalibi: I copy pasted the execute_async_task function from pedro's PR for now

PR #2477, opened 2026-02-06
implement config override on load

ljeub-pometry:

### DIFF
--- a/db4-storage/src/persist/config.rs
+++ b/db4-storage/src/persist/config.rs
@@ -13,10 +13,14 @@ pub const DEFAULT_MAX_PAGE_LEN_NODES: u32 = 131_072; // 2^17
 pub const DEFAULT_MAX_PAGE_LEN_EDGES: u32 = 1_048_576; // 2^20
 pub const CONFIG_FILE: &str = "config.json";
 
-pub trait ConfigOps: Serialize + DeserializeOwned + Args {
+pub trait ConfigOps: Serialize + DeserializeOwned + Args + Sized {
     fn max_node_page_len(&self) -> u32;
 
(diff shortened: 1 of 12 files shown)


db4-storage/src/persist/config.rs
```diff
@@ -123,6 +125,10 @@ impl ConfigOps for BaseConfig {
     fn with_node_types(&self, _node_types: impl IntoIterator<Item = impl AsRef<str>>) -> Self {
         *self
     }
+
+    fn update(&mut self, _new: Self) {
+        // cannot update page lengths for an existing graph
```
fabubaker, line 144: Might as well panic here.

raphtory/src/db/graph/graph.rs
```diff
@@ -198,7 +198,29 @@ impl Graph {
...
+    /// * `config` - The new config (note that it is not possible to change the page sizes)
+    /// # Returns
+    /// A raphtory graph loaded from the specified path
+    /// # Example
+    /// ```no_run
+    /// use raphtory::prelude::Graph;
+    /// let g = Graph::load_from_path("/path/to/storage");
+    ///
+    #[cfg(feature = "io")]
+    pub fn load_from_path_with_config(
```
fabubaker: For the `load` functions, the `from_part` seems redundant? eg this could just be `load_with_config`.

fabubaker (approved):
LGTM besides some minor nits.

db4-storage/src/persist/config.rs
ljeub-pometry: I would rather we don't panic randomly here, means we can call this safely in other places without shooting ourselves in the foot.

PR #2481, opened 2026-02-09
Add wal support for `add_node` / `create_node`

fabubaker:
Adds wal logging and replay for `add_node` & `create_node` operations.

### DIFF
--- a/raphtory/src/db/api/mutation/addition_ops.rs
+++ b/raphtory/src/db/api/mutation/addition_ops.rs
@@ -300,7 +300,7 @@ impl<G: InternalAdditionOps<Error: Into<GraphError>> + StaticGraphViewOps> Addit
         // Flush the wal entry to disk.
         // Any error here is fatal.
         if let Err(e) = wal.flush(lsn) {
-            panic!("Failed to flush add_edge wal entry: {}", e);
+            return Err(GraphError::FatalWriteError(e));
         }
 
         Ok(EdgeView::new(
@@ -417,7 +417,7 @@ fn add_node_impl<
     // Flush the wal entry to disk.
     // Any error here is fatal.
     if let Err(e) = wal.flush(lsn) {
-        panic!("Failed to flush add_node wal entry: {}", e);
+        return Err(GraphError::FatalWriteError(e));
     }
 
     Ok(NodeView::new_internal(graph.clone(), node_id))
--- a/raphtory/src/errors.rs
+++ b/raphtory/src/errors.rs
@@ -468,6 +468,9 @@ pub enum GraphError {
 
     #[error(transparent)]
     StorageError(#[from] StorageError),
+
+    #[error("Fatal write error: {0}")]
+    FatalWriteError(StorageError),
 }
 
 impl From<MetadataError> for GraphError {


db4-graph/src/replay.rs
```diff
@@ -233,4 +233,96 @@ where
...
+                .logical_to_physical
+                .set(name.as_ref(), node_id)?;
+        }
+
+        // Resolve segment and check LSN.
+        let node_max_page_len = self.graph().extension().config().max_node_page_len();
+        let (segment_id, pos) = resolve_pos(node_id, node_max_page_len);
+        let resize_vid = VID::from(node_id.index() + 1);
+        self.resize_chunks_to_vid(resize_vid);
+
```
ljeub-pometry: would be clearer if the `x+1` happens as part of `resize_chunks_to_vid`

db4-graph/src/replay.rs
```diff
@@ -233,4 +233,96 @@ where
...
+        if let Some(ref name) = node_name {
+            self.graph()
+                .logical_to_physical
+                .set(name.as_ref(), node_id)?;
+        }
+
+        // Resolve segment and check LSN.
+        let node_max_page_len = self.graph().extension().config().max_node_page_len();
+        let (segment_id, pos) = resolve_pos(node_id, node_max_page_len);
+        let resize_vid = VID::from(node_id.index() + 1);
```
ljeub-pometry: size feels like it should be a method on the graph?

raphtory/src/db/api/mutation/addition_ops.rs
```diff
@@ -161,31 +164,96 @@ impl<G: InternalAdditionOps<Error: Into<GraphError>> + StaticGraphViewOps> Addit
...
                 false,
                 self.node_meta(),
                 props.into_iter().map(|(k, v)| (k, v.into())),
             )
             .map_err(into_graph_err)?;
+
+        let node_gid = match node_ref {
+            NodeRef::Internal(_) => None,
+            NodeRef::External(gid_ref) => Some(gid_ref),
+        };
```
ljeub-pometry: we should just add this as a `gid()` method on `NodeRef` such that we don't have to keep writing this match statement.

raphtory/src/db/api/mutation/addition_ops.rs
```diff
@@ -161,31 +164,96 @@ impl<G: InternalAdditionOps<Error: Into<GraphError>> + StaticGraphViewOps> Addit
...
+
+            self.core_graph()
+                .transaction_manager()?
+                .end_transaction(transaction_id);
+
+            // Segment lock can be released before flush to allow
+            // other operations to proceed.
+            drop(writer);
+
+            self.core_graph().wal()?.flush(lsn)?;
```
ljeub-pometry: Since you put everything into a block already, you should just move this line out of the block which will remove the need for the explicit `drop` and make the scoping more clear.

ljeub-pometry: Also, is there any way we can avoid having so many failure points in this critical section? The whole point was that this shouldn't go wrong!

raphtory/src/db/api/mutation/addition_ops.rs
```diff
@@ -202,37 +270,99 @@ impl<G: InternalAdditionOps<Error: Into<GraphError>> + StaticGraphViewOps> Addit
...
+
```
ljeub-pometry: move the wall flush out of this block to avoid the explicit drop

raphtory/src/db/api/mutation/addition_ops.rs
```diff
@@ -202,37 +270,99 @@ impl<G: InternalAdditionOps<Error: Into<GraphError>> + StaticGraphViewOps> Addit
...
+            Ok::<_, GraphError>(node_id)
+        }?;
```
ljeub-pometry: This whole thing is the same between `add_node` and `create_node`, we should make it a function.

raphtory/src/db/api/mutation/addition_ops.rs
```diff
@@ -357,19 +487,20 @@ impl<G: InternalAdditionOps<Error: Into<GraphError>> + StaticGraphViewOps> Addit
...
             // Flush the wal entry to disk.
-            // Any error here is fatal
+            // Any error here is fatal.
             self.core_graph().wal()?.flush(lsn)?;
-            Ok::<_, GraphError>((edge_id, src_id, dst_id, layer_id))
-        }
-        .unwrap();
+
+            Ok::<_, GraphError>((edge_id.inner(), src_id, dst_id, layer_id))
+        }?;
```
ljeub-pometry: there is a reason this was panicking before, the graph is corrupted if this fails

ljeub-pometry: We should at least signal that in the error message if we don't like the panic!

db4-graph/src/replay.rs
fabubaker: It turns out there was already a +1 in `resize_chunks_to_vid` and the extra +1 here isn't needed.

db4-graph/src/replay.rs
fabubaker: yep, switched to using the appropriate methods on `GraphStore`.

raphtory/src/db/api/mutation/addition_ops.rs
fabubaker: Switched to using `as_gid_ref().left()`

raphtory/src/db/api/mutation/addition_ops.rs
fabubaker: More broadly, `add_node` and `create_node` are pretty much the same, so I've created an `add_node_impl` and used it instead.

raphtory/src/db/api/mutation/addition_ops.rs
fabubaker: Since it's really only the wal flush that's problematic, I'm inclined to move the panic there instead of on the whole block. The other failures all happen before the writers are acquired, which at least prevents anything from getting to disk. In fact, there are no `Error` returns after the writers are acquired and before the `flush`, so nothing should get corrupted.

Lmk if you think otherwise.

fabubaker: I'm thinking we should maybe even be pushing the `panic!` into `wal.flush` if it fails...

fabubaker: As discussed we're removing the panic for now and replacing it with a critical path error.

PR #2486, opened 2026-02-11
Add python test for float round trip error in GraphQL 

arienandalibi:
Added a python test which reproduces a bug where float values as properties in a graph change very slightly after being sent to the GraphQL server and being retrieved in some way. It seems that some floats never change and other floats change every time.

### DIFF
--- a/docs/reference/graphql/graphql_API.md
+++ b/docs/reference/graphql/graphql_API.md
@@ -9198,7 +9198,7 @@ Object.
 <td valign="top"><a href="#string">String</a></td>
 <td>
 
-Timezone-aware datetime
+Timezone-aware datetime.
 
 </td>
 </tr>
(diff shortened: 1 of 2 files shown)


python/tests/test_base_install/test_graphql/edit_graph/test_graphql.py
```diff
@@ -840,6 +840,47 @@ def test_graph_persistence_across_restarts():
...
+                          }}
+                        }}
+                      }}
+                    }}
+                  }}
+                }}
+            """
+            resp = client.query(query)
+            retrieved_float = resp["graph"]["node"]["at"]["properties"]["get"]["value"]
+            if retrieved_float != num:
```
fabubaker: We should switch these to asserts.

python/tests/test_base_install/test_graphql/edit_graph/test_graphql.py
```diff
@@ -840,6 +840,47 @@ def test_graph_persistence_across_restarts():
...
+                print("Value of float changed after roundtrip from GraphQL server")
+                print(f"Old value: {num}")
+                print(f"New value: {retrieved_float}\n")
+            else:
+                print("Success: Float did not change value")
+
```
fabubaker, line 878: As a sanity check, try loading the graph using `Graph.load(os.path.join(tmp_work_dir, "g"))` and see if the float values are the same or not.

python/tests/test_base_install/test_graphql/edit_graph/test_graphql.py
```diff
@@ -840,6 +840,47 @@ def test_graph_persistence_across_restarts():
         }
 
 
+def test_reproduce_float_changes_on_roundtrip():
```
fabubaker: Should link to https://github.com/Pometry/Raphtory/issues/2487 as a comment here.

PR #2490, opened 2026-02-12
Fix concurrent updates

ljeub-pometry:

Both node and edge additions were making newly added nodes available before they were initialised, leading to errors with concurrent read operations

rework of the internal apis only

the graphql stress-test

the addition apis need a major cleanup

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,6 +8,7 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
+
 jobs:
   rust-format-check:
     name: Rust format check
@@ -22,12 +23,12 @@ jobs:
     with:
       test_python_lower: true
     secrets: inherit
-  # call-test-ui:
-  #   name: Run UI Tests
-  #   uses: ./.github/workflows/test_ui.yml
-  #   secrets: inherit
-  #   permissions:
-  #     contents: read
+  call-test-ui:
+    name: Run UI Tests
+    uses: ./.github/workflows/test_ui.yml
+    secrets: inherit
+    permissions:
+      contents: read
   call-benchmark-workflow-in-local-repo:
     name: Run benchmarks
     uses: ./.github/workflows/benchmark.yml
--- a/.github/workflows/test_during_pr.yml
+++ b/.github/workflows/test_during_pr.yml
@@ -21,12 +21,12 @@ jobs:
     with:
       test_python_lower: false
     secrets: inherit
-  # call-test-ui-in-local-repo:
-  #   name: Run UI Tests
-  #   uses: ./.github/workflows/test_ui.yml
-  #   secrets: inherit
-  #   permissions:
-  #     contents: read
+  call-test-ui-in-local-repo:
+    name: Run UI Tests
+    uses: ./.github/workflows/test_ui.yml
+    secrets: inherit
+    permissions:
+      contents: read
   call-benchmark-workflow-in-local-repo:
     name: Run benchmarks
     uses: ./.github/workflows/benchmark.yml
@@ -43,15 +43,14 @@ jobs:
     name: Formatting autofixes
     uses: ./.github/workflows/format.yml
     secrets: inherit
-    needs:
-      [
-        call-test-rust-workflow-in-local-repo,
-        call-test-python-workflow-in-local-repo,
-        call-test-ui-in-local-repo,
-        call-benchmark-workflow-in-local-repo,
-        call-graphql-bench-workflow-in-local-repo,
-        call-stress-test-workflow-in-local-repo,
-      ]
+    needs: [
+      call-test-rust-workflow-in-local-repo,
+      call-test-python-workflow-in-local-repo,
+      call-test-ui-in-local-repo,
+      call-benchmark-workflow-in-local-repo,
+      call-graphql-bench-workflow-in-local-repo,
+      call-stress-test-workflow-in-local-repo
+    ]
   # call-code-coverage:
   #   name: Code Coverage
   #   uses: ./.github/workflows/code_coverage.yml


db4-storage/src/pages/node_store.rs
```diff
@@ -140,10 +151,22 @@ impl<NS: Send + Sync, EXT: PersistenceStrategy> NodeStorageInner<NS, EXT> {
...
+            self.get_segment(id)
+                .expect("segment should exist given count")
+        })
     }
 
+    pub fn segments_count(&self) -> usize {
+        self.segments.count()
+    }
+
+    // pub fn segments(&self) -> &boxcar::Vec<Arc<NS>> {
```
fabubaker: delete

db4-storage/src/pages/node_store.rs
```diff
@@ -140,10 +151,22 @@ impl<NS: Send + Sync, EXT: PersistenceStrategy> NodeStorageInner<NS, EXT> {
...
 
-    pub fn segments(&self) -> &boxcar::Vec<Arc<NS>> {
-        &self.segments
+    pub fn segments_iter(&self) -> impl Iterator<Item = &NS> {
+        let count = self.segments.count();
+        (0..count).map(|id| {
+            self.get_segment(id)
+                .expect("segment should exist given count")
+        })
     }
```
fabubaker, line 160: Can't this just be `self.segments.iter()`?

fabubaker: NVM, I read the boxcar docs on `iter`: https://docs.rs/boxcar/latest/boxcar/struct.Vec.html#method.iter

We should at least leave a comment about this here, although ideally I think we should wrap boxcar in a custom type to prevent footguns when calling `iter` or `count` + `get`.

raphtory-core/src/entities/nodes/node_ref.rs
```diff
@@ -21,10 +21,10 @@ pub trait AsNodeRef: Send + Sync {
         }
     }
 
-    fn as_gid_ref(&self) -> Either<GidRef<'_>, VID> {
+    fn as_gid_ref(&self) -> Option<GidRef<'_>> {
```
fabubaker, line 24: I did want to make this change at some point during one of my wal PRs because:

A. Using `Either` here is not very descriptive
B. it's confusing when a method named `as_gid_ref` returns a `VID`

db4-storage/src/pages/node_store.rs
```diff
@@ -264,10 +285,47 @@ impl<NS: NodeSegmentOps<Extension = EXT>, EXT: PersistenceStrategy> NodeStorageI
...
+    pub fn reserve_segment_row(&self, segment: &NS) -> Option<u32> {
         // TODO: if this becomes a hotspot, we can switch to a fetch_add followed by a fetch_min
         // this means when we read the counter we need to clamp it to max_page_len so the iterators don't break
-        increment_and_clamp(segment.nodes_counter(), self.max_segment_len())
+        increment_and_clamp(segment.nodes_counter(), 1, self.max_segment_len())
+    }
+
+    fn reserve_segment_rows(&self, segment: &NS, rows: u32) -> Option<u32> {
+        increment_and_clamp(segment.nodes_counter(), rows, self.max_segment_len())
     }
```
fabubaker, line 354: I think we need comments describing why these two methods return `Option`...

db4-storage/src/pages/node_store.rs
```diff
@@ -264,10 +285,47 @@ impl<NS: NodeSegmentOps<Extension = EXT>, EXT: PersistenceStrategy> NodeStorageI
         }
     }
 
-    fn reserve_segment_row(&self, segment: &Arc<NS>) -> Option<u32> {
+    pub fn reserve_and_lock_segment(
+        &self,
+        row: usize,
+        required_space: u32,
```
fabubaker, line 315: For clarity I think these are better named:

```suggestion
        vid: usize,
        num_rows: u32,
```
since we are reserving a fixed number of rows in a segment for a vid.

db4-storage/src/pages/node_store.rs
```diff
@@ -264,10 +285,47 @@ impl<NS: NodeSegmentOps<Extension = EXT>, EXT: PersistenceStrategy> NodeStorageI
...
+                    page_id = *self.free_segments[slot_idx].read_recursive();
+                }
+                Some(writer) => {
+                    match self.reserve_segment_rows(writer.page, required_space) {
+                        None => {
+                            // segment is full, we need to create a new one
+                            let mut slot = self.free_segments[slot_idx].write();
+                            if *slot == page_id {
+                                // page_id is unchanged, no other thread created a new segment before we got the lock
+                                page_id = self.push_new_segment();
```
fabubaker, line 335: Don't we have to update `slot = push_new_segment` here?

db4-storage/src/pages/node_store.rs
```diff
@@ -264,10 +285,47 @@ impl<NS: NodeSegmentOps<Extension = EXT>, EXT: PersistenceStrategy> NodeStorageI
         }
     }
 
-    fn reserve_segment_row(&self, segment: &Arc<NS>) -> Option<u32> {
+    pub fn reserve_and_lock_segment(
```
fabubaker, line 312: This method is fairly complex and definitely needs an overview of what the locking strategy is...

db4-storage/src/pages/node_store.rs
fabubaker: Also you avoid starvation here if the thread that created the new segment is able to lock it and immediately use it, rather than returning to the loop and calling `try_writer`.

But that might be easier said than done because we'll need a `push_and_lock_new_segment`.

db4-storage/src/pages/node_store.rs
fabianmurariu: row is not vid, it's just a number that the caller needs to keep incrementing such that we don't always hit the same segment in the free list, also required space basically asks the free list how much space do we need from a segment, so I'd say the names are fine, i'll add a comment on the function

db4-storage/src/pages/node_store.rs
fabianmurariu: I'm adding a comment in increment_and_clamp

fabianmurariu (approved):
LGTM

PR #2493, opened 2026-02-13
Implement wal logging and replay for `EdgeView`, `NodeView`, `delete_edge` and graph properties

fabubaker:
Adds wal logging and replay for `add_metadata`, `update_metadata` and `add_updates` on an `EdgeView` and `NodeView`.

Additionally adds wal support for `delete_edge` and all operations on graph properties.

### DIFF
--- a/db4-storage/src/persist/config.rs
+++ b/db4-storage/src/persist/config.rs
@@ -4,9 +4,8 @@ use clap::{
     builder::{TypedValueParser, ValueParser},
     error::{ContextKind, ContextValue, Error},
 };
-use itertools::Itertools;
-use serde::{Deserialize, Deserializer, Serialize, de::DeserializeOwned};
-use std::{env::var, ffi::OsStr, iter, path::Path};
+use serde::{Deserialize, Serialize, de::DeserializeOwned};
+use std::{iter, path::Path};
 use tracing::error;
 
 pub const DEFAULT_MAX_PAGE_LEN_NODES: u32 = 131_072; // 2^17
@@ -16,13 +15,13 @@ pub const CONFIG_FILE: &str = "config.json";
 pub trait ConfigOps: Serialize + DeserializeOwned + Args + Sized {
     fn max_node_page_len(&self) -> u32;
 
-    fn with_max_node_page_len(self, page_len: u32) -> Self;
-
     fn max_edge_page_len(&self) -> u32;
 
-    fn with_max_edge_page_len(self, page_len: u32) -> Self;
-
     fn node_types(&self) -> &[String];
+
+    fn with_max_node_page_len(self, page_len: u32) -> Self;
+
+    fn with_max_edge_page_len(self, page_len: u32) -> Self;
 
     fn with_node_types(&self, node_types: impl IntoIterator<Item = impl AsRef<str>>) -> Self;
 
@@ -48,6 +47,7 @@ pub trait ConfigOps: Serialize + DeserializeOwned + Args + Sized {
 pub struct BaseConfig {
     #[arg(long, default_value_t=DEFAULT_MAX_PAGE_LEN_NODES, env="RAPHTORY_MAX_NODE_PAGE_LEN")]
     max_node_page_len: u32,
+
     #[arg(long, default_value_t=DEFAULT_MAX_PAGE_LEN_EDGES, env="RAPHTORY_MAX_EDGE_PAGE_LEN")]
     max_edge_page_len: u32,
 }
@@ -118,13 +118,13 @@ impl ConfigOps for BaseConfig {
         self.max_node_page_len
     }
 
+    fn max_edge_page_len(&self) -> u32 {
+        self.max_edge_page_len
+    }
+
     fn with_max_node_page_len(mut self, page_len: u32) -> Self {
         self.max_node_page_len = page_len;
         self
-    }
-
-    fn max_edge_page_len(&self) -> u32 {
-        self.max_edge_page_len
     }
 
     fn with_max_edge_page_len(mut self, page_len: u32) -> Self {
@@ -150,7 +150,6 @@ mod tests {
     use crate::persist::config::{
         BaseConfig, DEFAULT_MAX_PAGE_LEN_EDGES, DEFAULT_MAX_PAGE_LEN_NODES,
     };
-    use std::env;
 
     #[test_log::test]
     fn test_default() {
--- a/db4-storage/src/persist/strategy.rs
+++ b/db4-storage/src/persist/strategy.rs
@@ -55,7 +55,7 @@ pub trait PersistenceStrategy: Debug + Clone + Send + Sync + 'static {
     ) where
         Self: Sized;
 
-    /// Indicate whether the strategy persists to disk or not.
+    /// Indicates whether the strategy persists to disk or not.
     fn disk_storage_enabled() -> bool;
 }
 


db4-graph/src/replay.rs
```diff
@@ -168,24 +165,13 @@ where
...
-                    }
-                    Some(old_dtype) => {
-                        let dtype = prop_value.dtype();
-                        let mut unified = false;
-                        let new_dtype = unify_types(&old_dtype, &dtype, &mut unified)?;
-                        if unified {
-                            prop_mapper.set_dtype(*prop_id, new_dtype);
-                        }
-                    }
-                }
```
fabubaker, line 188: `set_or_unify_id_and_dtype` is not exactly the same as this block, since it always sets the id whereas this doesn't, but minor differences aside lmk if it's ok to use this method here.

db4-graph/src/replay.rs
```diff
@@ -222,7 +208,102 @@ where
...
+                let mut write_locked_mapper = prop_mapper.write_locked();
+
+                write_locked_mapper.set_or_unify_id_and_dtype(
+                    prop_name.as_ref(),
+                    *prop_id,
+                    prop_value.dtype(),
+                )?;
+            }
+
+            let mut edge_writer = self.edges.get_mut(edge_segment_id).unwrap().writer();
```
fabianmurariu: `unwrap()` i'd switch to [ok_or_else](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else)

db4-graph/src/replay.rs
```diff
@@ -222,7 +208,102 @@ where
...
+                let mut write_locked_mapper = prop_mapper.write_locked();
+
+                write_locked_mapper.set_or_unify_id_and_dtype(
+                    prop_name.as_ref(),
+                    *prop_id,
+                    prop_value.dtype(),
+                )?;
+            }
+
+            let mut node_writer = self.nodes.get_mut(segment_id).unwrap().writer();
```
fabianmurariu: more unwrap

db4-graph/src/replay.rs
```diff
@@ -222,7 +208,102 @@ where
...
+
+            let props_vec: Vec<_> = props.iter().map(|(_, id, p)| (*id, p.clone())).collect();
```
fabianmurariu: I don't think collect is needed

db4-graph/src/replay.rs
```diff
@@ -222,7 +208,102 @@ where
...
+
+            let mut edge_writer = self.edges.get_mut(edge_segment_id).unwrap().writer();
+
+            let (src, dst) = edge_writer.get_edge(layer_id, edge_pos).ok_or_else(|| {
+                StorageError::GenericFailure(format!(
+                    "Edge {eid:?} not found in layer {layer_id} during replay_add_edge_metadata"
+                ))
+            })?;
+
+            let props_vec: Vec<_> = props.iter().map(|(_, id, p)| (*id, p.clone())).collect();
```
fabianmurariu: probably needless collect

db4-graph/src/replay.rs
```diff
@@ -168,24 +177,13 @@ where
             // Insert prop ids into edge meta.
             for (prop_name, prop_id, prop_value) in &props {
                 let prop_mapper = edge_meta.temporal_prop_mapper();
+                let mut write_locked_mapper = prop_mapper.write_locked();
```
ljeub-pometry: this should be outside the loop! you keep locking and unlocking the mapper for no reason

ljeub-pometry: Actually, maybe it would make more sense to have this as part of the initialisation of the write-locked graph itself?

db4-graph/src/replay.rs
ljeub-pometry: we can always optimise that later, so using that method is fine.

db4-graph/src/replay.rs
```diff
@@ -222,7 +226,164 @@ where
...
+            .get_or_create_segment(edge_segment_id);
+
+        let immut_lsn = segment.immut_lsn();
+
+        if immut_lsn < lsn {
+            let edge_meta = self.graph().edge_meta();
+
+            for (prop_name, prop_id, prop_value) in &props {
+                let prop_mapper = edge_meta.metadata_mapper();
+                let mut write_locked_mapper = prop_mapper.write_locked();
```
ljeub-pometry: As before, we should not have this inside the loop. Probably lock these as part of the write-locked graph initialisation.

db4-graph/src/replay.rs
```diff
@@ -222,7 +226,164 @@ where
...
+
+        Ok(())
+    }
+
+    fn replay_delete_edge(
+        &mut self,
+        lsn: LSN,
+        _transaction_id: TransactionID,
+        t: EventTime,
+        _src_name: Option<GID>,
```
ljeub-pometry: It is possible to create new nodes on a delete_edge call, you need to handle these!

db4-graph/src/replay.rs
```diff
@@ -263,23 +424,13 @@ where
...
-                    Some(old_dtype) => {
-                        let dtype = prop_value.dtype();
-                        let mut unified = false;
-                        let new_dtype = unify_types(&old_dtype, &dtype, &mut unified)?;
-                        if unified {
-                            prop_mapper.set_dtype(*prop_id, new_dtype);
-                        }
-                    }
-                }
+                let mut write_locked_mapper = prop_mapper.write_locked();
```
ljeub-pometry: As before, don't call this in the loop.

db4-graph/src/replay.rs
```diff
@@ -313,7 +470,163 @@ where
...
+            .get_or_create_segment(segment_id);
+
+        let immut_lsn = segment.immut_lsn();
+
+        if immut_lsn < lsn {
+            let node_meta = self.graph().node_meta();
+
+            for (prop_name, prop_id, prop_value) in &props {
+                let prop_mapper = node_meta.metadata_mapper();
+                let mut write_locked_mapper = prop_mapper.write_locked();
```
ljeub-pometry: Need to extract this from the loop

db4-graph/src/replay.rs
```diff
@@ -313,7 +470,163 @@ where
...
+    ) -> Result<(), StorageError> {
+        let segment = self.graph().storage().graph_props().segment();
+        let immut_lsn = segment.immut_lsn();
+
+        if immut_lsn < lsn {
+            let graph_props_meta = self.graph().graph_props_meta();
+
+            for (prop_name, prop_id, prop_value) in &props {
+                let prop_mapper = graph_props_meta.temporal_prop_mapper();
+                let mut write_locked_mapper = prop_mapper.write_locked();
```
ljeub-pometry: move out

db4-graph/src/replay.rs
```diff
@@ -313,7 +470,163 @@ where
...
+    ) -> Result<(), StorageError> {
+        let segment = self.graph().storage().graph_props().segment();
+        let immut_lsn = segment.immut_lsn();
+
+        if immut_lsn < lsn {
+            let graph_props_meta = self.graph().graph_props_meta();
+
+            for (prop_name, prop_id, prop_value) in &props {
+                let prop_mapper = graph_props_meta.metadata_mapper();
+                let mut write_locked_mapper = prop_mapper.write_locked();
```
ljeub-pometry: move out

db4-graph/src/replay.rs
```diff
@@ -313,7 +470,163 @@ where
...
+
+                write_locked_mapper.set_or_unify_id_and_dtype(
+                    prop_name.as_ref(),
+                    *prop_id,
+                    prop_value.dtype(),
+                )?;
+            }
+
+            let writer = self.graph_props.writer();
+            let props = props.iter().map(|(_, id, p)| (*id, p.clone()));
```
ljeub-pometry: use into_iter to avoid the clone

db4-graph/src/replay.rs
```diff
@@ -313,7 +470,163 @@ where
...
```
ljeub-pometry: use into_iter to avoid the clone

db4-graph/src/replay.rs
```diff
@@ -313,7 +470,163 @@ where
...
+            }
+
+            let node_writer = self.nodes.get_mut(segment_id).ok_or_else(|| {
+                StorageError::GenericFailure(format!(
+                    "Node segment {segment_id} not found during replay_add_node_metadata"
+                ))
+            })?;
+
+            let mut node_writer = node_writer.writer();
+            let props = props.iter().map(|(_, id, p)| (*id, p.clone()));
```
ljeub-pometry: use into_iter to avoid the clone

db4-graph/src/replay.rs
```diff
@@ -222,7 +226,164 @@ where
...
+
+            let mut edge_writer = edge_writer.writer();
+
+            let (src, dst) = edge_writer.get_edge(layer_id, edge_pos).ok_or_else(|| {
+                StorageError::GenericFailure(format!(
+                    "Edge {eid:?} not found in layer {layer_id} during replay_add_edge_metadata"
+                ))
+            })?;
+
+            let props = props.iter().map(|(_, id, p)| (*id, p.clone()));
```
ljeub-pometry: use into_iter to avoid the clone

db4-graph/src/replay.rs
ljeub-pometry: use into_iter to avoid the clone

db4-storage/src/pages/node_store.rs
```diff
@@ -281,59 +284,47 @@ impl<NS: NodeSegmentOps<Extension = EXT>, EXT: PersistenceStrategy> NodeStorageI
...
+                    let slot = self.free_segments[slot_idx].read_recursive();
+                    segment_id = *slot;
                 }
                 Some(writer) => {
-                    match self.reserve_segment_rows(writer.page, required_space) {
+                    match self.reserve_segment_rows(writer.page, num_rows) {
                         None => {
-                            // segment is full, we need to create a new one
+                            // The current segment is full, drop its lock and try to find
+                            // another free segment.
```
ljeub-pometry: We are not trying to find a free segment here, we are going to replace the segment at this slot with a new one (unless another thread managed to replace it before we get to it)

db4-storage/src/segments/graph_prop/mod.rs
```diff
@@ -83,6 +84,10 @@ impl<P: PersistenceStrategy> GraphPropSegmentOps for GraphPropSegmentView<P> {
         self.is_dirty.store(dirty, Ordering::Release);
     }
 
+    fn immut_lsn(&self) -> LSN {
+        self.head.read().lsn()
```
ljeub-pometry: That doesn't seem like the immut_lsn since you are reading the mutable part?

db4-graph/src/replay.rs
fabubaker: Need to fix this...

raphtory/src/db/graph/edge.rs
fabubaker, line 167: Need to add wal logging for `EdgeView.delete` here

raphtory-storage/src/mutation/deletion_ops.rs
fabubaker, line 41: I noticed that here we're calling `session.add_static_edge` in addition to `delete_edge_from_layer`, but in `EdgeWriteLock for AtomicAddEdge`, we are only calling `delete_edge_from_layer`. Is this intentional? Do we not need to create the static edge in `AtomicAddEdge`?

fabubaker: Right, I see, it's already handled inside `atomic_add_edge`.

PR #2504, opened 2026-02-26
Db v4 layer optim

ljeub-pometry:

Optimise layers to actually use the available indexes in v4


db4-graph/src/lib.rs
```diff
@@ -197,13 +207,80 @@ where
...
+                .map(|segment| {
+                    let locked = segment.locked();
+                    locked
+                        .iter_entries()
+                        .filter(|entry| !entry.node_additions(STATIC_GRAPH_LAYER_ID).is_empty())
+                        .count()
+                })
+                .sum(),
+            LayerIds::All => self.storage.nodes().num_nodes(),
+            LayerIds::One(id) => self
```
fabianmurariu, line 225: so the layer counts still don't work?

fabianmurariu (approved):
Looks fine, just internal_num_nodes doesn't seem to use the layer counts

db4-graph/src/lib.rs
ljeub-pometry: Unfortunately not, they don't take updates to the node properties into account. We need to sort this out properly when we have layered nodes implemented.

PR #2507, opened 2026-02-26
Cleanup wal file on graceful shutdown

fabubaker:


db4-storage/src/pages/mod.rs
```diff
@@ -553,7 +562,32 @@ impl<
...
+                //
+                // Thus, we can safely discard all records with LSN <= latest_immut_lsn
+                // by rotating the WAL.
+                let latest_immut_lsn = self.latest_immut_lsn();
+
+                if let Err(e) = wal.rotate(latest_immut_lsn) {
+                    eprintln!("Failed to rotate WAL in drop: {}", e);
+                }
+
+                // FIXME: If the process crashes here after rotation, we lose the
```
fabubaker, line 572: I think this is fine for now? I'm thinking of having a separate control file for the wal for this, but prefer to do it properly with some time ahead of me.

fabianmurariu (approved):

PR #2530, opened 2026-03-25
Fix deadlock in resolving nodes in add_edge

ljeub-pometry:

make sure to resolve nodes in fixed order to avoid deadlocking the resolution


fabianmurariu (approved):
LGTM

PR #2532, opened 2026-03-26
Fix free segment eviction

ljeub-pometry:

Fix deadlock when free segments fill up


db4-storage/src/pages/node_store.rs
```diff
@@ -301,43 +303,24 @@ impl<NS: NodeSegmentOps<Extension = EXT>, EXT: PersistenceStrategy<NS = NS>>
...
-                            } else {
-                                // There is another free segment in the slot, retry using it.
-                                segment_id = *slot;
-                            }
-                        }
-                        Some(local_pos) => return (LocalPOS(local_pos), writer),
-                    }
-                }
+        // No point in multiple threads getting past here as they would just content on the writer lock
+        let mut slot = self.free_segments[slot_idx].write();
```
fabianmurariu, line 307: should then make free_segments Mutex ?

fabianmurariu (changes requested):
LGTM one small comment

PR #2541, opened 2026-03-26
Switch to using byte offsets as LSNs

fabubaker:
Log Sequence Numbers used in the WAL are currently simple sequential counters.
By switching to using byte offsets into the underlying WAL file, we can allow for simpler and faster recovery.

Currently, we always cleanup the WAL file on a clean shutdown, so that there is no need to parse through the full WAL file on reboot. This is unnecessary as WAL files only need to be removed when they reach a particular size on disk.

This PR changes clean shutdowns to record the latest LSN in the WAL stream to disk and seek to that position on reboot. This allows for faster reboot without needing to always remove the WAL file on clean shutdowns.

No

A mix of unit tests and proptests, most of which is in https://github.com/Pometry/pometry-storage/pull/223.

We will need some sort of background thread that goes in and cleans up outdated WAL files.


db4-storage/src/pages/mod.rs
```diff
@@ -350,32 +355,29 @@ impl<
...
-
-                // Log a checkpoint record so we can restore the next LSN after reload.
+                // Log a checkpoint record in the WAL, indicating that the DB was shutdown
+                // with all the segments flushed to disk.
+                // On startup, recovery is skipped since there are no pending writes to replay.
                 let checkpoint_lsn = wal
-                    .log_checkpoint(latest_lsn_on_disk)
-                    .expect("Failed to log checkpoint in drop");
+                    .log_shutdown_checkpoint()
+                    .expect("Failed to log shutdown checkpoint in drop");
```
ljeub-pometry: no panics in drop, this should just be logged

ljeub-pometry (approved):
Probably shouldn't panic in drop

db4-storage/src/pages/mod.rs
fabubaker: Removed the `expect` calls; errors are logged and the Drop returns immediately .

PR #2548, opened 2026-04-08
Fix deep_clone on DictMapper not cloning the underlying map

arienandalibi:
Previously, deep_clone on DictMapper would only clone the Arc holding the map, not the map itself. This led to problems involving correctness and deadlocks.

### DIFF
--- a/raphtory-api/src/core/storage/dict_mapper.rs
+++ b/raphtory-api/src/core/storage/dict_mapper.rs
@@ -330,6 +330,93 @@ impl DictMapper {
     }
 }
 
+#[derive(Debug)]
+pub struct AllKeys<T> {
+    pub(crate) guard: ArcRwLockReadGuard<Vec<T>>,
+}
+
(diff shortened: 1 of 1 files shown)


raphtory-api/src/core/storage/dict_mapper.rs
```diff
@@ -209,10 +209,11 @@ impl DictMapper {
     }
 
     pub fn deep_clone(&self) -> Self {
+        let map = self.map.read_recursive().clone();
         let reverse_map = self.read_lock_reverse_map().clone();
 
         Self {
-            map: self.map.clone(),
+            map: Arc::new(RwLock::new(map)),
```
fabubaker: Can you add a unit test that reproduces the bug below? So create a mapper, add some entries, call `deep_clone` to create a new mapper, add more entries, check keys/values are sane across both.

fabubaker (changes requested):

fabubaker (approved):

PR #2549, opened 2026-04-08
Change materialize to use the parquet encoders and df_loaders

arienandalibi:
Instead of materialize having its own custom ingestion implementation, we change it to leverage the parquet encoders/serialization pathway which generate RecordBatches, and then ingest these using the df_loaders pathway.

Reduce complexity by removing an unused ingestion pathway, and promote future maintainability.

It shouldn't

Tested using the SNB SF1 and SF3 datasets and compared new materialize implementation with the old one.

There shouldn't be


python/src/lib.rs
```diff
@@ -10,6 +10,12 @@ use raphtory::python::{
 };
 use raphtory_graphql::python::pymodule::base_graphql_module;
 
+#[cfg(not(target_env = "msvc"))]
```
fabianmurariu: I don't want Jemalloc on Linux and it doesn't build on Windows so I think the `target_os="macos"` is what we want

python/Cargo.toml
```diff
@@ -38,5 +38,5 @@ proto = ["raphtory/proto"]
 [build-dependencies]
 pyo3-build-config = { workspace = true }
 
-#[target.'cfg(not(target_env = "msvc"))'.dependencies]
-#tikv-jemallocator.workspace = true
+[target.'cfg(not(target_env = "msvc"))'.dependencies]
```
fabianmurariu: I don't want Jemalloc on Linux and it doesn't build on Windows so I think `the target_os="macos"` is what we want

raphtory/src/db/api/view/graph.rs
```diff
@@ -243,6 +265,424 @@ fn edges_inner<'graph, G: GraphView + 'graph>(g: &G, locked: bool) -> Edges<'gra
...
+    let stream_capacity = 10;
+    let (tx, rx) = crossbeam_channel::bounded::<RecordBatchMessage>(stream_capacity);
+
+    if let Some(max_eid) = graph.edges().iter().map(|edge| edge.edge.pid().0).max() {
+        let mut write_locked_graph = materialized.write_lock()?;
+        write_locked_graph.resize_segments_to_eid(EID(max_eid));
+    }
+
+    let mut scope_result = Ok(());
+    std::thread::scope(|scope| {
```
fabianmurariu: rayon scope not good?

raphtory/src/io/mod.rs
```diff
@@ -9,9 +9,16 @@ pub mod neo4j_loader;
...
 
-static LOAD_POOL: LazyLock<ThreadPool> = LazyLock::new(|| {
+pub(crate) static LOAD_POOL: LazyLock<ThreadPool> = LazyLock::new(|| {
     ThreadPoolBuilder::new()
         .thread_name(|idx| format!("PS Bulk Load Thread-{idx}"))
         .build()
         .unwrap()
 });
+
+pub(crate) static ENCODE_POOL: LazyLock<ThreadPool> = LazyLock::new(|| {
```
fabianmurariu: I don't think we need a separate thread pool

raphtory/src/db/api/view/graph.rs
fabianmurariu: You can probably use LOAD_POOL.install all the way up here too

raphtory/src/parquet/edges.rs
fabianmurariu, line 1: There are a few unused warnings

fabianmurariu: Seems the other files in `parquet` also do

raphtory/tests/test_materialize.rs
```diff
@@ -1,14 +1,101 @@
...
-use std::ops::Range;
+use std::{
+    fs, io,
+    ops::Range,
+    path::{Path, PathBuf},
+    time::{Duration, Instant},
+};
+
+#[cfg(feature = "io")]
+fn default_sf10_graph_path() -> PathBuf {
```
fabianmurariu: maybe we should move the snb materialize stuff in it's own file

fabianmurariu (changes requested):
LGTM, but it needs a few changes, nothing major

raphtory/src/io/mod.rs
arienandalibi: Since the channel has a size limit, I noticed that all threads in the thread pool were blocking in the encoder waiting for the channel to have room, but this never happened because all threads were blocked so no RecordBatch could be ingested. I fixed this by having two separate thread pools, but there might be a better solution here.

raphtory/src/db/api/view/graph.rs
arienandalibi: Keeping this unresolved in case we want to use one thread pool everywhere

ljeub-pometry (approved):

PR #2553, opened 2026-04-14
Db v4 merge master

ljeub-pometry:

merge master into db_v4

** DO NOT SQUASH! **


python/tests/test_base_install/test_graphdb/test_algorithms.py
```diff
@@ -37,13 +35,15 @@ def test_local_clustering_coefficient():
...
+
     actual = {
         str(i): algorithms.local_clustering_coefficient(g, g.node(i))
         for i in range(1, 9)
     }
     assert actual == expected
+    expected = {k: {"lcc": v} for k, v in expected.items()}
     actual = algorithms.local_clustering_coefficient_batch(g, list(range(1, 9)))
-    actual = {str(i): actual[i] for i in range(1, 9)}
+    # actual = {str(i): actual[i]["lcc"] for i in range(1, 9)}
```
fabianmurariu: We don't want to test this?

raphtory-graphql/src/model/plugins/algorithms.rs
```diff
@@ -23,6 +24,12 @@ pub(crate) struct PagerankOutput {
     rank: f64,
 }
 
+impl From<(String, HashMap<String, Option<Prop>>)> for PagerankOutput {
+    fn from(_value: (String, HashMap<String, Option<Prop>>)) -> Self {
+        todo!()
```
fabianmurariu: This seems fishy 🐟 🐟 🐟

raphtory/src/algorithms/centrality/degree_centrality.rs
```diff
@@ -15,18 +25,23 @@ use rayon::prelude::*;
...
     };
 
-    let values: Vec<_> = g
+    let values: Vec<CentralityScore> = g
         .nodes()
         .degree()
         .into_par_iter_values()
-        .map(|v| (v as f64) / max_degree as f64)
+        .map(|v| CentralityScore {
+            score: (v as f64) / max_degree as f64,
```
fabianmurariu: won't this blow up on a graph with no edges?

raphtory/src/db/api/state/node_state_ord_ops.rs
```diff
@@ -53,9 +50,17 @@ impl<T: IsFloat> AsOrd for &Option<T> {
     }
 }
 
-pub trait OrderedNodeStateOps<'graph>: NodeStateOps<'graph>
+impl<T: IsFloat> AsOrd for Option<T> {
+    type Ordered = Option<OrderedFloat<T>>;
+    fn as_ord(&self) -> &Option<OrderedFloat<T>> {
+        // Safety: OrderedFloat is #[repr(transparent)] over T; Option<OrderedFloat<T>> has the same layout as Option<T> for the same T.
+        unsafe { &*(self as *const Option<T> as *const Option<OrderedFloat<T>>) }
```
fabianmurariu, line 57: OrderedFloat has a bytemuck flag, technically cast should do the trick? non essential really

raphtory/src/python/filter/node_filter_builders.rs
```diff
@@ -39,7 +41,7 @@ pub struct PyNodeIdFilterBuilder(Arc<NodeIdFilterBuilder>);
 impl PyNodeIdFilterBuilder {
     /// Returns a filter expression that checks whether the node ID
     /// is equal to the given value.
-    ///
+    ///pub(crate) pub(crate)
```
fabianmurariu: should not be here

fabianmurariu (changes requested):

python/tests/test_base_install/test_graphdb/test_algorithms.py
ljeub-pometry: This conversion isn't needed, the equality check is implemented for the output node state.

raphtory-graphql/src/model/plugins/algorithms.rs
ljeub-pometry: It's not used :)

raphtory/src/algorithms/centrality/degree_centrality.rs
ljeub-pometry: good point

raphtory/src/db/api/state/node_state_ord_ops.rs
ljeub-pometry: This doesn't work with bytemuck. The `Option` has padding which doesn't matter here as we are only casting to another `Option` but means it is not `POD`.

PR #2554, opened 2026-04-16
Add personalization to pagerank

phdavis1027:
Two tests with expected values computed via NetworkX:
- test_personalized_page_rank: all teleport weight on node 1
- test_personalized_page_rank_partial: weight split between nodes 1 and 3

Add a personalization parameter to `unweighted_page_rank`

If `personalization` is None, then we dispatch via trait to the exact same calculations as the previous, uniform variant to avoid rounding errors from non-commutative f64 arithmetic. 

To compute personalized page rank (e.g., in fraud-detection situations)

Yes, it adds the `personalization` parameter to `unweighted_pagerank`. This is added to docstrings in both `python` and `algorithms/centrality`

The unit tests added were validated by computing the same values using networkx.

I am quite happy to also add a weighted variant

### DIFF
--- a/examples/rust/src/bin/bench/main.rs
+++ b/examples/rust/src/bin/bench/main.rs
@@ -82,6 +82,6 @@ fn main() {
         g
     };
     info!("Data loaded\nPageRanking");
-    page_rank(&graph, None, Some(25), Some(8), None, true, None);
+    page_rank(&graph, None, Some(25), Some(8), None, true, None, None);
     info!("Done PR");
 }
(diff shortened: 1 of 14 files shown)


phdavis1027:
So the failure says that the global personalization state is empty, but I confirmed under LLDB that it is present when ss = 0. The first failing reading happens at ss = 1, seemingly because the VecArray double-buffer didn't get its buffers swapped, so we read the empty one.  It looks kinda like `ShuffleComputeState::copy_over_next_ss` just doesn't copy global state?

PR #2555, opened 2026-04-16
Add optional weight to pagerank

phdavis1027:

Add an optional weight parameter to pagerank

To capture the semantics of weighted pagerank. For example, in fraud detection

Yes, it adds an extra param which is captured in the docstrings

Test scenarios were checked for exact match with networkx. Additionally, basic semantic sanity checking like 'weight =None' should be treated as equivalent to all weights = 1.0

No

### DIFF
--- a/raphtory/src/algorithms/centrality/pagerank.rs
+++ b/raphtory/src/algorithms/centrality/pagerank.rs
@@ -3,7 +3,7 @@ use crate::{
     db::{
         api::{
             state::NodeState,
-            view::{NodeViewOps, StaticGraphViewOps},
+            view::{EdgeViewOps, NodeViewOps, StaticGraphViewOps},
         },
         task::{
             context::Context,
@@ -11,21 +11,22 @@ use crate::{
             task_runner::TaskRunner,
         },
     },
-    prelude::GraphViewOps,
+    prelude::{GraphViewOps, PropertiesOps},
 };
 use num_traits::abs;
+use raphtory_api::core::entities::properties::prop::PropUnwrap;
 
 #[derive(Clone, Debug, Default)]
 struct PageRankState {
     score: f64,
-    out_degree: usize,
+    weighted_out_degree: f64,
 }
 
 impl PageRankState {
     fn new(num_nodes: usize) -> Self {
         Self {
             score: 1f64 / num_nodes as f64,
-            out_degree: 0,
+            weighted_out_degree: 0f64,
         }
     }
 
@@ -40,6 +41,7 @@ impl PageRankState {
 /// # Arguments
 ///
 /// - `g`: A GraphView object
+/// - `weight`: Edge property key to use as weight. If None, all edges have weight 1.0.
 /// - `iter_count`: Number of iterations to run the algorithm for
 /// - `threads`: Number of threads to use for parallel execution
 /// - `tol`: The tolerance value for convergence
@@ -50,8 +52,9 @@ impl PageRankState {
 ///
 /// An [AlgorithmResult] object containing the mapping from node ID to the PageRank score of the node
 ///
-pub fn unweighted_page_rank<G: StaticGraphViewOps>(
+pub fn page_rank<G: StaticGraphViewOps>(
     g: &G,
+    weight: Option<&str>,
     iter_count: Option<usize>,
     threads: Option<usize>,
     tol: Option<f64>,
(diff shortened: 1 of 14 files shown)


raphtory/src/algorithms/centrality/pagerank.rs
```diff
@@ -76,38 +79,53 @@ pub fn unweighted_page_rank<G: StaticGraphViewOps>(
...
-        Step::Continue
+    let weight_key: Option<String> = weight.map(|s| s.to_string());
+
+    let step1 = ATask::new({
+        let weight_key = weight_key.clone();
+        move |s| {
+            let weighted_out_degree = s.out_edges().iter().fold(0.0f64, |acc, edge| {
+                weight_key
+                    .as_ref()
+                    .and_then(|key| edge.properties().get(key))
```
fabianmurariu: this should be get_by_id as it's a lot quicker than looking up via the string every time.

raphtory/src/algorithms/centrality/pagerank.rs
```diff
@@ -76,38 +79,53 @@ pub fn unweighted_page_rank<G: StaticGraphViewOps>(
...
         }
 
-        for t in s.in_neighbours() {
-            let prev = t.prev();
-
-            s.get_mut().score += prev.score / prev.out_degree as f64;
+        for edge in s.in_edges() {
+            let w = weight_key
+                .as_ref()
+                .and_then(|key| edge.properties().get(key))
```
fabianmurariu: same here use `get_by_id`

fabianmurariu (changes requested):
looking up the property every time by string is costly, we would look it up once via the graph node metadata, then calling `get_by_id`

raphtory/src/algorithms/centrality/pagerank.rs
phdavis1027: Done

raphtory/src/algorithms/centrality/pagerank.rs
phdavis1027: Done

raphtory/src/algorithms/centrality/pagerank.rs
```diff
@@ -107,7 +125,7 @@ pub fn unweighted_page_rank<G: StaticGraphViewOps>(
     let step3 = ATask::new(move |s| {
         let state: &mut PageRankState = s.get_mut();
 
-        if state.out_degree == 0 {
+        if state.weighted_out_degree == 0.0 {
```
phdavis1027: Perhaps we want some tolerance here instead of raw == ?

phdavis1027 (commented):
- applied feedback by looking up weight property by index instead of string
- removed some random formatting changes that muddied the diff
- noticed floating point ==, which seems smelly
- rust tests pass, i forget how i made the python side compile without clam though
- some conflicts with current `db_v4` that i don't have time to iron out right now but i can get on it as soon as this looks otherwise ready to ship

phdavis1027:
conflicts resolved to the best of my ability, plus switched to delta check rather than raw f64 equality

fabianmurariu (approved):
LGTM

wyatt-joyner-pometry:
This looks good to me! Iterating over every edge between every pair of vertices every iteration might suffer on graphs with lots of transactions, though. An optimization to address this would be to cache the cumulative pairwise weights for vertices with a sufficiently large number of edges between them.

phdavis1027:
> This looks good to me! Iterating over every edge between every pair of vertices every iteration might suffer on graphs with lots of transactions, though. An optimization to address this would be to cache the cumulative pairwise weights for vertices with a sufficiently large number of edges between them.

By 'transactions' do you mean number of rows after e.g., exploding the DataFrame? When I first wrote this PR, I wasn't quite aware of the exact shape of this API, but have encountered it while testing for internal application. In any case, it seems to me that pagerank just sees the state of each edge after the latest update?

wyatt-joyner-pometry:
> > This looks good to me! Iterating over every edge between every pair of vertices every iteration might suffer on graphs with lots of transactions, though. An optimization to address this would be to cache the cumulative pairwise weights for vertices with a sufficiently large number of edges between them.
> 
> By 'transactions' do you mean number of rows after e.g., exploding the DataFrame? When I first wrote this PR, I wasn't quite aware of the exact shape of this API, but have encountered it while testing for internal application. In any case, it seems to me that pagerank just sees the state of each edge after the latest update?

Oh yep you're right! I was thinking of iterating over an exploded edge, which is what you want to do if you're trying to get the weighted pagerank in a transaction network (rather than just fetching latest transaction).

phdavis1027:
> Oh yep you're right! I was thinking of iterating over an exploded edge, which is what you want to do if you're trying to get the weighted pagerank in a transaction network (rather than just fetching latest transaction).

Well, in fact, now that I revisit this PR in the light of some more hands-on experience, that is the use-case I had in mind. Am I reading you correctly that there is an API for pagerank over the graph induced by the exploded edge set? If not, potentially a follow-up PR?

ljeub-pometry:
> > Oh yep you're right! I was thinking of iterating over an exploded edge, which is what you want to do if you're trying to get the weighted pagerank in a transaction network (rather than just fetching latest transaction).
> 
> Well, in fact, now that I revisit this PR in the light of some more hands-on experience, that is the use-case I had in mind. Am I reading you correctly that there is an API for pagerank over the graph induced by the exploded edge set? If not, potentially a follow-up PR?

This would probably be a follow-up PR as this version is very useful in a lot of contexts (mostly those that use the `PersistentGraph` semantics). If you can fix the failing python tests (these are returning dicts at the moment in the v4 apis) we can merge this PR.

phdavis1027:
Hopefully caught all the weirdness introduced by unintentional format-on-save. The python tests that CI tripped over pass. Gotta sleep for tonight but I'm running `make test-all` overnight, so hopefully I get to any lurking failures before someone pushes the CI button

EDIT: Got the congratulations message

phdavis1027:
Seems the benchmark workflow is still misconfigured for external contribution? Also, not really sure how this could have regressed so many benchmarks...

ljeub-pometry:
> Seems the benchmark workflow is still misconfigured for external contribution? Also, not really sure how this could have regressed so many benchmarks...

The benchmarks are still comparing against the old master and there are still some performance regressions in the v4 branch that we need to sort out. This is expected and has nothing to do with this PR. It seems the functionality to post a PR comment is failing for external PRs, need to look into that.

PR #2584, opened 2026-04-21
Add fast path for last on properties

ljeub-pometry:


fabianmurariu:
replaced by https://github.com/Pometry/Raphtory/pull/2593

PR #2586, opened 2026-04-23
Optimisations for node validity and neighbour filtering

ljeub-pometry:

Improve the implementations of node_valid and use src/dst from edge ref for filtering when available

Performance

No


fabianmurariu (approved):
LGTM

PR #2588, opened 2026-04-24
Change temporal values from strings to prop values and add aggregates

### LINKED ISSUE #2102: Need to add aggregates to graphql propery APIs

miratepuffin:
### GraphQL: typed values for temporal properties (breaking)

`TemporalProperty` fields that used to return stringified values now return properly typed `PropertyOutput` scalars (the same one already used by `Property.value`), so numbers come back as numbers, booleans as booleans, etc.:

| Field | Before | After |
|---|---|---|
| `values` | `[String!]!` | `[PropertyOutput!]!` |
| `latest` | `String` | `PropertyOutput` |
| `at(t)` | `String` | `PropertyOutput` |
| `unique` | `[String!]!` | `[PropertyOutput!]!` |

No other fields (including the list-of-`Property` shape of `Properties.values` / `Metadata.values`) are changed.

### GraphQL: new aggregation fields on `TemporalProperty`

Parity with the Python `TemporalProperty` aggregations, exposed as fields on each temporal property (so they compose with `window(...)` / `layer(...)` / etc.):

| Field | Type | Notes |
|---|---|---|
| `sum` | `PropertyOutput` | null if dtype isn't additive; strings concatenate in temporal order |
| `mean` / `average` | `PropertyOutput` | always `F64`; null if any value can't be converted to f64 |
| `count` | `Int` | number of updates |
| `min` | `PropertyTuple` | `(time, value)` pair; null if dtype isn't comparable |
| `max` | `PropertyTuple` | `(time, value)` pair; null if dtype isn't comparable |
| `median` | `PropertyTuple` | lower median on even-length inputs |

### Rust core: aggregations promoted to `TemporalPropertyView`

`sum` / `min` / `max` / `count` / `mean` / `average` / `median` previously only existed on the Python wrapper (`PyTemporalProp`) via private helpers. They've been moved onto the core `TemporalPropertyView` in `raphtory::db::api::properties`, so pure-Rust consumers get them directly, and Python + GraphQL now share the one implementation.

### Tests

New `python/tests/test_base_install/test_graphql/test_gql_temporal_aggregates.py`:
- Numeric node property — all aggregations on full timeline, windowed view, and empty-window edge case.
- Float edge property on two layers — all aggregations under full view, per-layer view, and combined `layer + window`.
- Non-numeric semantics — strings (sum concatenates, mean null, min/max/median lexicographic) and bools (sum/mean null, min/max/median work with `False < True`).

Fixes #2102

### DIFF
--- a/raphtory/src/db/api/properties/temporal_props.rs
+++ b/raphtory/src/db/api/properties/temporal_props.rs
@@ -115,6 +115,79 @@ impl<P: InternalPropertiesOps + Clone> TemporalPropertyView<P> {
         unique_props.into_iter().collect()
     }
 
+    /// Compute the sum of all property values, or `None` if the dtype is not additive or the
+    /// property is empty.
+    pub fn sum(&self) -> Option<Prop> {
+        generalised_reduce(self.values(), |a, b| a.add(b), |p| p.dtype().has_add())
+    }
+
+    /// Find the minimum `(time, value)` pair, or `None` if the dtype is not comparable or the
+    /// property is empty.
+    pub fn min(&self) -> Option<(EventTime, Prop)> {
+        generalised_reduce(
+            self.iter(),
+            |a, b| {
+                if a.1.partial_cmp(&b.1)?.is_le() {
+                    Some(a)
+                } else {
+                    Some(b)
+                }
+            },
+            |(_, v)| v.dtype().has_cmp(),
+        )
+    }
+
+    /// Find the maximum `(time, value)` pair, or `None` if the dtype is not comparable or the
+    /// property is empty.
+    pub fn max(&self) -> Option<(EventTime, Prop)> {
+        generalised_reduce(
+            self.iter(),
+            |a, b| {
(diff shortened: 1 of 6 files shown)


raphtory/src/db/api/properties/temporal_props.rs
```diff
@@ -115,6 +115,79 @@ impl<P: InternalPropertiesOps + Clone> TemporalPropertyView<P> {
...
+            sum += value.as_f64()?;
+            count += 1;
+        }
+        Some(Prop::F64(sum / count as f64))
+    }
+
+    /// Alias for `mean`.
+    pub fn average(&self) -> Option<Prop> {
+        self.mean()
+    }
```
ljeub-pometry, line 177: do we need the alias?

ljeub-pometry (approved):

raphtory/src/db/api/properties/temporal_props.rs
miratepuffin: No, but it has been there for ages and people might be using it

Issue #2302, opened 2025-09-30
Stubsgen: identify python types of filter operators


jbaross-pometry:
For operators a lot of the python return types are unclear, this includes many of the entries that were previously 'value'

```
ERROR:raphtory.filter.PropertyFilterOps.contains:Invalid return type 'FilterExpr': Unknown type FilterExpr
WARNING:raphtory.filter.PropertyFilterOps.contains:missing parameter value in docs.
ERROR:raphtory.filter.PropertyFilterOps.contains:parameter PropValue appears in documentation but does not exist.
ERROR:raphtory.filter.PropertyFilterOps.fuzzy_search:Invalid annotation 'usize' for parameter levenshtein_distance: Unknown type usize
ERROR:raphtory.filter.PropertyFilterOps.fuzzy_search:Invalid return type 'FilterExpr': Unknown type FilterExpr
ERROR:raphtory.filter.PropertyFilterOps.is_in:Invalid return type 'FilterExpr': Unknown type FilterExpr
WARNING:raphtory.filter.PropertyFilterOps.is_in:missing parameter values in docs.
ERROR:raphtory.filter.PropertyFilterOps.is_in:parameter list[PropValue] appears in documentation but does not exist.
ERROR:raphtory.filter.PropertyFilterOps.is_none:Invalid return type 'FilterExpr': Unknown type FilterExpr
ERROR:raphtory.filter.PropertyFilterOps.is_not_in:Invalid return type 'FilterExpr': Unknown type FilterExpr
WARNING:raphtory.filter.PropertyFilterOps.is_not_in:missing parameter values in docs.
ERROR:raphtory.filter.PropertyFilterOps.is_not_in:parameter list[PropValue] appears in documentation but does not exist.
ERROR:raphtory.filter.PropertyFilterOps.is_some:Invalid return type 'FilterExpr': Unknown type FilterExpr
ERROR:raphtory.filter.PropertyFilterOps.not_contains:Invalid return type 'FilterExpr': Unknown type FilterExpr
WARNING:raphtory.filter.PropertyFilterOps.not_contains:missing parameter value in docs.
ERROR:raphtory.filter.PropertyFilterOps.not_contains:parameter PropValue appears in documentation but does not exist.
WARNING:raphtory.filter.NodeFilterBuilder.contains:Missing return type annotation
WARNING:raphtory.filter.NodeFilterBuilder.contains:missing parameter value in docs.
[...]

Issue #2301, opened 2025-09-30
stubsgen: Newly exposed itterables need documenting


jbaross-pometry:
need to idetify where to add comments so that docs for the following are picked up by stubsgen

```
WARNING:raphtory.iterables.NestedUtcDateTimeIterable.collect:Missing documentation
WARNING:raphtory.iterables.NestedGIDIterable.collect:Missing documentation
WARNING:raphtory.iterables.NestedGIDIterable.max:Missing documentation
WARNING:raphtory.iterables.NestedGIDIterable.min:Missing documentation
WARNING:raphtory.iterables.GIDIterable.collect:Missing documentation
WARNING:raphtory.iterables.GIDIterable.max:Missing documentation
WARNING:raphtory.iterables.GIDIterable.min:Missing documentation
WARNING:raphtory.iterables.StringIterable.collect:Missing documentation
WARNING:raphtory.iterables.OptionArcStringIterable.collect:Missing documentation
WARNING:raphtory.iterables.UsizeIterable.collect:Missing documentation
WARNING:raphtory.iterables.UsizeIterable.max:Missing documentation
WARNING:raphtory.iterables.UsizeIterable.mean:Missing documentation
WARNING:raphtory.iterables.UsizeIterable.min:Missing documentation
WARNING:raphtory.iterables.UsizeIterable.sum:Missing documentation
WARNING:raphtory.iterables.OptionI64Iterable.collect:Missing documentation
WARNING:raphtory.iterables.OptionI64Iterable.max:Missing documentation
WARNING:raphtory.iterables.OptionI64Iterable.min:Missing documentation
WARNING:raphtory.iterables.NestedOptionArcStringIterable.collect:Missing documentation
WARNING:raphtory.iterables.NestedStringIterable.collect:Missing documentation
WARNING:raphtory.iterables.NestedOptionI64Iterable.collect:Missing documentation
WARNING:raphtory.iterables.NestedOptionI64Iterable.max:Missing documentation
WARNING:raphtory.iterables.NestedOptionI64Iterable.min:Missing documentation
WARNING:raphtory.iterables.NestedI64VecIterable.collect:Missing documentation
WARNING:raphtory.iterables.NestedUsizeIterable.collect:Missing documentation
WARNING:raphtory.iterables.NestedUsizeIterable.max:Missing documentation
WARNING:raphtory.iterables.NestedUsizeIterable.mean:Missing documentation
[...]

Issue #2300, opened 2025-09-30
Stubsgen: handle python types for unresolved classes


jbaross-pometry:
Follows on from https://github.com/Pometry/Raphtory/pull/2273

For the stubsgen errors as follows we need to:
- Expose types to python
- Identify infered type in pthon
- Handle context dependent outputs

```
WARNING:raphtory.Graph.create_index_in_ram_with_spec:missing parameter py_spec in docs.
ERROR:raphtory.Graph.create_index_in_ram_with_spec:parameter IndexSpec appears in documentation but does not exist.
ERROR:raphtory.PersistentGraph.create_index_in_ram_with_spec:parameter IndexSpec appears in documentation but does not exist.
ERROR:raphtory.Nodes.metadata:Invalid return type 'MetadataView': Unknown type MetadataView
ERROR:raphtory.PathFromNode.metadata:Invalid return type 'MetadataView': Unknown type MetadataView
ERROR:raphtory.PathFromGraph.metadata:Invalid return type 'MetadataListList': Unknown type MetadataListList
ERROR:raphtory.PathFromGraph.properties:Invalid return type 'NestedPropsIterable': Unknown type NestedPropsIterable
ERROR:raphtory.Edge.explode:Invalid return type 'Exploded': Unknown type Exploded
ERROR:raphtory.Edge.explode_layers:Invalid return type 'Exploded': Unknown type Exploded
ERROR:raphtory.Edges.deletions:Invalid return type 'PyGenericIterable': Unknown type PyGenericIterable
ERROR:raphtory.Edges.explode:Invalid return type 'Exploded': Unknown type Exploded
ERROR:raphtory.Edges.explode_layers:Invalid return type 'Exploded': Unknown type Exploded
ERROR:raphtory.Edges.history:Invalid return type 'PyGenericIterable': Unknown type PyGenericIterable
ERROR:raphtory.Edges.metadata:Invalid return type 'MetadataView': Unknown type MetadataView
ERROR:raphtory.Edges.time:Invalid return type 'I64Iterable': Unknown type I64Iterable
ERROR:raphtory.NestedEdges.explode:Invalid return type 'Exploded': Unknown type Exploded
ERROR:raphtory.NestedEdges.explode_layers:Invalid return type 'Exploded': Unknown type Exploded
ERROR:raphtory.NestedEdges.layer_name:Invalid return type 'NestedArcStringIterable': Unknown type NestedArcStringIterable
[...]

Issue #2238, opened 2025-08-15
If I have a numerically id'd graph I can't actually update from graphql


miratepuffin:
<img width="1217" height="367" alt="Image" src="https://github.com/user-attachments/assets/2e594664-e0af-4475-800f-9ed886059d7c" />

PR #2589, opened 2026-04-24
Graphql docs and tests

### LINKED ISSUE #1937: Add secondary index to graphql update apis

miratepuffin:
# `graphql-docs-and-tests` — Top-level Summary

Branch focused on hardening Raphtory's GraphQL surface: documenting it, expanding what its inputs accept, and verifying it from Python. Also picks up housekeeping on the Python bindings (PyO3 stubs, PyProp constructors).

Fixes #1937 #2101 #2238 #2302 #2301 #2300

Need to get https://github.com/smmoosavi/dynamic-graphql/pull/42 merged 

## 1. GraphQL schema: documentation + tests

- Every resolver type, field, and argument now carries a doc comment intended for the public docs site.
- Permission requirements on every root-query / mutation now follow a consistent format (`Requires READ on the graph.`, `Requires WRITE on the parent namespace.`, etc.).
- Filled in coverage for previously untested resolvers; added Python integration tests around them.
- Picked up a forked `dynamic-graphql` branch that exposes per-argument `desc` attributes (the released crate has no such hook).

## 2. New / widened input scalars

Made a new `NodeId` for `String` **or** non-negative `Int` similar to TimeInput. The two scalars are wired through every relevant resolver: lookups (`node`, `edge`, `hasNode`, `hasEdge`), view transforms (`subgraph`, `excludeNodes`, `sharedNeighbours`), all mutations (`addNode`, `addEdge`, `deleteEdge`, batch variants, `MutableGraph.{node,edge}`), `createSubgraph`, and `VectorSelection.{addNodes,addEdges}`.

`Node.id` and `Edge.id` now return `NodeId` (so an integer-indexed graph returns integers, not stringified ones).

Filters already accepted both forms via the existing `Value` scalar — no schema change there, just verified.

## 3. Graph-semantics override

`graph(path: ..., graphType: EVENT | PERSISTENT)` (and the same arg on `MutableGraph.graph`) lets a caller re-interpret a stored graph at query time:

- An event graph queried with `graphType: PERSISTENT` carries values forward in windows after the addition.
- A persistent graph queried with `graphType: EVENT` only sees updates exactly at the queried time.

## 4. Temporal-property API changes

- `TemporalProperty.values` now returns typed `PropValue`s instead of stringified values.
- Added Python-style aggregates on `TemporalProperty`: `sum`, `mean`, `median`, `min`, `max`, `count`, `average`. Returning `null` for non-numeric value types.
- `unique` and `orderedDedupe` return a `TemporalProperty` so the aggregates compose.

## 5. Python tests added

| File | Coverage |
|---|---|
| `test_gql_node_id.py` | `NodeId` (Int + String) on every read/write surface |
| `test_gql_mutation_time_input.py` | `TimeInput` shapes (Int / RFC3339 / object) on every mutation |
| `test_gql_graph_type.py` | `graphType` override × stored type × window semantics |
| `test_graphql_vectors.py` | `VectorisedGraphWindow` with `TimeInput` |
| `test_props.py` | New `Prop` constructors (datetime, naive_datetime, decimal) — strings, Python `Decimal`, ints, floats; graph round-trip; list unification |
| `test_load_from_pandas.py` | Schema-driven loading with `PropType.decimal(scale)`, `PropType.datetime()`, `PropType.naive_datetime()` — string columns, Arrow `Decimal128`, both nodes and edges |

### New `Prop` / `PropType` constructors
- `Prop.aware_datetime(dt)` — tz-aware datetime (UTC-stored). Naive input rejected.
- `Prop.naive_datetime(dt)` — tz-naive datetime.
- `Prop.decimal(value)` — accepts `decimal.Decimal`, `str`, `int`, or `float`.
- `PropType.decimal(scale)` — schema-side companion (`PropType.datetime()` / `PropType.naive_datetime()` already existed). Lets `load_nodes`/`load_edges` accept `schema={"col": PropType.decimal(2)}` to cast string or Arrow `Decimal128` columns.

### Stub generation (`gen-stubs.py`)
- Audited every PyO3 method for missing docs / return types / parameter docs.
- Registered iterable wrapper classes (`PyPropValueListList`, `PyTemporalPropsList`, `PyTemporalPropsListList`, `PyPropHistValueList`, `PyPropHistValueListList`, `PyTemporalPropList`, `PyTemporalPropListList`, `PyPropHistItemsList`, `PyPropHistItemsListList`, `PropIterable`) plus `EmbeddingServer`, `RunningEmbeddingServer`, `OutputNodeState` so they resolve in generated stubs.
- Aliased `raphtory.iterables` and `raphtory.gql` into `sys.modules` so submodule imports resolve.
- Renamed Rust types in docstrings to match exported Python names (`PyOutputNodeState` → `OutputNodeState`, `TypedNodeState` → `OutputNodeState`, `Str` → `str`, `NumpyArray` → `NDArray`, etc.).
- Result: `gen-stubs.py` reports zero warnings; `mypy -m raphtory` passes. (other than GQL which we can fix later)

## 7. dynamic-graphql fork

Submitted upstream PR adding `desc` attribute support on `ResolvedObjectFieldsArg` / `InterfaceMethodArg` / `ExpandObjectFieldsArg`. Branch `add-arg-descriptions` of `miratepuffin/dynamic-graphql` is referenced from this repo's `Cargo.toml` until upstream merges.


python/tests/test_base_install/test_graphql/test_gql_mutation_time_input.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+def test_add_node_accepts_int_string_and_object_time():
+    """`addNode` accepts every `TimeInput` shape — Int, RFC3339 string, and
+    `{timestamp, eventId}` object. Each insertion lands at its expected
+    timestamp and is queryable afterwards."""
+    work_dir = tempfile.mkdtemp()
+    with GraphServer(work_dir, create_index=True).start(PORT) as server:
+        client = server.get_client()
+        client.send_graph(path="g", graph=Graph())
```
ljeub-pometry: this is a weird way to make a new empty graph

python/tests/test_base_install/test_graphql/test_gql_mutation_time_input.py
```diff
@@ -0,0 +1,296 @@
...
+                  }
+                }
+              }
+            }
+            """,
+        )
+        score = result["graph"]["properties"]["temporal"]["get"]
+        timestamps = [h["timestamp"] for h in score["history"]["list"]]
+        assert sorted(timestamps) == [100, 200, 300]
+        assert sorted(score["values"]) == [1, 2, 3]
```
ljeub-pometry: probably shouldn't sort the two lists independently, the history should already be sorted! (i.e., this check should always pass without the sorting or we would consider that a bug)

python/tests/test_base_install/test_graphql/test_gql_mutation_time_input.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+def test_add_edges_batch_accepts_time_input_shapes():
+    """`addEdges` is the batch counterpart of `addNodes`. Each per-update
+    `time` field on its `TemporalPropertyInput` entries accepts every
+    `TimeInput` shape — Int, RFC3339 string, and `{timestamp, eventId}` object."""
+    work_dir = tempfile.mkdtemp()
+    with GraphServer(work_dir, create_index=True).start(PORT) as server:
+        client = server.get_client()
+        client.send_graph(path="g", graph=Graph())
```
ljeub-pometry: weird

python/tests/test_base_install/test_graphql/test_gql_mutation_time_input.py
```diff
@@ -0,0 +1,296 @@
...
+                }
+                edge(src: "a", dst: "b") {
+                  properties { temporal { get(key: "w") { values } } }
+                  isValid
+                }
+              }
+            }
+            """,
+        )
+        assert sorted(
```
ljeub-pometry: again, no sorting here, these should always come back in temporal order

python/tests/test_base_install/test_graphql/test_gql_node_edge_surface.py
```diff
@@ -0,0 +1,553 @@
...
+                        "firstUpdate": {"timestamp": 10},
+                        "lastUpdate": {"timestamp": 50},
+                        "edgeHistoryCount": 7,
+                    }
+                }
+            },
+        )
+    )
+
+    # Windowed [15, 40): first event for A is at 15 (A->C), last is at 30 (A->B).
```
ljeub-pometry, line 77: would be good to also add a test where the firstUpdate is not the start of the window

python/tests/test_base_install/test_props.py
```diff
@@ -158,3 +159,127 @@ def test_map_with_nested_list_that_is_heterogeneous_rejected():
...
+
+
+def test_aware_datetime():
+    dt = datetime(2024, 6, 1, 12, 30, 45, tzinfo=timezone.utc)
+    p = Prop.aware_datetime(dt)
+    assert str(p.dtype()) == "DTime"
+    assert "2024-06-01" in repr(p)
+
+
+def test_aware_datetime_rejects_naive_input():
```
ljeub-pometry: Is this the correct behaviour? We treat naive date times as implicitly UTC in other places I think

ljeub-pometry (approved):
some minor test things to fix

ljeub-pometry (approved):

Cargo.toml
```diff
@@ -72,7 +72,7 @@ optd-core = { version = "0.18.0", path = "optd/optd/core" }
 async-graphql = { version = "7.2.1", features = ["dynamic-schema"] }
 bincode = { version = "2", features = ["serde"] }
 async-graphql-poem = "7.2.1"
-dynamic-graphql = "0.10.1"
+dynamic-graphql = { git = "https://github.com/miratepuffin/dynamic-graphql", branch = "add-arg-descriptions" }
```
smmoosavi, line 75: dynamic-graphql v1.10.2 is published. I think it should changed

@miratepuffin
@ljeub-pometry 

```suggestion
dynamic-graphql = "0.10.2"
```

miratepuffin: Thank you @smmoosavi we shall swap over!

PR #2594, opened 2026-05-01
Add read only version of graph to allow python access

miratepuffin:
# Branch notes

1. **Read-only graph (disk-backed)** — separate Python processes can
   attach to a live server's graph directory
2. **Read-only graph (in-memory)** — `g.read_only()` takes the segment
   locks and lets reads go fast
3. **Metadata fix in namespace** — `MetaGraph.metadata` no longer forces
   a full graph load and dispatches on the saved format
4. **Explicit `flush` mutation in GraphQL** — server-side knob to make
   in-memory writes visible to read-only handles

---

## 1. Read-only disk-backed graphs

Use case: a GraphQL server holds the writer's lock on a graph directory.
A separate Python process needs to query the same graph **without**
fighting for that lock and without risking writes.

### API

```python
from raphtory import Graph

# In the server process: the normal load takes the writer lock.
g_writer = Graph.load("/data/graphs/social")

# In another process: read_only=True opens without the writer lock.
g_reader = Graph.load("/data/graphs/social", read_only=True)

g_reader.count_nodes()                     # works
g_reader.add_node(t, "x")                  # raises: "graph is locked"
```

Many read-only handles can coexist with the writer. The handle skips
crash recovery (which would mutate the graph and is unsafe with other
attachments).

---

## 2. Read-only in-memory graphs

Use case: you've finished writing to a `Graph()` and want fast reads.
Calling `g.read_only()` acquires per-segment read locks once, so
subsequent reads skip the per-call write-coordination work.

### API

```python
from raphtory import Graph

g = Graph()
g.add_edge(1, "ben", "hamza", {"weight": 1.5})
g.add_node(2, "ben", {"role": "engineer"})

ro = g.read_only()
ro.count_nodes()                          # fast
ro.add_node(3, "lucas")                   # raises: "graph is locked"
```

### Semantics

- The handle shares the underlying `TemporalGraph` — **not a snapshot**.
- All mutations on `ro` raise via the same `Immutable::ReadLockedImmutable`
  path as the disk read-only handle. Coverage proven for the full
  graph/node/edge mutation surface (11 entry points, see tests below).
- **Footgun**: while `ro` is alive, writes on the original `g` block on
  the read locks `ro` holds. Drop `ro` before mutating. A friendlier
  `try_write`-style error is a planned follow-up.

---

### Problem

Listing many graphs in a namespace called `MetaGraph.metadata` per
graph. The original code did `data.get_graph(...)` — a **full graph
load** — just to read constant graph properties. For N graphs, that's N
full deserializations.

Three-tier resolution:

1. **In-memory cache hit**: serve from `data.get_cached_graph(...)`.
2. **Disk-backed graph**: read the `graph_props` segment directly via
   `read_constant_graph_properties` — no full load.
3. **Parquet-backed graph**: read the parquet metadata footer via
   `decode_graph_metadata` — no full load.

---

## 4. Explicit `flush` mutation in GraphQL

Pairs directly with item 1: a server holding the writer can have made
in-memory writes that haven't hit disk yet. A read-only handle opened
elsewhere only sees what's on disk at attach time. Issuing
`{ updateGraph(path: "g") { flush } }` on the server forces the writer
to persist, so the next read-only handle sees the writes.

---


raphtory-graphql/src/model/graph/meta_graph.rs
```diff
@@ -105,26 +106,39 @@ impl MetaGraph {
...
+                .collect());
+        }
+
+        if Extension::disk_storage_enabled() {
+            let graph_path = self
+                .folder
+                .graph_folder()
+                .graph_path()
+                .map_err(GraphError::from)?;
+            let pairs = read_constant_graph_properties(&graph_path).map_err(GraphError::from)?;
```
ljeub-pometry, line 132: should look at the metadata file which can tell you if it is saved as parquet/v4 format and then either call `read_constant_graph_properties` or `decode_graph_metadata` as appropriate

ljeub-pometry (approved):
should fix the metadata for v4 with disk storage loading from parquet

ljeub-pometry:
also, we need to move the tests for the storage feature into the private repo

raphtory/src/db/api/storage/storage.rs
```diff
@@ -211,6 +211,19 @@ impl Storage {
...
+    /// Produce a read-only handle backed by the same `TemporalGraph` as
+    /// this storage. Mutations on the returned `Storage` return
+    /// `Immutable::ReadLockedImmutable`. The handle holds read locks on
+    /// every segment, so writers on the original `Storage` will block
+    /// until the read-only handle is dropped — this is not a snapshot.
+    pub fn read_only(&self) -> Self {
+        Self {
+            graph: self.graph.lock(),
+            #[cfg(feature = "search")]
+            index: RwLock::new(GraphIndex::Empty),
```
ljeub-pometry: this should point at the index on the underlying graph (when we fix the search feature)

ljeub-pometry (approved):

PR #2596, opened 2026-05-05
Add proptest for testing RBAC permissions

fabubaker:


python/tests/test_permissions.py
```diff
@@ -1288,6 +1288,53 @@ def test_child_namespace_restriction_overrides_parent():
         assert "team/restricted/secret" in paths
```
shivamka1, line 1288: `test_permissions` is moved to `pometry_storage`

raphtory-graphql/tests/permissions.rs
```diff
@@ -0,0 +1,204 @@
...
+    const PROPTEST_CASES: u32 = 10;
+    const NAMESPACE_SIZE: RangeInclusive<usize> = 1..=100;
+    const NUM_GRANTS: RangeInclusive<usize> = 1..=100;
+    const NUM_USERS: RangeInclusive<usize> = 1..=20;
+
+    proptest!(
+        ProptestConfig::with_cases(PROPTEST_CASES),
+        |(case in permissions_strategy(NAMESPACE_SIZE, NUM_GRANTS, NUM_USERS))| {
+            let total_graphs = case.graph_paths.len() as u64;
+            let (_server, _tempdir) = start_server(PORT, PUB_KEY, total_graphs);
```
shivamka1, line 145: The server is never explicitly stopped between iterations. `_server` is just dropped, which aborts the background task without sending a shutdown signal. The OS socket is left open, so the next iteration's `start_server` call might fail to bind port `43871` with "address already in use". The abort via drop is fast enough that the OS usually reclaims the socket before the next iteration tries to bind but it's a race condition. Under load, a slow CI machine, or with OS scheduling delays, the socket won't be released in time and the test will fail.

shivamka1: Probably, we reuse the same server for all iterations, cleaning up before next iteration or use OS-assigned port per iteration?

shivamka1: Also we don't seem to be cleaning up the `_tempdir`. Leaves directories on disk permanently.

raphtory-graphql/tests/utils/validate.rs
```diff
@@ -0,0 +1,317 @@
...
+pub fn validate_namespace_grant(
+    path: &str,
+    permission: Option<Permission>,
+    client: &RaphtoryGraphQLClient,
+    children: &[String],
+) {
+    match permission {
+        Some(permission_enum) => match permission_enum {
+            Permission::Discover => {
+                assert!(
```
shivamka1, line 144: missing:

```
 assert!(
    !can_introspect_namespace(path, children, client).unwrap(),
    "namespace {path} should not be introspectable"
);
assert!(
    !can_write_namespace(path, client).unwrap(),
    "namespace {path} should not be writable"
);
```

raphtory-graphql/tests/utils/validate.rs
```diff
@@ -0,0 +1,317 @@
...
+                    can_discover_namespace(path, client).unwrap(),
+                    "namespace {path} should be discoverable"
+                );
+            }
+            Permission::Introspect => {
+                assert!(
```
shivamka1, line 150: missing:

```
assert!(
    can_discover_namespace(path, client).unwrap(),
    "namespace {path} should be discoverable"
);
```

raphtory-graphql/tests/utils/validate.rs
```diff
@@ -0,0 +1,317 @@
...
+                    can_introspect_namespace(path, children, client).unwrap(),
+                    "namespace {path} should be introspectable"
+                );
+                assert!(
+                    !can_write_namespace(path, client).unwrap(),
+                    "namespace {path} should not be writable"
+                );
+            }
+            Permission::Read => {
+                assert!(
```
shivamka1, line 160: missing:

```
assert!(
    can_discover_namespace(path, client).unwrap(),
    "namespace {path} should be discoverable"
);
```

raphtory-graphql/tests/utils/validate.rs
```diff
@@ -0,0 +1,317 @@
...
+                    "namespace {path} should be introspectable"
+                );
+                assert!(
+                    !can_write_namespace(path, client).unwrap(),
+                    "namespace {path} should not be writable"
+                );
+            }
+            Permission::Write => {
+                assert!(
+                    can_introspect_namespace(path, children, client).unwrap(),
```
shivamka1, line 171: missing:

```
assert!(
    can_discover_namespace(path, client).unwrap(),
    "namespace {path} should be discoverable"
);
```

python/tests/test_permissions.py
shivamka1: Probably also makes sense to move the rbac prop tests as well to `pometry-storage`

raphtory-graphql/src/client/raphtory_client.rs
```diff
@@ -116,9 +116,9 @@ impl RaphtoryGraphQLClient {
                     .join("\n\t"),
                 _ => format!("{}", errors),
             };
+
             return Err(ClientError::GraphQLErrors(format!(
-                "After sending query to the server:\n\t{}\nGot the following errors:\n\t{}",
-                query, message
+                "Error while executing query: {message}"
```
shivamka1, line 121: This makes debugging harder. When a test fails you no longer know which query caused the error. Worth keeping the query in the message, even if reformatted.

shivamka1 (changes requested):

fabubaker:
Closing since the permission model has changed significantly; tests will need a significant rewrite.

PR #2597, opened 2026-05-06
Adding node layers to tests and loaders

arienandalibi:
Node layers were previously not tested rigorously. They are now being added to tests, proptests, the loaders, and the parquet encoders. The loaders and parquet encoders are also used by materialize.

Fix node layer related bugs that we find.

It shouldn't

proptests

There shouldn't be


raphtory/src/db/api/view/internal/time_semantics/persistent_semantics.rs
```diff
@@ -368,42 +371,57 @@ impl NodeTimeSemanticsOps for PersistentSemantics {
...
         prop_ids: Arc<[usize]>,
     ) -> impl Iterator<Item = (EventTime, LayerId, Vec<(usize, Prop)>)> + Send + Sync + 'graph {
-        let start = w.start;
-        let first_row = if node
-            .additions()
-            .range(EventTime::MIN..start)
-            .iter()
-            .next()
-            .is_some()
+        let mut exact_layers = node.layer_ids_iter(view.layer_ids()).collect_vec();
```
fabianmurariu: no collecting allowed

fabianmurariu (changes requested):
I'd like this to be tested in some of our workflows before we merge

PR #2598, opened 2026-05-06
make list properties always return numpy arrays

ljeub-pometry:

List properties always return numpy arrays instead of python lists or pyarrow arrays


miratepuffin (approved):
remove from email.policy import strict

PR #2600, opened 2026-05-07
Move deletion flag to edge id

ljeub-pometry:

Move deletion flag to the edge id to make it easier to handle when writing to disk


fabianmurariu (approved):
LGTM

PR #2601, opened 2026-05-07
Fix metadata read

miratepuffin:


db4-storage/src/api/edges.rs
```diff
@@ -118,6 +122,13 @@ pub trait EdgeSegmentOps: Send + Sync + std::fmt::Debug + 'static {
     fn immut_lsn(&self) -> LSN;
 
     fn flush(&self) -> Result<(), StorageError>;
+
+    fn check_metadata_disk<PR: AsPropRef>(
```
fabubaker: `check_metadata_immut` seems more consistent with the other methods but not a blocker.

fabubaker (approved):
LGTM

PR #2605, opened 2026-05-12
Enable argument pass-through for tox and fix race condition in file download

ljeub-pometry:

- add the option to pass arguments to the pytest commands run with tox
- make the file download function atomic such that we don't end up with partial files if multiple downloads are triggered simultaneously 

allow for debugging and selecting tests


fabianmurariu (approved):
LGTM

PR #2607, opened 2026-05-13
New cache implementation for the server

ljeub-pometry:

- new cache based on quick_cache
- no more evicting of dirty graphs
- atomic replace of cached graphs
- specifying port=0 for server start now works 
- starting a server without specifying a port now falls back to using an arbitrary port if the default port is in use
- python tests now work even if there is a graphql server running on 1736
- server startup no longer pings the network interface to figure out if the server has started (it uses messaging between threads instead).

- race conditions in the server when replacing graphs

- improved the tests such that they fail when the flush on drop fails (but only when compiled with a special test feature)


raphtory-graphql/Cargo.toml
```diff
@@ -42,6 +42,7 @@ tracing-opentelemetry = { workspace = true }
 tracing-subscriber = { workspace = true }
 walkdir = { workspace = true }
 ordered-float = { workspace = true }
+dashmap.workspace = true
```
fabianmurariu: Remove, probably not used

raphtory-graphql/src/cache.rs
```diff
@@ -0,0 +1,203 @@
...
+        F: Future<Output = Result<GraphWithVectors, E>>,
+        InsertionError: From<E>,
+    {
+        let cache_guard = self
+            .cache
+            .entry_async(key, |key, value| EntryAction::<()>::ReplaceWithGuard)
+            .await;
+        let guard = match cache_guard {
+            EntryResult::Replaced(guard, old_graph) => {
+                drop(old_graph);
```
fabianmurariu: needs blocking

raphtory-graphql/src/cache.rs
```diff
@@ -0,0 +1,203 @@
...
+            EntryResult::Replaced(_guard, graph) => {
+                blocking_compute(move || drop(graph)).await;
+            }
+            _ => {}
+        }
+    }
+
+    /// remove a graph from the cache, locking the cache entry until the graph is dropped and the future has completed.
+    /// if the graph exists, it is dropped first before the future runs
+    pub async fn invalidate_with<E>(&self, key: &str, with: impl Future<Output = E>) -> E {
```
fabianmurariu: maybe closure here instead of just Future

fabianmurariu (changes requested):
LGTM, small changes

fabianmurariu (approved):
LGTM

PR #2611, opened 2026-05-15
remove self dependency in raphtory

fabianmurariu:


miratepuffin (approved):

raphtory/Cargo.toml
```diff
@@ -116,7 +116,9 @@ tokio = { workspace = true }                             # for vector testing
 dotenv = { workspace = true }                            # for vector testing
 streaming-stats = { workspace = true }
 indoc = { workspace = true }
-raphtory = { workspace = true, features = ["test-utils"] } # enable test-utils for integration tests
+proptest.workspace = true
```
ljeub-pometry, line 119: Making this a default feature is terrible, it basically means we are compiling the test code all the time. There is a good chance the actual problem isn't the self-dependency but rather that this has a specified version constraint because it is using the workspace dependency definition. Does making this a path dependency without version fix the issue? If that doesn't work, the other option is to move all the integration tests to a separate crate outside of raphtory which can then depend on the `test-utils` feature without problems.

ljeub-pometry (changes requested):

raphtory-test-utils/tests/algo_tests/mod.rs
```diff
@@ -1,12 +1,20 @@
 use std::collections::HashMap;
 
+#[cfg(test)]
```
ljeub-pometry: these `#[cfg(test)]` shouldn't be needed as this lives inside the integration tests, as in, the whole thing is only compiled when testing anyway?

raphtory-test-utils/src/lib.rs
```diff
@@ -0,0 +1,2 @@
+pub mod assertions;
+pub mod test_utils;
```
ljeub-pometry: Seems like it would be less confusing to call the crate `raphtory-tests` with a mod `utils`? Or you should move the actual tests to a different crate called `integration-tests` or something. It is a bit strange that all the tests live inside a crate called `test-utils`...

ljeub-pometry (changes requested):
naming things is hard...

raphtory-test-utils/src/lib.rs
fabianmurariu: It was on my mind that it might be confusing, but I think I'm going to go with `raphtory-tests` + `utils` feature

PR #2613, opened 2026-05-19
Degree filtering

DanielLacina:


raphtory/src/db/graph/views/filter/model/degree_filter.rs
```diff
@@ -0,0 +1,153 @@
...
+    ) -> Result<model::CompositeExplodedEdgeFilter, GraphError>
+    {
+       Err(GraphError::NotSupported) 
+    } 
+    fn try_as_composite_node_filter(&self) -> Result<model::CompositeNodeFilter, GraphError> {
+        Err(GraphError::NotSupported)
+    }
+}  
+
+impl<M> InternalPropertyFilterBuilder for DegreeFilterBuilder<M>
```
ljeub-pometry, line 107: this seems wrong, we shouldn't have stuff like property_ref for degrees

raphtory/src/db/graph/views/filter/model/degree_filter.rs
```diff
@@ -0,0 +1,153 @@
...
+use crate::db::graph::views::filter::model::property_filter::builders::{PropertyExprBuilder, PropertyExprBuilderInput};
+use crate::db::graph::views::filter::model::{CombinedFilter, EntityMarker, InternalPropertyFilterBuilder, TryAsCompositeFilter};
+use crate::db::graph::views::filter::model;
+use crate::db::graph::views::filter::node_filtered_graph::NodeFilteredGraph;
+use crate::db::{api::view::{GraphViewOps, NodeViewOps}, graph::views::filter::model::{FilterOperator, property_filter::PropertyFilterValue}};
+use crate::errors::GraphError;
+
+
+#[derive(Clone)]
+pub struct DegreeFilterBuilder<M> {
```
ljeub-pometry, line 17: This shouldn't be generic over the entity as it only makes sense for nodes

raphtory/tests/test_filters.rs
```diff
@@ -1664,24 +1657,221 @@ mod test_node_filter {
...
+        let filter = NodeFilter.degree(Direction::OUT).lt(1);
+        let mut node_ids_filtered_by_filter: Vec<VID> = graph.filter(filter).unwrap().nodes().into_iter().map(|node| node.node).collect();
+        node_ids_filtered_by_filter.sort();
+        let mut node_ids_filtered: Vec<VID> = graph.nodes().into_iter().filter(|node| node.out_degree() < 1).map(|node| node.node).collect();
+        node_ids_filtered.sort();
+        assert_eq!(node_ids_filtered, node_ids_filtered_by_filter);
+    }
+
+    #[test]
+    fn test_node_out_degrees_ne() {
```
ljeub-pometry: you should turn this into a prop-test

raphtory/src/db/graph/views/filter/model/degree_filter.rs
```diff
@@ -0,0 +1,153 @@
...
+            direction,
+            entity
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct DegreeFilter {
+    pub direction: Direction,
+    pub operator: FilterOperator,
```
ljeub-pometry, line 34: We should have a think about what these filter operators should look like (they really should be taking two expressions of some kind as input) as you should be able to do things like find me all nodes that have an out-degree bigger than their in-degree... (this applies to the property filters as well)

ljeub-pometry: Basically, I don't think our internal representation of the property filters composes and generalises as nicely as it should.

raphtory/src/db/graph/views/filter/model/degree_filter.rs
```diff
@@ -0,0 +1,153 @@
...
+    fn try_as_composite_edge_filter(&self) -> Result<model::edge_filter::CompositeEdgeFilter, GraphError> {
+         Err(GraphError::NotSupported)
+    }
+    fn try_as_composite_exploded_edge_filter(
+        &self,
+    ) -> Result<model::CompositeExplodedEdgeFilter, GraphError>
+    {
+       Err(GraphError::NotSupported) 
+    } 
+    fn try_as_composite_node_filter(&self) -> Result<model::CompositeNodeFilter, GraphError> {
```
ljeub-pometry, line 102: we will need to fix this or get rid of composite filters completely and rely only on the dynamic filters for graphql and index search as well

raphtory/src/db/graph/views/filter/model/degree_filter.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    fn entity(&self) -> Self::Marker {
+        self.entity.clone()
+    }
+
+    fn filter(&self, filter: PropertyFilterInput) -> Self::Filter {
+        let degree_val = match filter.prop_value {
+            PropertyFilterValue::Single(prop_val) => {
+                prop_val.try_cast(PropType::U64).unwrap().as_u64_lossless().unwrap()
```
ljeub-pometry: are we validating the input to this somewhere or does this panic if we try to filter degrees with e.g., a string value?

raphtory/src/db/graph/views/filter/model/degree_filter.rs
ljeub-pometry: note this probably means that we should rethink the way some of these traits are structured which would hopefully make the internal apis easier to understand

PR #2614, opened 2026-05-20
Db v4 fix metadata

arienandalibi:
Fix `.meta` file not updating node and edge counts when the graph is flushed/dropped. Previously, these would stay 0.

Helps fix bugs in the UI.

It shouldn't.

Some unit tests were added.

There shouldn't be.


raphtory/src/db/api/mutation/addition_ops.rs
```diff
@@ -304,7 +304,23 @@ impl<G: InternalAdditionOps<Error: Into<GraphError>> + StaticGraphViewOps> Addit
...
             .flush()
-            .map_err(|err| MutationError::from(err).into())
+            .map_err(|err| MutationError::from(err).into())?;
+
+        #[cfg(feature = "io")]
+        {
+            use crate::{
+                db::api::view::internal::InternalStorageOps,
+                serialise::metadata::refresh_disk_graph_metadata,
+            };
```
ljeub-pometry, line 314: move these imports to the top of the file

raphtory/tests/test_saved_graphs.rs
```diff
@@ -0,0 +1,568 @@
...
+    prelude::*,
+    serialise::{StableDecode, StableEncode},
+};
+use std::{
+    fs, io,
+    path::{Path, PathBuf},
+};
+use storage::{persist::strategy::PersistenceStrategy, Extension};
+
+// These mirror the (currently `pub(crate)`) column-name constants in
```
ljeub-pometry, line 20: Just make those constants pub, no reason to hide them!

raphtory/tests/test_saved_graphs.rs
```diff
@@ -0,0 +1,568 @@
...
+        .add_edge(5, "alice", "bob", [("msg", Prop::str("hi"))], None)
+        .unwrap();
+
+    // edges_d
+    graph
+        .delete_edge(8, "bob", "server-1", Some("uses"))
+        .unwrap();
+}
+
+#[test]
```
ljeub-pometry, line 496: this test should live in pometry-storage, no reason for it to be in the open source

raphtory/tests/test_saved_graphs.rs
```diff
@@ -0,0 +1,568 @@
...
+//     cargo test ... -- --ignored build_v4_saved_disk_graphs
+#[test]
+fn validate_v4_disk_graphs() {
+    // standalone Raphtory has this disabled but pometry-storage has it enabled
+    if !Extension::disk_storage_enabled() {
+        eprintln!("skipping validate_v4_disk_graphs: disk storage backend not enabled");
+        return;
+    }
+
+    // Load mutates the WAL / control file even via `load_read_only`, so copy
```
ljeub-pometry, line 529: that seems like a bug for load_read_only?

raphtory/src/serialise/metadata.rs
```diff
@@ -29,9 +31,117 @@ impl GraphMetadata {
...
+    }
+
+    let meta_path = data_folder.join(GRAPH_META_PATH);
+    let json = match fs::read_to_string(&meta_path) {
+        Ok(json) => json,
+        Err(err) if err.kind() == std::io::ErrorKind::NotFound => return Ok(()),
+        Err(err) => return Err(err.into()),
+    };
+
+    let mut meta: Metadata = serde_json::from_str(&json)?;
```
ljeub-pometry, line 54: you should verify that the metadata file is actually pointing at the graph that is trying to update it and error otherwise (mostly to avoid really strange and hard to debug nonsense in the future)

raphtory/src/db/api/mutation/addition_ops.rs
```diff
@@ -304,7 +304,23 @@ impl<G: InternalAdditionOps<Error: Into<GraphError>> + StaticGraphViewOps> Addit
...
+
+            if let Some(disk_path) = self.disk_storage_path() {
```
ljeub-pometry, line 316: actually, all this logic should live in pometry-storage and there shouldn't be any need for this to exist in multiple places

ljeub-pometry (changes requested):
- All the logic for handling the disk-side flushing can live in pometry-storage (the tests should also move there)
- This PR includes all of the node layer changes, can you cherry pick out the commits that actually fix the metadata problem such that we can merge that (the node layers is a lot more complicated)

PR #2615, opened 2026-05-21
Added Node Degree Filter

DanielLacina:
I added a Node Degree Filter that can be passed into graph.filter to filter nodes based on if their degrees follow the filter expr represented by the Node Degree Filter. Filter expression include: ge, le, ne, lt, gt, is_in, not_is_in. If any other expression is called, the code panics. 
The whole filtering module is used to provide a more user friendly Python interface when it comes to filtering. We would like to make filtering based on degrees user friendly so that's why I added this feature. 

Yes. It is not documented. 

Through prop tests and Python tests

We will definitely need to refactor the Rust filtering module because it's too coupled to certain operations such as property filtering which fails to generalize and make the code reusable ultimately leading to repeated and inflexible code.


raphtory/src/db/graph/views/filter/model/node_filter/mod.rs
```diff
@@ -118,6 +104,13 @@ impl InternalPropertyFilterFactory for NodeFilter {
     }
 }
 
+impl DegreeFilterFactory for NodeFilter {
+    fn degree(&self, direction: Direction) -> DegreeFilterBuilder {
```
ljeub-pometry: this should have separate methods for `in_degree`, `out_degree` and `degree` (similar to the other public apis), you can have an internal trait with this method to implement it if that seems worth doing

raphtory/src/db/graph/views/filter/model/degree_filter.rs
```diff
@@ -0,0 +1,149 @@
...
+         Err(GraphError::NotSupported)
+    }
+    fn try_as_composite_exploded_edge_filter(
+        &self,
+    ) -> Result<model::CompositeExplodedEdgeFilter, GraphError>
+    {
+       Err(GraphError::NotSupported) 
+    } 
+    fn try_as_composite_node_filter(&self) -> Result<model::CompositeNodeFilter, GraphError> {
+        Err(GraphError::NotSupported)
```
ljeub-pometry: this should be supported

raphtory/src/python/filter/node_filter_builders.rs
```diff
@@ -403,6 +404,21 @@ impl PyNodeFilter {
...
 
+    /// Selects node degree for filtering in a given direction.
+    ///
+    /// Arguments:
+    ///     direction (Direction): Degree direction (`IN`, `OUT`, or `BOTH`).
+    ///
+    /// Returns:
+    ///     filter.FilterOps
+    #[staticmethod]
+    fn degree<'py>(
```
ljeub-pometry: make these separate methods for in/out degree

ljeub-pometry (changes requested):
- make the public api consistent
- shouldn't comment out the gql apis

Cargo.toml
```diff
@@ -14,8 +14,6 @@ members = [
     "raphtory-storage",
     "raphtory-api-macros",
     "raphtory-itertools",
-    "clam-core",
```
fabianmurariu: you need to put these back

raphtory/src/db/graph/views/filter/model/degree_filter.rs
```diff
@@ -0,0 +1,173 @@
...
+        &[]
+    }
+
+    fn entity(&self) -> Self::Marker {
+        NodeFilter
+    }
+
+    fn filter(&self, filter: PropertyFilterInput) -> Self::Filter {
+        let value = match filter.prop_value {
+           PropertyFilterValue::Single(ref prop_val) => PropertyFilterValue::Single(prop_val.clone().try_cast(PropType::U64).unwrap_or(prop_val.clone())),
```
shivamka1: This line attempts to cast the user-supplied filter value to U64, but silently fall back to the original value on failure:
```rust
prop_val.clone().try_cast(PropType::U64).unwrap_or(prop_val.clone())
```

This creates two distinct failure modes depending on the operator used.

**Failure mode 1:** comparison operators (>, <, >=, <=, ==, !=)

If the cast fails (e.g. user passes a Python float like 2.5, which arrives as `Prop::F64`), the filter stores `Prop::F64(2.5)`. At match time (line 56), the actual degree is always `Prop::U64(n)`. In `apply_to_property`, comparing these two different `Prop` variants calls `partial_cmp`, which returns `None` for cross-type comparisons, and the `unwrap_or(false)` means every node silently fails the filter. The result is an empty graph with no error raised.

**Failure mode 2:** `is_not_in` with wrong types

If the filter set contains `{F64(1.5), F64(2.5)}` and the node's degree is `U64(3)`, then s`et.contains(&Prop::U64(3))` is `false` (different enum variant), so `!false = true ` and every node silently passes regardless of its actual degree.

---

Since degree is always a non-negative integer, reject non-castable values at construction time rather than silently proceeding:

```rust 
let cast = prop_val.clone().try_cast(PropType::U64)
    .ok_or_else(|| GraphError::InvalidInput(
        format!("degree filter expects an integer value, got {:?}", prop_val)
    ))?;
PropertyFilterValue::Single(cast)
```

This requires changing the return type to `Result<Self::Filter, GraphError>` and propagating the error up through `create_node_filter` / `create_filter`, which already return `Result`.

raphtory/src/db/graph/views/filter/model/degree_filter.rs
```diff
@@ -0,0 +1,173 @@
...
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct DegreeFilter {
+    pub direction: Direction,
+    pub operator: FilterOperator,
+    pub value: PropertyFilterValue
+}
+
+impl DegreeFilter {
+    pub fn matches<'graph, G: GraphViewOps<'graph>>(&self, graph: &G, node: VID) -> bool {
+        let node_view = graph.node(node).unwrap();
```
shivamka1: I would suggest to move the degree computation into `NodeDegreeFilterOp::apply` directly, and delete `DegreeFilter::matches` entirely. This removes the high-level `graph.node()` call, stays consistent with the rest of the module, and avoids the `unwrap()` on principle even though it's safe.

raphtory/src/db/graph/views/filter/model/degree_filter.rs
```diff
@@ -0,0 +1,173 @@
...
+           PropertyFilterValue::None => PropertyFilterValue::None
+        };  
+        DegreeFilter {
+             value,
+             direction: self.direction,
+             operator: filter.operator 
+        }
+    }
+
+    fn with_expr_builder(&self, builder: PropertyExprBuilderInput) -> Self::ExprBuilder {
```
shivamka1, line 180: `filter.Node.degree()` returns a `PyPropertyExprBuilder`, which exposes `.first()`, `.last()`, `.sum()`, `.avg()`, `.len()`, `.min()`, `.max()`, `.any()`, `.all()`, `.temporal()` to Python users. Calling any of them routes through `DegreeFilterBuilder::with_expr_builder`, which panics unconditionally.

None of these make semantic sense on a degree. Degree isn't a list property and has no temporal history, so they should be unavailable, not available-but-crashing.

The fix would be to return a purpose-built thin type from `degree()` instead of `PyPropertyExprBuilder`, the same way `name()` returns `PyNodeNameFilterBuilder`. That type only exposes the operators that actually work (==, !=, <, <=, >, >=, is_in, is_not_in), and Python users get an `AttributeError` at the call site rather than a panic inside Rust.

raphtory-graphql/src/model/graph/filtering.rs
```diff
@@ -1388,6 +1408,26 @@ impl TryFrom<GqlNodeFilter> for CompositeNodeFilter {
                     operator,
                 }))
             }
+            GqlNodeFilter::Degree(degree) => {
+                let core_direction = match degree.direction {
```
shivamka1: Instead of the inline match:
```rust
impl From<DegreeDirection> for Direction {
    fn from(d: DegreeDirection) -> Self {
        match d {
            DegreeDirection::In   => Direction::IN,
            DegreeDirection::Out  => Direction::OUT,
            DegreeDirection::Both => Direction::BOTH,
        }
    }
}
```
Then the conversion site becomes `degree.direction.into()` instead of three lines.

python/tests/test_base_install/test_filters/test_node_filter.py
```diff
@@ -1,4 +1,5 @@
-from raphtory import filter
+from raphtory import filter, Prop
```
shivamka1: The tests cover individual operators but there's no test for `NodeFilter.degree().gt(2).and(NodeFilter.in_degree().lt(3))` style composition, which exercises the `TryAsCompositeFilter` impl on `DegreeFilter`.

shivamka1 (changes requested):

raphtory/src/db/graph/views/filter/model/degree_filter.rs
DanielLacina: Now I understand the ideal business logic. However, this involves modifying previously declared traits such as InternalPropertyFilterBuilder because I would have to change the return type of its filter method.

raphtory/src/db/graph/views/filter/model/degree_filter.rs
DanielLacina: Like I mentioned earlier, this would involve modifying traits that were previously defined such as InternalPropertyFilterBuilder -> with_expr_builder. I had to use InternalPropertyFilterBuilder to avoid writing redundant code. This means my filter has to implement methods it doesn't need. We don't have traits that generalize the overall logic of the filtering module. This leads to two undesirable scenarios where we either write redundant code or use reusable code that forces filters to implement un related methods.

raphtory/src/db/graph/views/filter/model/degree_filter.rs
ljeub-pometry: You can defer the validation until the call to create_filter which already returns a result (that is how the property filters work as they do not know the actual type before being applied to the graph). We can then refactor this later as rejecting nonsense values early would certainly be a nicer user experience.

raphtory/src/db/graph/views/filter/model/degree_filter.rs
ljeub-pometry: you shouldn't panic unconditionally though, as per my other comment, you can record the operation and then error when this is turned into an actual filter. While that is not ideal, it is probably the best we can do with the existing apis before we refactor this.

raphtory/src/db/api/state/ops/filter.rs
```diff
@@ -226,6 +221,31 @@ impl<G: GraphView> NodeOp for NodePropertyFilterOp<G> {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct NodeDegreeFilterOp<G> {
```
ljeub-pometry, line 225: this probably should wrap `Degree<G>` to get the value instead of implementing it using the high-level apis. No need to track graph and direction separately here in that case, just 
```rust
pub struct NodeDegreeFilterOp<G> {
    degree: Degree<G>,
    operator: FilterOperator,
    value: PropertyFilterValue
}
```

fabianmurariu (approved):
LGTM

PR #2617, opened 2026-05-22
remove proto completely as we are no longer planning to support it

ljeub-pometry:

removes all code related to the old proto graph format


fabianmurariu (approved):
LGTM, but I didn't see prost being removed from Raphtory/Cargo.toml

PR #2623, opened 2026-05-26
don't set port in tests

ljeub-pometry:

- remove more hard-coded ports in python tests so they don't fail if a server is running on that port


miratepuffin (approved):
LGTM

PR #2624, opened 2026-05-27
Db v4 move graph folder

arienandalibi:
Move graph_folder.rs constants (directory and file names regarding exported graphs) out of the raphtory crate and into raphtory-api.

Following previous changes, we needed access to GRAPH_META_PATH in pometry-storage, where we can't import from raphtory.

No

Tests still pass

No


raphtory/src/serialise/graph_folder.rs
```diff
@@ -25,29 +31,6 @@ use std::{
 use walkdir::WalkDir;
 use zip::{write::FileOptions, ZipArchive, ZipWriter};
```
ljeub-pometry: We can just move this entire file to raphtory-api, then you can just call the method in here to update the metadata.

ljeub-pometry (changes requested):

raphtory/src/db/api/storage/storage.rs
```diff
@@ -70,13 +74,19 @@ pub struct Storage {
 impl Drop for Storage {
     fn drop(&mut self) {
         if let Some(disk_path) = self.graph.disk_storage_path() {
```
ljeub-pometry: this should just happen in the flush for the disk storage, no reason to have this in OS raphtory any more and `Storage` should no longer implement `Drop`

ljeub-pometry (changes requested):
the Drop implementation for storage should be gone, the metadata write can now be part of the flush in the disk storage

raphtory/src/db/api/mutation/addition_ops.rs
```diff
@@ -309,13 +315,17 @@ impl<G: InternalAdditionOps<Error: Into<GraphError>> + StaticGraphViewOps> Addit
...
+                    disk_path.file_name().and_then(|name| name.to_str()),
+                ) {
+                    let meta = Metadata {
+                        path: graph_dir.to_string(),
+                        meta: build_graph_metadata(self),
+                    };
+                    meta.write_atomic(data_folder, &data_folder.join(GRAPH_META_PATH))
+                        .map_err(|err| MutationError::from(StorageError::from(err)).into())?;
+                }
             }
```
ljeub-pometry: this isn't needed here either once you have the actual logic fixed in the storage

fabianmurariu (approved):
LGTM

PR #2626, opened 2026-05-31
Guix package

phdavis1027:

`guix.scm` and `python/guix.scm`, which produce packages for the raphtory graphql server and python bindings respectively.

It creates a reproducible environment for testing, deploying, and developing. Guix has also made an effort to serve the scientific computing ecosystem; in that sense it is a natural fit.

No, if users don't care about Guix they can completely ignore this.

Built both, shipped as self-contained tarballs using `guix pack`, and am currently developing scripts against them on a remote server.

No

### DIFF
--- a/guix.scm
+++ b/guix.scm
@@ -1,0 +1,55 @@
+(define-module (raphtory packages)
+  #:use-module (guix packages)
+  #:use-module (guix git-download)
+  #:use-module (guix build-system cargo)
+  #:use-module (guix gexp)
+  #:use-module ((guix licenses) #:prefix license:)
+  #:use-module (guix import crate)
+  #:use-module (gnu packages rust)
+  #:use-module (gnu packages cmake)
+  #:use-module (gnu packages pkg-config)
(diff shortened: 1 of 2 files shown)


phdavis1027:
Quick update: During testing against this build, I have observed that when I run `graph.edges.to_df(explode=True)`, I get a DataFrame where every column of update_history  is 0, but if i instead run `graph.edges.to_df().explode('update_history')` the update_history column has all values I'd expect.

It is, I suppose, possible that this is a bug in Raphtory v0.17.0, but it seems much more likely to be a bug in this Guix build. When I have an extra moment, I will try to reproduce on a host-system build, which will hopefully clarify the situation.

fabianmurariu:
@phdavis1027 does this need to live here? We can't commit to maintaining this. Can this live in some other repository unrelated to Raphtory?

phdavis1027:
> @phdavis1027 does this need to live here? We can't commit to maintaining this. Can this live in some other repository unrelated to Raphtory?

Understood, figured I'd propose it just in case

PR #2627, opened 2026-06-01
Global cache on edge loads

fabianmurariu:


clam-core
ljeub-pometry, line 1: why did this update?

ljeub-pometry (approved):
This makes more sense than the previous per-chunk cache but ideally this would just be part of the on-disk resolver itself?

clam-core
fabianmurariu: because we were pointing to a random commit

PR #2629, opened 2026-06-02
fix to_parquet on NodeState

ljeub-pometry:

make the ids column `LargeStringArray` to avoid integer overflows on large graphs


rodrigolece:
Merging :)

PR #2638, opened 2026-06-05
fix week alignment (should align to Monday, not Thursday)

ljeub-pometry:

`AlignmentUnit::Week` now uses Monday as the start of the week 

1970-1-1 is a Thursday...

Fixes an alignment bug

Added a prop test to verify the alignment


fabianmurariu (approved):
LGTM

PR #2642, opened 2026-06-10
Fix UI assuming raphtory is at 1736

louisch:

Fixes raphtory-graphql's UI assuming raphtory is at 1736 (even when it isn't) (this has been fixed in the past, I just accidentally broke that recently)

Without this, the UI doesn't work whenever raphtory's port is changed

No

In the UI tests raphtory is run on port 1737 and the tests run on the bundled UI (this was already the case, it was just using the default 1736 before) so that if this regresses the CI should catch it next time.

No


ljeub-pometry (changes requested):
We need to add a test to make sure we don't break this again!

ljeub-pometry (approved):

PR #2647, opened 2026-06-12
Group raphtory-tests into fewer binaries

arienandalibi:
Group test files based on test contents. By grouping them into folders, we make multiple test files compile into a single "crate" (binary). This saves us time when building because, for each single test file not in a folder, a whole binary is built, and this binary needs to link all of raphtory and its associated crates. By limiting the number of binaries, we also limit this overhead.

Faster build times

No

All the tests are still run and still pass

No


raphtory-tests/tests/filter_tests/mod.rs
```diff
@@ -0,0 +1,89 @@
+use raphtory::{db::api::view::StaticGraphViewOps, prelude::*};
+
+mod cached_view;
+mod edge_property_filter;
+mod node_property_filter;
+mod subgraph_tests;
+mod test_filters;
+mod test_layers;
+mod tests_node_type_filtered_subgraph;
+mod views_test;
```
fabubaker, line 11: We're missing `exploded_edge_property_filter` here, which means those tests aren't being run...

raphtory-tests/tests/view_tests/cached_view.rs
```diff
@@ -0,0 +1,125 @@
...
+    test_storage!(&graph, |graph| {
+        let sg = graph.cache_view();
+
+        let actual = sg.materialize().unwrap().into_events().unwrap();
+        assert_graph_equal(&actual, &sg);
+    });
+}
+
+#[test]
+fn test_optimisation_flags() {}
```
fabubaker: Probably remove this empty test, must have been a holdover from refactoring.

raphtory-tests/tests/filter_tests/exploded_edge_property_filter.rs
fabubaker, line 1: I want to be sure we're not missing any tests, so on `db_v4` there are a total of 1287 Rust tests being run - make sure the number of tests on this branch matches that.

fabubaker (changes requested):
Need to add back some missing tests. 

Please change the PR title to be more descriptive, for eg, "Group raphtory-tests into fewer binaries"

PR #2649, opened 2026-06-16
update python tests

fabianmurariu:
re-arrange some of the tests to isolate fireducks as a dependency


python/tests/test_ingestion_equivalence_df.py
ljeub-pometry, line 1: these tests aren't running right now with `tox`

ljeub-pometry (changes requested):
Need to add the fireducks test to tox.ini

PR #2652, opened 2026-06-18
Patch fault-injector errors

fabubaker:
Fixes all fault-injector/LazyFS bugs that show up on CI.

Quick summary of changes:
* Call `sync_all` on the `.meta` file.
* Use atomic rename when modifying the `Config` file.
* Run a clean shutdown only if the DB was in `Running` state.
* Update the WAL position correctly when replaying a WAL that has no entries.

These changes should fix any durability bugs revealed by LazyFS.

No.

Running the fault-injector CI on pometry-storage: https://github.com/Pometry/pometry-storage/pull/304

No.


ljeub-pometry (approved):
Looks good

PR #2655, opened 2026-06-19
Add more benchmarks to algobench

DanielLacina:
Added algo benchmarks that benchmark all algorithms with different views
To give insights on what algorithms need to be improved
no
its a benchmark
no


raphtory-benchmark/benches/algobench.rs
```diff
@@ -138,6 +2203,57 @@ criterion_group!(
...
+    graphgen_internal_star_event,
+    graphgen_internal_init_star_count,
+    graphgen_internal_new_triangle_edge,
+    graphgen_internal_init_tri_count,
+    graphgen_internal_global_triangle_motifs,
+    graphgen_internal_local_triangle_motifs,
     temporal_motifs,
 );
-criterion_main!(benches);
+criterion_main!(benches);
```
fabubaker: Missing newline at the end of the file - opening in vscode with your updated config should automatically add it.

raphtory-benchmark/benches/algobench.rs
```diff
@@ -1,134 +1,2199 @@
...
+            let result = run();
+            black_box(result);
+        });
     });
+    group.finish()
+}
 
-    group.finish();
+fn large_random_attachment_graph() -> Graph {
+    let graph = Graph::new();
```
fabubaker:
```suggestion
}

// Graph Constructors

fn large_random_attachment_graph() -> Graph {
    let graph = Graph::new();
```

Just to break things up so it's easier to eyeball where the graph setup/benchmarking starts.

raphtory-benchmark/benches/algobench.rs
```diff
@@ -1,134 +1,2199 @@
...
+}
+
+fn large_typed_random_attachment_layered() -> impl StaticGraphViewOps {
+    let graph = large_typed_random_attachment_graph();
+    graph.default_layer()
+}
+
+pub fn local_triangle_count_analysis(c: &mut Criterion) {
+    graph_benchmark_with_setup(
+        c,
```
fabubaker:
```suggestion
    graph.default_layer()
}

// Benchmarks

pub fn local_triangle_count_analysis(c: &mut Criterion) {
    graph_benchmark_with_setup(
        c,
```

fabubaker (approved):
LGTM

Edit: Will need to fix the errors on CI...

raphtory-benchmark/benches/algobench.rs
```diff
@@ -1,134 +1,2220 @@
...
+        large_random_attachment_layered,
+        |graph, _| louvain::<ModularityUnDir, _>(graph, 1.0, None, None),
+    );
+    graph_benchmark(
+        c,
+        "graphgen_large_louvain_graph_filtered",
+        20,
+        10,
+        large_random_attachment_filtered,
+        |graph, _| louvain::<ModularityUnDir, _>(graph, 1.0, None, None),
```
fabubaker: This is broken on CI

fabubaker (dismissed):

raphtory-benchmark/src/algobench_common.rs
```diff
@@ -0,0 +1,292 @@
...
+    sync::OnceLock,
+    time::{Duration, Instant},
+};
+
+// Criterion normally chooses how many times to call the benchmarked routine per sample based
+// on the measurement time budget, which is what let algobench_slow balloon to 500-1,000 real
+// algorithm calls per benchmark even with a short measurement window. Pinning this instead
+// guarantees every algobench benchmark does exactly `sample_size * ITERS_PER_SAMPLE` real runs
+// of the algorithm, independent of how Criterion's timing estimate comes out.
+const ITERS_PER_SAMPLE: u64 = 10;
```
ljeub-pometry: This doesn't make sense unless some algorithms have extremely variable runtimes? Criterion automatically adjusts this to get reasonable statistics for short benchmarks.

raphtory-benchmark/src/algobench_common.rs
```diff
@@ -0,0 +1,292 @@
...
+{
+    let mut group = c.benchmark_group(name);
+    let graph = build_graph();
+    let setup_data = setup(&graph);
+
+    group.sampling_mode(SamplingMode::Flat);
+    group.warm_up_time(Duration::from_millis(200));
+    group.sample_size(sample_size);
+    group.bench_function(name, |b| {
+        b.iter_custom(|iters| {
```
ljeub-pometry: Standard criterion iter should be fine with sample size set appropriately (it will run slow algorithms sample-size times and fast algorithms more often to hit the allocated budget).

raphtory-benchmark/src/algobench_common.rs
```diff
@@ -0,0 +1,292 @@
...
+// algorithm calls per benchmark even with a short measurement window. Pinning this instead
+// guarantees every algobench benchmark does exactly `sample_size * ITERS_PER_SAMPLE` real runs
+// of the algorithm, independent of how Criterion's timing estimate comes out.
+const ITERS_PER_SAMPLE: u64 = 10;
+
+pub fn graph_benchmark_with_setup<G, BuildGraph, Setup, Run, SetupData, Output>(
+    c: &mut Criterion,
+    name: &str,
+    sample_size: usize,
+    build_graph: BuildGraph,
```
ljeub-pometry, line 22: You can just pass in `graph: G`?

raphtory-benchmark/src/algobench_common.rs
```diff
@@ -0,0 +1,292 @@
...
+    setup: Setup,
```
ljeub-pometry, line 23: Same as above, no reason to pass in a closure, you are only calling it once, you can just pass in the result

raphtory-benchmark/src/algobench_common.rs
```diff
@@ -0,0 +1,292 @@
...
+
+pub fn build_large_typed_random_attachment_graph() -> Graph {
+    let graph = build_large_random_attachment_graph();
+    for id in graph.nodes().id().iter_values() {
+        graph
+            .add_node(0, id, NO_PROPS, Some("Right"), None)
+            .expect("unable to set node type");
+    }
+    graph
+}
```
ljeub-pometry, line 138: this might deadlock

raphtory-benchmark/src/algobench_common.rs
```diff
@@ -0,0 +1,292 @@
...
+pub fn medium_weighted_random_attachment_graph() -> Graph {
+    static GRAPH: OnceLock<Graph> = OnceLock::new();
+    GRAPH
+        .get_or_init(build_medium_weighted_random_attachment_graph)
+        .clone()
+}
+
+pub fn build_medium_typed_random_attachment_graph() -> Graph {
+    let graph = build_medium_random_attachment_graph();
+    for id in graph.nodes().id().iter_values() {
```
ljeub-pometry: this might dead-lock

ljeub-pometry (changes requested):
- I don't think you need the custom iter (unless some algorithm has extremely variable runtime, but then benchmarking it would be essentially useless anyway)
- don't shrink the warmup time too much (might contribute to bad timing estimates)
- Possible deadlocks in graph construction (see comments)
- Could simplify some code by getting rid of closures that get only called once and pass in the values directly (not really a problem that you need to change, more of a stylistic comment for the future)

raphtory-benchmark/src/algobench_common.rs
```diff
@@ -0,0 +1,286 @@
...
+    let graph = build_graph();
+    let setup_data = setup(&graph);
+
+    group.sampling_mode(SamplingMode::Flat);
+    group.measurement_time(std::time::Duration::from_secs(measurement_secs));
+    group.sample_size(sample_size);
+    group.bench_function(name, |b| {
+        b.iter(|| {
+            let result = run(&graph, &setup_data);
+            black_box(result);
```
ljeub-pometry: just for future reference, if you return the result, `iter` will call `black_box` on it for you

raphtory-benchmark/src/algobench_common.rs
```diff
@@ -0,0 +1,286 @@
...
+
+pub fn large_random_attachment_subgraph() -> NodeSubgraph<Graph> {
+    let graph = large_random_attachment_graph();
+    let subgraph = graph.subgraph(graph.nodes());
+    subgraph
+}
+
+pub fn large_random_attachment_filtered() -> impl StaticGraphViewOps {
+    large_random_attachment_graph()
+        .filter(NodeFilter.degree().ge(0u64))
```
ljeub-pometry: you might want to replace this with `ge(1u64)`, we might get clever in the future and optimise this filter away as it is always the entire graph

raphtory-benchmark/benches/algobench_fast.rs
```diff
@@ -0,0 +1,344 @@
...
+pub fn graphgen_alternating_mask(c: &mut Criterion) {
+    graph_benchmark(
+        c,
+        "graphgen_alternating_mask",
+        3,
+        10,
+        large_random_attachment_graph,
+        |graph, _| alternating_mask(graph),
+    );
+}
```
ljeub-pometry: we should get rid of this silly example algorithm

raphtory-benchmark/benches/algobench_fast.rs
```diff
@@ -0,0 +1,344 @@
...
+    simple_benchmark(c, "graphgen_internal_new_triangle_edge", 2, 10, || {
+        new_triangle_edge(true, 1, 0, 1, 100)
+    })
+}
+
+pub fn graphgen_internal_init_tri_count(c: &mut Criterion) {
+    simple_benchmark(c, "graphgen_internal_init_tri_count", 2, 10, || {
+        init_tri_count(128)
+    })
+}
```
ljeub-pometry: these are internal functions of the motif counting algorithm, no reason to benchmark them separately, they probably shouldn't be public in the first place

raphtory-benchmark/src/algobench_common.rs
```diff
@@ -0,0 +1,286 @@
+#![allow(dead_code)]
```
ljeub-pometry: why would we want dead code here?

raphtory-benchmark/benches/algobench_medium.rs
```diff
@@ -0,0 +1,383 @@
...
+use raphtory_api::core::Direction;
+use raphtory_benchmark::algobench_common::{
+    first_node_id, graph_benchmark, graph_benchmark_with_setup, medium_random_attachment_filtered, medium_random_attachment_graph, medium_random_attachment_layered, medium_random_attachment_subgraph, medium_typed_random_attachment_graph, medium_weighted_random_attachment_graph,
+};
+
+pub fn graphgen_clustering_coeff(c: &mut Criterion) {
+    graph_benchmark(
+        c,
+        "graphgen_clustering_coeff",
+        10,
```
ljeub-pometry: I'm assuming the different sampling times are to avoid warnings about not being able to complete 10 samples in the default 5s? If not, you can reduce them.

ljeub-pometry (changes requested):
Basically fine, some minor changes and we can get this merged
 
- remove benchmarks for internal functions
- review the sample times
- we probably want a non-trivial filter for the filtering benchmarks

raphtory-benchmark/benches/algobench_medium.rs
DanielLacina: I don't think I have a parameter for different sampling times. I think the parameter you're looking at is the number of samples.

ljeub-pometry (approved):

PR #2658, opened 2026-06-23
Implement property presence bitset per layer

arienandalibi:
Adding a bitset (Vec<bool>) to keep track of which properties exist in which layers. Use this to optimize iterators (especially relevant when using disk storage). Also optimized the generation of schemas (node schema, edge schema, ...) by iterating edges only once and implementing a cache for already generated schemas. 

Optimizations can be made to improve UI experience and disk accesses.

It shouldn't

Will need to be tested using the UI

There shouldn't be


db4-storage/src/pages/edge_page/writer.rs
```diff
@@ -57,6 +57,13 @@ impl<'a, MP: DerefMut<Target = MemEdgeSegment> + std::fmt::Debug, ES: EdgeSegmen
         layer_id: LayerId,
     ) -> LocalPOS {
         self.graph_stats.update_time(t.t());
+        // Update the per-layer property presence bitset in Meta.
+        // `.inspect` runs once per emitted item as the iterator is consumed in `insert_edge_internal`
+        let meta = self.writer.edge_meta().clone();
```
fabianmurariu: This is going to light up in bulk loading quite badly, we need a clone free variant

raphtory-api/src/core/entities/properties/meta.rs
```diff
@@ -285,7 +301,31 @@ impl PropMapper {
...
+    /// O(1) check: has property `prop_id` ever been observed in `layer_id`?
+    /// `false` is authoritative; callers can safely skip column reads for
+    /// this (layer, prop). `true` means at least one entity in `layer_id`
+    /// has prop `prop_id`.
+    #[inline]
+    pub fn layer_has(&self, layer_id: LayerId, prop_id: usize) -> bool {
+        self.layer_prop_presence
+            .read_recursive()
+            .get(layer_id.0)
+            .and_then(|row| row.get(prop_id))
```
fabianmurariu, line 317: is_some?

raphtory-graphql/src/model/schema/layer_schema.rs
```diff
@@ -1,48 +1,159 @@
...
+) -> Vec<PropertySchema> {
+    let mut schema: HashMap<(String, String), HashSet<String>> = HashMap::new();
+    for props in props_per_edge {
+        for ((key, value), id) in props.iter().zip(props.ids()) {
+            let Some(value) = value else { continue };
+            let key_with_prop_type = (
+                key.to_string(),
+                mapper
+                    .get_dtype(id)
+                    .expect("type for internal id should always exist")
```
fabianmurariu: unwrap_or_else with a panic inside is a bit better

fabianmurariu (changes requested):
Maybe we can avoid some clones?

PR #2661, opened 2026-06-23
Fix graphql meta graph

ljeub-pometry:

- `node_count` and `edge_count` on `MetaGraph` update when the graph is modified in memory

- on-disk metadata is only updated when the graph is evicted from the cache

added a test to check the counts update properly


louisch (approved):
lgtm

PR #2668, opened 2026-06-26
Open telemetry tests

DanielLacina:
Added Open Telemetry tests and configurations to the TracingConfig to make spans and logs accessible when performing tests

Tests are needed to verify that spans/logs are being emitted and that config behavior works as intended.

No.

This patch contains tests. 

No.


raphtory-graphql/src/config/otlp_config.rs
```diff
@@ -119,6 +124,8 @@ pub struct TracingConfig {
     /// Headers to use when transport_protocol is set to HTTP
     pub transport_headers: HashMap<String, String>,
     pub transport_certificate: Option<PathBuf>,
+    #[cfg(test)]
+    pub exporters: InMemoryExporters,
```
fabubaker: A comment here explaining why we're storing these in the struct during `cfg(test)` would be helpful.

raphtory-graphql/src/config/otlp_config.rs
```diff
@@ -279,3 +299,42 @@ impl TracingConfig {
...
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let _ = <Option<()>>::deserialize(deserializer)?;
+        Ok(Self {
+            span_exporter: InMemorySpanExporter::default(),
+            log_exporter: InMemoryLogExporter::default(),
+        })
+    }
```
fabubaker: Rather than implementing dummy `Serialize` / `Deserialize` for `InMemoryExporters`, it's cleaner to just tag the `exporters` field with `#[serde(skip)]`.

raphtory-graphql/src/server.rs
```diff
@@ -494,9 +484,15 @@ async fn server_termination(
...
+    use crate::{
+        client::raphtory_client::RaphtoryGraphQLClient, config::{
+            app_config::AppConfigBuilder, otlp_config::{ESSENTIAL_TRACE_SPANS, TracingLevel, TracingProtocol},
+        }, server::GraphServer,
+    };
     use chrono::prelude::*;
-    use raphtory::{
+    use opentelemetry::trace::SpanId;
+    use opentelemetry_sdk::{logs::in_memory_exporter::LogDataWithResource, trace::SpanData};
+use raphtory::{
```
fabubaker: Need to correctly indent this line.

raphtory-graphql/src/server.rs
```diff
@@ -535,6 +531,49 @@ mod server_tests {
 
         running.stop().await
     }
+    use url::Url;
```
fabubaker: Missing a newline after this method.

raphtory-graphql/src/server.rs
```diff
@@ -572,4 +611,363 @@ mod server_tests {
         sleep(Duration::from_secs(5)).await;
         handler.await.unwrap().stop().await
     }
+
+
+    fn expected_operations(
+        defined_query_operations: Vec<String>,
+        tracing_level: TracingLevel,
+    ) -> HashSet<String> {
+        let mut expected_operations_: HashSet<String> = match tracing_level {
```
fabubaker:
```suggestion
        let mut expected_operations: HashSet<String> = match tracing_level {
```

Can remove the trailing underscore?

raphtory-graphql/src/server.rs
```diff
@@ -572,4 +611,363 @@ mod server_tests {
...
+            "request".to_string(),
+            "parse".to_string(),
+            "validation".to_string(),
+            "execute".to_string(),
+        ]);
+        expected_operations_
+    }
+
+    fn all_operations(graph_path: String) -> (String, Vec<String>) {
+        let operations: Vec<String> = vec![
```
fabubaker: Let's cut down the number of operations we're checking to ~ 10, with a mix of essential/complete and read/write operations. We don't need this test to be very strict, just a basic spot test that the levels are working fine.

raphtory-graphql/src/server.rs
```diff
@@ -572,4 +611,363 @@ mod server_tests {
...
+        expected_operations_.extend([
+            "request".to_string(),
+            "parse".to_string(),
+            "validation".to_string(),
+            "execute".to_string(),
+        ]);
+        expected_operations_
+    }
+
+    fn all_operations(graph_path: String) -> (String, Vec<String>) {
```
fabubaker: You can extract out the operations and query into constant variables rather than returning them from this function. Parameterizing `graph_path` is not super helpful here...

raphtory-graphql/src/server.rs
```diff
@@ -572,4 +611,363 @@ mod server_tests {
...
+            missing_from_actual,
+            unexpected_in_actual
+        );
+    }
+
+    #[tokio::test]
+    async fn test_open_telemetry_spans_complete() {
+        let tracing_level = TracingLevel::COMPLETE;
+        let (query, defined_query_operations) = all_operations("g".to_string()); 
+        let (finished_spans, _emitted_logs) = get_spans_and_logs(tracing_level.clone(), query).await;
```
fabubaker: We seem to not be doing anything with `emitted_logs`...

raphtory-graphql/src/server.rs
```diff
@@ -572,4 +611,363 @@ mod server_tests {
         sleep(Duration::from_secs(5)).await;
         handler.await.unwrap().stop().await
     }
+
+
+    fn expected_operations(
```
fabubaker: Rather than programmatically extracting out the expected operations, I would just hard-code them as constant variables for the three tracing levels and then do a simple assert.

raphtory-graphql/src/server.rs
```diff
@@ -572,4 +611,363 @@ mod server_tests {
...
+        let emitted_logs = app_config
+            .tracing
+            .exporters
+            .log_exporter
+            .get_emitted_logs()
+            .unwrap();
+        (finished_spans, emitted_logs)
+    }
+
+    fn assert_spans(tracing_level: TracingLevel, defined_query_operations: Vec<String>, finished_spans: Vec<SpanData>) {
```
fabubaker: Once you've hardcoded the query and the operations for the three tracing levels, I'm expecting this function to go away. 

In general, try to embed asserts directly in the test methods as much as you can instead of handing it off to another function, so that it's easy to eyeball what the test is doing.

raphtory-graphql/src/server.rs
```diff
@@ -572,4 +611,363 @@ mod server_tests {
...
+                    }}
+                    flush
+                }}
+            }}",
+            graph_path
+        );
+        (query, operations) 
+    }
+
+    async fn get_spans_and_logs(tracing_level: TracingLevel, query: String) -> (Vec<SpanData>, Vec<LogDataWithResource>) {
```
fabubaker: I would try and inline much of the code in this function in the test function itself so that you are directly and explicitly documenting what you are testing. This is a nice article that explains this: https://davistobias.com/articles/dont-dry-your-tests/

fabubaker (changes requested):
Most looks good, some minor things to cleanup. I would also fill in the PR body 🙂.

But `test_open_telemetry_spans_complete` seems to be failing on CI though...

raphtory-graphql/src/config/otlp_config.rs
fabubaker: It's pretty obvious that the exporters are here for testing because of the `#[cfg(test)]` to begin with. I was looking for a comment that says the exporters are here so we can pull the spans/logs from them during testing. So something like

> Expose in-memory exporters to retrieve spans and logs in tests.

raphtory-graphql/src/server.rs
```diff
@@ -572,4 +583,262 @@ mod server_tests {
...
+        }
+    }
+
+    fn print_span_node(
+        span: &opentelemetry_sdk::trace::SpanData,
+        depth: usize,
+        children: &mut HashMap<SpanId, Vec<&opentelemetry_sdk::trace::SpanData>>,
+    ) {
+        println!(
+            "{}{} [{} -> {}]",
```
fabubaker: If we're not using these methods in tests I would remove them.

raphtory-graphql/src/server.rs
```diff
@@ -572,4 +583,262 @@ mod server_tests {
...
+            Config::default(),
+        )
+        .await
+        .unwrap();
+        let handler = server.start_with_port(0).await.unwrap();
+
+        let endpoint = Url::parse(&format!("http://localhost:{}/", handler.port())).unwrap();
+        let client = RaphtoryGraphQLClient::new(endpoint, None);
+
+        let query = "query {
```
fabubaker: If you want, you could move this out into a const variable right before `test_open_telemetry_spans_complete` and reuse it in all the tests.

```rust
const OPEN_TELEMETRY_QUERY = "query {
...
...
...
}";

#[tokio::test]
async fn test_open_telemetry_spans_complete() {
         // Use OPEN_TELEMETRY_QUERY here
}
```

raphtory-graphql/src/server.rs
```diff
@@ -572,4 +583,262 @@ mod server_tests {
...
+            .build();
+
+        let server = GraphServer::new(
+            tmp_dir.path().to_path_buf(),
+            Some(app_config.clone()),
+            Config::default(),
+        )
+        .await
+        .unwrap();
+        let handler = server.start_with_port(0).await.unwrap();
```
fabubaker: If you want, you could move this into a method called `open_telemetry_setup` that takes the tracing level as an arg and returns the handler.

raphtory-graphql/src/server.rs
```diff
@@ -572,4 +583,262 @@ mod server_tests {
...
+            "updateGraph".to_string(),
+            "validation".to_string(),
+        ]));
+        let emitted_logs = app_config
+            .tracing
+            .exporters
+            .log_exporter
+            .get_emitted_logs()
+            .unwrap(); 
+        // assert!(emitted_logs.len() == 1);
```
fabubaker: We need to make sure logs are also being emitted correctly...

fabubaker (changes requested):
The tests are still failing on CI...

raphtory-graphql/src/server.rs
```diff
@@ -572,4 +583,262 @@ mod server_tests {
...
+                    success
+                }
+                graph {
+                    countNodes
+                    hasNode(name: 1)
+                }
+                flush
+            }
+        }";
+        client.query(&query, HashMap::new()).await;
```
fabubaker: There's a compiler warning here about an unused `Result`, please silence it.

fabubaker: Ditto for the other tests.

raphtory-graphql/Cargo.toml
```diff
@@ -35,7 +35,7 @@ dynamic-graphql = { workspace = true }
 async-graphql-poem = { workspace = true }
 futures-util = { workspace = true }
 opentelemetry = { workspace = true }
-opentelemetry_sdk = { workspace = true }
+opentelemetry_sdk = { workspace = true, features = ["testing"] }
```
fabubaker: This needs to be moved to `dev-dependency`

raphtory-graphql/Cargo.toml
```diff
@@ -71,8 +71,12 @@ tempfile = { workspace = true }
 pretty_assertions = { workspace = true }
 arrow = { workspace = true }
 parquet = { workspace = true }
+# Self-reference lets integration tests compile this crate with the `integration-test` feature enabled.
```
fabubaker:
```suggestion
# Self-reference lets cargo test compile this crate with the `integration-test` feature enabled.
```

raphtory-graphql/tests/open_telemetry.rs
```diff
@@ -0,0 +1,164 @@
+use raphtory::db::api::storage::storage::Config;
```
fabubaker, line 2: Can you add a comment at the top of this file explaining why these OTEL tests have to be isolated into a separate binary? i.e. registry sets the logging dispatch globally across the process.

raphtory-graphql/src/server.rs
```diff
@@ -274,20 +274,16 @@ impl GraphServer {
         let registry = Registry::default().with(filter).with(
             fmt::layer().pretty().with_span_events(FmtSpan::NONE), //(FULL, NEW, ENTER, EXIT, CLOSE)
         );
-        match tp.clone() {
-            Some((span, log)) => {
+        let dispatch = match tp.clone() {
```
fabubaker: We've isolated the log dispatch to per thread, but because we spawn rayon threads to process requests internally, those new threads won't have the same dispatch. So we have to revert back to using `try_init`.

fabubaker (changes requested):
Some more changes plus I think we can add an e2e test using a mock HTTP server in this PR too...

raphtory-graphql/src/config/otlp_config.rs
```diff
@@ -109,6 +121,19 @@ pub const DEFAULT_OTLP_TRANSPORT_PROTOCOL: TracingProtocol = TracingProtocol::TO
 pub const DEFAULT_OTLP_AGENT_PORT_TONIC: u16 = 4317;
 pub const DEFAULT_OTLP_TRACING_SERVICE_NAME: &'static str = "Raphtory";
 
+#[cfg(feature = "integration-test")]
+#[derive(Clone)]
+pub struct GlobalExporters {
```
fabubaker, line 121: You can put this comment back here: 

> Export in-memory exporters to retrieve spans and logs in tests.

raphtory-graphql/src/config/otlp_config.rs
```diff
@@ -279,3 +343,8 @@ impl TracingConfig {
         }
     }
 }
+
```
fabubaker: There's a bunch of stray trailing newlines here.

If you add these to your VSCode `settings.json` it'll automatically trim these:

```
{
  "files.trimFinalNewlines": true,
  "files.insertFinalNewline": true,
  "files.trimTrailingWhitespace": true
}
```

raphtory-graphql/src/server.rs
```diff
@@ -477,6 +477,7 @@ async fn server_termination(
     match tp {
         None => {}
         Some((tp, lp)) => {
+            #[cfg(not(feature="integration-test"))]
```
fabubaker: Are we gating this so that spans/logs don't get cleared out when the server shuts down? Might want to leave a comment explaining the gating.

raphtory-graphql/tests/open_telemetry_e2e.rs
```diff
@@ -0,0 +1,100 @@
+use mock_collector::{MockServer, Protocol};
```
fabubaker, line 1: I would prefer naming the file `open_telemetry_e2e.rs`

raphtory-graphql/src/config/otlp_config.rs
```diff
@@ -109,6 +121,19 @@ pub const DEFAULT_OTLP_TRANSPORT_PROTOCOL: TracingProtocol = TracingProtocol::TO
...
 
+#[cfg(feature = "integration-test")]
+#[derive(Clone)]
+pub struct GlobalExporters {
+    pub span: InMemorySpanExporter,
+    pub log: InMemoryLogExporter,
+}
+
+#[cfg(feature = "integration-test")]
+pub static GLOBAL_EXPORTERS: LazyLock<GlobalExporters> = LazyLock::new(|| GlobalExporters {
```
fabubaker, line 135: Please add this comment explaining why this is global:

> GraphServer registers span and log exporters across the entire process, which can conflict when starting up servers with their own exporters. Making in-memory exporters global allows them to be initialized once and reused across multiple tests.

raphtory-graphql/tests/open_telemetry.rs
```diff
@@ -0,0 +1,169 @@
...
+		])
+	);
+
+	let emitted_logs = log_exporter.get_emitted_logs().unwrap();
+	assert!(!emitted_logs.is_empty());
+}
+
+#[tokio::test]
+async fn test_open_telemetry_spans() {
+	test_open_telemetry_spans_complete().await;
```
fabubaker: Mention why these tests need to be run sequentially.

raphtory-graphql/tests/open_telemetry.rs
```diff
@@ -0,0 +1,169 @@
...
+async fn setup_for_span_tests(
+	tracing_level: TracingLevel,
+) -> (
+	RaphtoryGraphQLClient,
+	RunningGraphServer,
+	InMemorySpanExporter,
+	InMemoryLogExporter,
+	TempDir,
+) {
+    let span_exporter = GLOBAL_EXPORTERS.span.clone();
```
fabubaker, line 42: Explicitly mention why we're resetting the exporters here.

raphtory-graphql/src/server.rs
```diff
@@ -477,6 +480,11 @@ async fn server_termination(
     match tp {
         None => {}
         Some((tp, lp)) => {
+            /* avoid shutting down tracing exporters on server shutdown during integration tests
+               since the tracing exporters will be used for next test because tracing exporters
+               are global
```
fabubaker:
```suggestion
            /* Avoid shutting down global tracing exporters on server shutdown during integration tests
               since they are reused across multiple tests.
```

raphtory-graphql/tests/open_telemetry.rs
```diff
@@ -0,0 +1,179 @@
+// tests must be separated into their own binary to avoid conflicts with initializing the tracing subscriber multiple times in the same process
```
fabubaker:
```suggestion
// OpenTelemetry Tests must be separated into their own binary to prevent polluting other tests since the span and log exporters are set globally.
```

raphtory-graphql/tests/open_telemetry.rs
```diff
@@ -0,0 +1,179 @@
...
+    );
+
+    let emitted_logs = log_exporter.get_emitted_logs().unwrap();
+    assert!(!emitted_logs.is_empty());
+}
+
+#[tokio::test]
+async fn test_open_telemetry_spans() {
+    // All tests use the same exporters so all tests need to be ran sequentlially to avoid conflicts
+    // with spans and logs from all tests being mixed together on each test's span/log assertions.
```
fabubaker:
```suggestion
    // The following tests share the same global in-memory exporters and hence need to be run sequentially to prevent spans and logs from getting mangled.
```

fabubaker (approved):
LGTM besides some minor edits on the comments.

PR #2672, opened 2026-07-01
fix: gate read-only Graph Drop and flush against on-disk writes

shivamka1:
## Summary

Read-only `Graph.load(path, read_only=True)` handles were writing to the
graph directory on drop, corrupting a live writer's crash recovery:

- **Bug #1:** `Drop for GraphStore` appended a shutdown checkpoint to the
  shared WAL at LSN 256, clobbering the writer's live records.  On next
  reopen after a writer crash, recovery failed with
  `Expected checkpoint at given LSN`.
- **Bug #2:** `Drop for Storage` rewrote `.meta` even when the reader's
  snapshot was stale, rolling the cached counts backwards.

Fix: add a `read_only: bool` field to `GraphStore` and `Storage`, set by
`load_read_only`, and short-circuit both `Drop` impls (plus `flush()`)
when the flag is true.  Read-only handles are now byte-passive with
respect to the graph directory.

- `GraphStore` gets `load_read_only` + Drop/flush guards
- `Storage` gets a matching flag + Drop guard for the `.meta` refresh
- `TemporalGraph::load_inner` threads `read_only=true` through to
  `Layer::load_read_only`
- New `StorageError::ReadOnly` variant, returned by `flush()` on a
  read-only handle
- In-process `Storage::read_only()` also sets the flag so it doesn't
  double-refresh `.meta` alongside the original writer

Bug #3 (reader `Graph.load` racing writer segment-file creation, ENOENT
at `disk_layer_stats/mod.rs:301`) is a separate, unrelated race in the
reader-open path; it's documented and covered by an `xfail` regression
test in this PR so a future fix will trip the sentinel.

## Test plan

New file: `python/tests/test_base_install/test_graphql/test_read_only_durability.py`

- [x] CONTROL — writer + SIGKILL with no readers recovers cleanly
- [x] Bugs #1/#2 — SIGSTOP the writer, snapshot every file in the graph
      dir, open+drop a read-only handle, snapshot again; assert
      byte-identical.  Fails on `main`, passes on this branch.
- [x] Bug #3 (`xfail`) — 4 reader threads + live writer; asserts no
      `disk_layer_stats/mod.rs` ENOENT.  Still fails; xfail tracker for
      the follow-up.
- [x] All 11 existing read-only tests
      (`test_read_only_load.py`, `test_read_only_in_memory.py`) pass
      unchanged.


fabubaker (commented):
I'm in favour of closing this PR given that we're deprecating read-only loads and the guard abstractions here solve problems specific to read-only...

PR #2673, opened 2026-07-02
Fix 404 "not found" when fetching non-root paths with --public-dir enabled

louisch:

Set the index file to "index.html" for the `--public-dir` command of raphtory-graphql's CLI

poem's fallback_to_index() only falls back to the configured index file, but none was actually configured, so direct navigation to any UI subpage (e.g. /graphs) returned 404 "not found" instead of index.html when using the --public-dir option (when using the server .

No, it's probably expected behaviour.

Unit tests were added in routes.rs for PublicFilesEndpoint, and a test was added to server.rs which tests the server with AppWithConfig.with_public_dir enabled.

No


fabianmurariu (approved):
LGTM

PR #2674, opened 2026-07-02
Graphql loaders

wyatt-joyner-pometry:
Add load_nodes_from_parquet and load_edges_from_parquet GraphQL endpoints. It is specifically load_nodes_from_parquet rather than load_nodes, because I only written for this capacity at first.

Note that I am also not allowing shared_metadata, since we'll have to come up with a format for how to interpret types.

Claude mostly wrote the AppConfig changes for the allowed_parquet_paths, but there are tests for negative and positive cases of writing to allowed parquet paths. 

We need endpoints for loading parquet files over GraphQL.

I've copied over the descriptions from the corresponding Python API.

Added tests alongside the other existing GraphQL tests.

add support for shared_metadata
add support for other input formats like csv and raw blob
(more generally) add support for column name-mapping and casting


raphtory-graphql/src/model/mod.rs
```diff
@@ -42,14 +44,81 @@ use raphtory::{
...
+            let prop_type = type_str
+                .parse::<PropType>()
+                .map_err(|e| format!("Column '{col}': {e}"))?;
+            Ok((col, prop_type))
+        })
+        .collect::<Result<HashMap<_, _>, _>>()
+        .map(Some)
+}
+
+/*
```
fabianmurariu: remove comment

raphtory-graphql/src/model/mod.rs
```diff
@@ -179,6 +250,7 @@ impl QueryRoot {
     }
 
     /// Returns a graph
+
```
fabianmurariu: why new line here? and everywhere else?

raphtory-graphql/src/model/mod.rs
```diff
@@ -372,7 +451,182 @@ impl Mut {
...
+            desc = "A value to use as the node type for all nodes. Cannot be used in combination with node_type_col."
+        )]
+        node_type: Option<String>,
+        #[graphql(
+            desc = "The node type column name in a dataframe. Cannot be used in combination with node_type."
+        )]
+        node_type_col: Option<String>,
+        #[graphql(desc = "List of node property column names.")] properties: Option<Vec<String>>,
+        #[graphql(desc = "List of node metadata column names.")] metadata: Option<Vec<String>>,
+        // #[graphql(desc = "A dictionary of metadata properties that will be added to every node.")] shared_metadata: Option<String>,
```
fabianmurariu: remove comment

raphtory-graphql/src/model/mod.rs
```diff
@@ -372,7 +451,182 @@ impl Mut {
...
+        #[graphql(desc = "Path to the parquet directory.")] data_path: String,
+        #[graphql(desc = "The column name for the update timestamps.")] time: String,
+        #[graphql(desc = "The column name for the source node IDs.")] src: String,
+        #[graphql(desc = "The column name for the destination node IDs.")] dst: String,
+        #[graphql(desc = "List of edge property column names. Defaults to None.")]
+        properties: Option<Vec<String>>,
+        #[graphql(desc = "List of edge metadata column names. Defaults to None.")] metadata: Option<
+            Vec<String>,
+        >,
+        // #[graphql(desc = "A dictionary of metadata properties that will be added to every edge.")] shared_metadata: Option<String>,
```
fabianmurariu: remove comment

raphtory-graphql/src/model/mod.rs
```diff
@@ -372,7 +451,182 @@ impl Mut {
...
+        let properties: Vec<&str> = properties_owned.iter().map(String::as_str).collect();
+
+        let metadata_owned = metadata.unwrap_or_default();
+        let metadata: Vec<&str> = metadata_owned.iter().map(String::as_str).collect();
+
+        let schema = parse_json_schema(schema.as_deref()).map_err(GqlGraphError::LoadError)?;
+
+        // extracting PathBuf handles Strings too
+        let data_path = PathBuf::from(data_path);
+        let is_parquet = is_parquet_path(&data_path)?;
```
fabianmurariu: I feel like this might not be needed since load_x_from_parquet already will fail if given non-parquet things

raphtory-api/src/core/entities/properties/prop/prop_type.rs
```diff
@@ -206,6 +206,15 @@ pub fn data_type_as_prop_type(dt: &DataType) -> Result<PropType, InvalidProperty
 #[error("{0:?} not supported as property type")]
 pub struct InvalidPropertyTypeErr(pub DataType);
 
+impl FromStr for PropType {
+    type Err = String;
```
fabianmurariu: We should use an error type here, so we don't have to call map_err below

raphtory-graphql/src/model/mod.rs
```diff
@@ -42,14 +44,81 @@ use raphtory::{
...
+            entry.map_or(false, |e| {
+                e.path().extension().and_then(OsStr::to_str) == Some("parquet")
+            })
+        }))
+    } else {
+        Ok(path.extension().and_then(OsStr::to_str) == Some("parquet"))
+    }
+}
+
+pub(crate) fn is_csv_path(path: &PathBuf) -> Result<bool, std::io::Error> {
```
fabianmurariu: is this used? even so I'd get rid of it as we don't care

raphtory-graphql/src/lib.rs
```diff
@@ -2602,4 +2602,348 @@ mod graphql_test {
...
+            .unwrap();
+
+        let schema = App::create_schema().data(data).finish().unwrap();
+
+        let parquet_path_str = parquet_path.to_str().unwrap().replace('\\', "/");
+        let mutation = format!(
+            r#"mutation {{
+                loadEdgesFromParquet(
+                    graphPath: "mygraph",
+                    dataPath: "{}",
```
fabianmurariu, line 2754: why is dataPath "{}" ? isn't it supposed to be a string?

raphtory-graphql/src/lib.rs
```diff
@@ -2602,4 +2602,348 @@ mod graphql_test {
             _ => None,
         }
     }
+
```
fabianmurariu, line 2605: can we add a test with the schema being parsed and one where the  parquet path is nested inside the allowed parquet config directory

raphtory-graphql/src/lib.rs
```diff
@@ -2602,4 +2602,348 @@ mod graphql_test {
...
+            .unwrap();
+
+        let schema = App::create_schema().data(data).finish().unwrap();
+
+        let parquet_path_str = parquet_path.to_str().unwrap().replace('\\', "/");
+        let mutation = format!(
+            r#"mutation {{
+                loadNodesFromParquet(
+                    graphPath: "mygraph",
+                    dataPath: "{}",
```
fabianmurariu, line 2663: could this even be optional?

fabianmurariu (changes requested):
LGTM but There are a few changes needed here and there

raphtory-graphql/src/lib.rs
wyatt-joyner-pometry: It's a format string. We're passing in parquet_path_str.

raphtory-graphql/src/lib.rs
wyatt-joyner-pometry: Done

raphtory-api/src/core/entities/properties/prop/prop_type.rs
wyatt-joyner-pometry: done.

PR #2681, opened 2026-07-16
CLI parsing tests for `raphtory-server`

DanielLacina:
Added cli-parsing tests that test the priority of sources to parse arguments from. 
To verify that arguments are parsed correctly.
No
These are tests.
No


raphtory-graphql/src/cli.rs
```diff
@@ -1,28 +1,19 @@
...
+        }, log_config::DEFAULT_LOG_LEVEL, otlp_config::{
+            DEFAULT_OTLP_TRACING_SERVICE_NAME, DEFAULT_OTLP_TRANSPORT_PROTOCOL, DEFAULT_TRACING_ENABLED, DEFAULT_TRACING_LEVEL, TracingLevel, TracingProtocol,
+        }, schema_config::DEFAULT_DISABLE_INTROSPECTION,
+    }, model::App, server::{DEFAULT_PORT, apply_server_extension},
 };
 use clap::{Parser, Subcommand};
 use raphtory::db::api::storage::storage::Config;
 use std::{collections::HashMap, io, path::PathBuf};
 use tokio::io::Result as IoResult;
+use serde_json::Value;
```
fabubaker: Unused import?

raphtory-graphql/src/cli.rs
```diff
@@ -48,7 +39,7 @@ enum Commands {
     #[command(about = "Print the GraphQL schema")]
     Schema,
 }
-#[derive(clap::Args, Debug)]
+#[derive(clap::Args, Debug, serde::Serialize)]
 struct ServerArgs {
```
fabubaker:
```suggestion
#[derive(Args, Debug, Serialize)]
struct ServerArgs {
```

Prefer non-qualified imports.

raphtory-graphql/src/cli.rs
```diff
@@ -352,3 +351,73 @@ pub fn python_cli() -> pyo3::PyResult<()> {
         .block_on(cli_with_args(args))
         .map_err(|err| pyo3::exceptions::PyIOError::new_err(err.to_string()))
 }
+
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::path::PathBuf;
```
fabubaker: Unused import.

raphtory-graphql/src/cli.rs
```diff
@@ -352,3 +351,73 @@ pub fn python_cli() -> pyo3::PyResult<()> {
...
+        )
+        .expect("failed to write temporary cache config");
+        config_file
+            .flush()
+            .expect("failed to flush temporary cache config");
+        config_file
+    }
+
+    async fn test_cli_parsing_no_arguments() {
+        let args: Vec<String> = vec![r"target\\debug\\raphtory-server".to_string(), "server".to_string()];
```
fabubaker: I don't think you need the `to_string` here and can just pass a `Vec<&str>`?

raphtory-graphql/src/cli.rs
```diff
@@ -352,3 +351,73 @@ pub fn python_cli() -> pyo3::PyResult<()> {
...
+        std::env::remove_var("RAPHTORY_CACHE_CAPACITY");
+        let (_s, app_config) = generate_config(args).unwrap().unwrap();
+        assert!(app_config.cache.capacity == DEFAULT_CACHE_CAPACITY);
+    } 
+
+    async fn test_cli_parsing_with_config_file() {
+        let config_file = config_file();
+        let args: Vec<String> = vec![
```
fabubaker: Ditto here for the `Vec<&str>` here.

fabubaker: Plus the other tests as well.

raphtory-graphql/src/cli.rs
```diff
@@ -352,3 +351,73 @@ pub fn python_cli() -> pyo3::PyResult<()> {
...
+            .flush()
+            .expect("failed to flush temporary cache config");
+        config_file
+    }
+
+    async fn test_cli_parsing_no_arguments() {
+        let args: Vec<String> = vec![r"target\\debug\\raphtory-server".to_string(), "server".to_string()];
+        std::env::remove_var("RAPHTORY_CACHE_CAPACITY");
+        let (_s, app_config) = generate_config(args).unwrap().unwrap();
+        assert!(app_config.cache.capacity == DEFAULT_CACHE_CAPACITY);
```
fabubaker: Use `_` instead of `_s`, the former is the idomatic way of saying "I don't care about this variable" while the latter still creates a variable binding but just tells the compiler to ignore the unused warning.

raphtory-server/tests/cli_parsing.rs
```diff
@@ -0,0 +1,210 @@
...
+    process::{Command, Stdio},
+    time::Duration,
+};
+
+use raphtory_graphql::config::{app_config::AppConfig, cache_config::DEFAULT_CACHE_CAPACITY};
+use serde::Deserialize;
+use tempfile::Builder;
+use serde_json::Value;
+
+fn workspace_root() -> std::path::PathBuf {
```
fabubaker: This might blow up when we run these tests on the pometry-storage side, which doesn't have raphtory-graphql under the workspace root.

I'd suggest moving this integration to raphtory-server, you'll find it a lot easier to grab the binary that way.

fabubaker (changes requested):

raphtory-server/tests/cli_parsing.rs
```diff
@@ -0,0 +1,185 @@
...
+    write!(config_file, "[cache]\ncapacity = 123\n").expect("failed to write temporary cache config");
+    config_file
+        .flush()
+        .expect("failed to flush temporary cache config");
+    config_file
+}
+
+#[test]
+fn test_cli_parsing_no_arguments() {
+    let workspace_root = workspace_root();
```
fabubaker: There's no real benefit to setting `workspace_root` as the current_dir here, so I'd just remove remove it.

fabubaker (approved):

PR #2689, opened 2026-07-23
feat: edge neighbour sort

louisch:

- raphtory-graphql API for sorting edges by properties on the nodes (neighbours) they connect to.
- Adding `type` and `name` sorting to NodeSortBy in raphtory-graphql

In the UI I'd like to be able to do a sort of exploded edges by the neighbours that they connect to, then page. The only thing missing was a way for the *edge sort* to be able to sort by the *neighbours* that they connect to.

Yes, it's new API endpoints. Docs added above each endpoint function.

Tests added for type sorting, and sorting edges by their neighbours

No


raphtory-graphql/src/model/sorting.rs
```diff
@@ -8,6 +10,11 @@ pub struct EdgeSortBy {
     pub src: Option<bool>,
     /// Destination
     pub dst: Option<bool>,
+    /// Sort by the neighbour node (the endpoint that is NOT the node these
```
ljeub-pometry: Shouldn't these also take a `NodeSortBy`? Otherwise, you can't sort a graph-level edge collection by src...

ljeub-pometry (changes requested):
The `src` and `dst` options for edge sorting should also take a `NodeSortBy` (especially useful for graph-level edges)

ljeub-pometry (approved):

PR #2691, opened 2026-07-29
Fix dependencies and remove tantivy search

ljeub-pometry:

- fork pyo3-arrow so we vectors/node state work properly again
- remove the dead tantivy-based indexing code

- supported arrow version mismatch between lancedb/datafusion and pyo3-arrow

no

tests

once lancedb and datafusion are updated, move back to mainline pyo3-arrow


raphtory-graphql/src/cli.rs
```diff
@@ -208,10 +206,6 @@ struct ServerArgs {
     #[arg(long, env = "RAPHTORY_PERMISSIONS_STORE_PATH", default_value = None, help = "Path to the JSON permissions store file.")]
     permissions_store_path: Option<PathBuf>,
 
-    #[cfg(feature = "search")]
-    #[arg(long, env = "RAPHTORY_CREATE_INDEX", help = help_with_default!("Enable index creation.", DEFAULT_CREATE_INDEX))]
```
miratepuffin, line 212: Is there any conf we have accidentally left available?

raphtory-graphql/src/model/graph/vector_selection.rs
```diff
@@ -187,114 +181,3 @@ impl GqlVectorSelection {
...
-    /// Add every document associated with the named edges to the selection.
-    /// Documents added this way receive a score of 0 (no similarity ranking).
-
-    async fn add_edges(
-        &self,
-        #[graphql(desc = "List of `{src, dst}` pairs identifying the edges.")] edges: Vec<
-            InputEdge,
-        >,
-    ) -> Self {
-        unreachable!("vectors feature disabled")
```
miratepuffin, line 236: Have these been moved/enabled?

miratepuffin (approved):
Couple of minor questions:
The vector node selection in graphql should be available no?
Was there any search config that we have that we can remove
The nodestate functions are back now, but still behind the datafusion flag?

raphtory-graphql/src/cli.rs
ljeub-pometry: I don't see any (the indexing-related conf was handled by a separate sub-struct)

raphtory-graphql/src/model/graph/vector_selection.rs
ljeub-pometry: the no-op implementations are no longer needed as the entry points are now properly removed from the server when the `vectors` feature is not enabled

ljeub-pometry:
> Couple of minor questions: The vector node selection in graphql should be available no? Was there any search config that we have that we can remove The nodestate functions are back now, but still behind the datafusion flag?

- The vector selection is still available but only when the `vectors` feature was enabled at compile time
- Search config should be removed (it was in a separate struct)
- Nodestate functions are back (still behind `datafusion` feature which is enabled by default, compile with `--no-default-features` to disable them)

Issue #2695, opened 2026-07-31
Follow-ups for algorithms in GraphQL


arienandalibi:
### Missing algorithms:
`temporal_bipartite_projection`, `kCore`, and `kCoreSet`. They return graphs and it's possible that the graphs generated here are huge.

### Modified algorithm:
`temporal_rich_club_coefficient` can't take an iterator of `GraphView`s in GraphQL. Instead, we take arguments for creating a rolling window and call the algorithm using that. As such, only rolling windows can be used for this algorithm in GraphQL.

Issue #2694, opened 2026-07-31
Follow-ups for GqlNodeState


arienandalibi:
### Missing functionality from core Rust `NodeState`:
- Merging. `NodeState`s can't be merged in GraphQL. Passing another `NodeState` as argument to the merge function is awkward to express in GraphQL queries.
- Parquet serialization. `to_parquet` and `from_parquet` weren't added because they would require filesystem access on the server's machine, which could present a security risk.

### Re-implemented:
`sum` and `mean`. In Rust, the trait bounds ensure that the inputs to these are numerical/additive. In GraphQL, we can't use these trait bounds to ensure the `NodeState` has numerical values. As such, these functions were re-implemented and return `None` if the values aren't numerical/additive at runtime. 

### Still need to add:
- Paging. Right now, accessing the data in a NodeState is done using `columns`/`nodes`/`rows`. These dump the entire `NodeState`'s contents (columns, nodes, or rows respectively) in a single response, which can be a problem on big NodeStates. Ideally, we can page the results like with other queries.
- Testing for `GqlNodeStateValue::Node`. It is currently present in GraphQL to match core Rust, but no algorithm returns a single-node column, so it is unused/untested.

PR #2696, opened 2026-07-31
Adding algorithms to GraphQL

arienandalibi:
Exposing the algorithms in GraphQL. They are currently only available in core Rust and Python.

Accessing them in GraphQL.

Yes, many GraphQL queries. All documented.

Unit tests for each algorithm.

Yes, potential follow-ups on GqlNodeState ( #2694 ) and the algorithms ( #2695 )


raphtory-graphql/src/model/algorithms/mod.rs
```diff
@@ -0,0 +1,751 @@
...
+pub(crate) mod strongly_connected_components;
+pub(crate) mod temporal_rich_club_coefficient;
+pub(crate) mod temporal_seir;
+pub(crate) mod temporally_reachable_nodes;
+pub(crate) mod triangle_count;
+pub(crate) mod triplet_count;
+pub(crate) mod weakly_connected_components;
+
+/// A graph algorithm executable through the GraphQL API.
+pub(crate) trait GqlExecutableAlgorithm: 'static {
```
miratepuffin: There is a lot packed into this mod - I would split it out into different files

miratepuffin: Ideally docs would be sync'd with python docs somehow

raphtory-graphql/src/model/graph/filtering.rs
```diff
@@ -585,6 +585,20 @@ pub enum GqlGraphFilter {
     Layers(GraphLayersExpr),
 }
 
+/// A composite filter producing a graph view, bundling the graph-view, node,
```
miratepuffin, line 588: I have something similar to this in my PR, we will need to merge - I don't think you are support And/Or here?

raphtory-graphql/src/lib.rs
```diff
@@ -582,6 +582,2160 @@ mod graphql_test {
         graph
     }
 
+    #[tokio::test]
+    async fn test_algorithm_node_state_ops() {
```
miratepuffin: I would put all of the tests into the actual algorithms themselves, gonna be easier to see which ones haven't been tested then.

Might not be a terrible idea to sanity check in python by running on the local graph and then via graphql and seeing if we get the same result?

miratepuffin (changes requested):
Probably a good idea to split the algorithms folder into a similar structure as the actual algorithms folder

raphtory-graphql/src/model/graph/node_state.rs
```diff
@@ -0,0 +1,1044 @@
...
+
+/// One column of a node state: the values of a single output field of the
+/// algorithm. Row-aligned with `NodeState.nodes`.
+#[derive(SimpleObject, Clone)]
+#[graphql(name = "NodeStateColumn")]
+pub(crate) struct GqlNodeStateColumn {
+    /// Name of the column.
+    name: String,
+    /// The values of this column; `values[i]` belongs to `NodeState.nodes[i]`.
+    values: Vec<GqlNodeStateValue>,
```
ljeub-pometry, line 90: This isn't great, you need to reallocate a new vector to make this instead of reusing the underlying column!

ljeub-pometry: Also, SimpleObject is probably not the right thing to use here? Should be resolved with accessors for stuff like sum etc...

raphtory-graphql/src/model/graph/node_state.rs
```diff
@@ -0,0 +1,1044 @@
...
+
+impl Hash for GroupKey {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.repr.hash(state);
+    }
+}
+
+impl GqlNodeState {
+    /// A slice of `limit` rows starting at `start`, as a node state of its own.
+    fn slice(&self, start: usize, limit: usize) -> GqlNodeState {
```
ljeub-pometry, line 207: we should expose an api like this on the NodeState itself

raphtory-graphql/src/model/graph/node_state.rs
```diff
@@ -0,0 +1,1044 @@
...
+                        .map(|(_, value)| value.into())
+                        .collect(),
+                })
+                .collect()
+        })
+        .await
+    }
+
+    /// Returns the values for a node, one entry per column; null if the node has no value in this NodeState.
+    async fn get(
```
ljeub-pometry, line 345: shouldn't this return a `Option<GqlNodeStateRow>`?

raphtory-graphql/src/model/graph/node_state.rs
```diff
@@ -0,0 +1,1044 @@
...
+                    .into_iter_groups()
+                    .map(|(key, nodes)| GqlNodeStateGroup {
+                        value: key.value.map(GqlPropertyOutputVal),
+                        nodes: GqlNodes::new(nodes),
+                    })
+                    .collect(),
+            )
+        })
+        .await
+    }
```
ljeub-pometry, line 550: We need proper single-column apis in the rust layer. However, the ordering based apis here should work with multiple columns, same as in rust.

raphtory-graphql/src/model/graph/node_state.rs
```diff
@@ -0,0 +1,1044 @@
...
+        let self_clone = self.clone();
+        blocking_compute(move || GqlNodeState {
+            node_state: self_clone.node_state.sort_by_id(),
+        })
+        .await
+    }
+
+    /// The columns of the node state, one per output field of the algorithm.
+    /// `values` are row-aligned with `nodes`.
+    async fn columns(&self) -> Vec<GqlNodeStateColumn> {
```
ljeub-pometry, line 589: there should be `column(name)` as well to get a single column

raphtory-graphql/src/model/algorithms/alternating_mask.rs
```diff
@@ -16,3 +16,55 @@ impl GqlExecutableAlgorithm for GqlAlternatingMask {
         Ok(alternating_mask(graph).into())
     }
 }
+
```
miratepuffin, line 19: I don't think we need to expose this - its a test algorithm

raphtory-graphql/src/model/algorithms/executable.rs
```diff
@@ -0,0 +1,41 @@
...
+}
+
+impl GqlAlgorithms {
+    /// Runs algorithm `A` on the blocking thread pool.
+    pub(crate) async fn run<A: GqlExecutableAlgorithm>(
+        &self,
+        args: A::Args,
+    ) -> Result<A::Output, GraphError> {
+        let graph = self.graph.clone();
+        blocking_compute(move || A::execute(&graph, args)).await
```
miratepuffin, line 39: This is probably not an issue right now, but we should probably highly weight some of these algorithms in terms of graphql complexity?

miratepuffin (approved):

raphtory-graphql/src/model/algorithms/resolvers.rs
```diff
@@ -0,0 +1,637 @@
...
+        #[graphql(desc = "Number of threads to use. Defaults to all available.")] threads: Option<
+            usize,
+        >,
+        #[graphql(desc = "Convergence tolerance. Defaults to 0.000001.")] tol: Option<f64>,
+        #[graphql(desc = "Probability that the spread continues. Defaults to 0.85.")]
+        damping_factor: Option<f64>,
+        #[graphql(desc = "Edge property to use as weight. If unset, all edges have weight 1.")]
+        weight: Option<String>,
+    ) -> Result<GqlNodeState, GraphError> {
+        self.run::<GqlPagerank>(GqlPagerankArgs {
```
ljeub-pometry, line 114: what is the point in the whole executable machinery when you are still writing all these functions here anyway? You could have e.g., just called `blocking_compute(|| {pagerank()).await` here?

ljeub-pometry, line 114: This whole executable machinery seems to be entirely useless? You could just change the run method to take a closure that has a DynamicGraph as input and a generic as output and have everything inline here without all the extra code for each algorithm!

ljeub-pometry (changes requested):
I think all the extra abstraction and code for each algorithm isn't actually needed?

PR #2697, opened 2026-07-31
Fix release

ljeub-pometry:

Moves graphql integration tests to avoid a self-dependency which breaks cargo publish


miratepuffin (approved):

raphtory-graphql-tests/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "raphtory-graphql-tests"
+version.workspace = true
```
fabubaker, line 5:
```suggestion
# Contains tests that use the `integration-test` feature flag on raphtory-graphql.
[package]
name = "raphtory-graphql-tests"
version.workspace = true
```

just to be explicit on why this crate exists...

PR #2698, opened 2026-07-31
graphql: flush mutations rewrite the metadata counts sidecar

louisch:

There was a bug where when calling updateGraph would not update the metadata

Both flush entry points — `updateGraph{ flush }` and the root `flush(graphPath:)` — flushed the storage engine but never rewrote the metadata file, and by clearing the dirty flag they disarmed the eviction path that otherwise would have. The metadata stayed frozen at the creation-time 0/0, so cache-miss namespace listings reported nodeCount/edgeCount 0 while the same row's lastUpdated and metadata were fresh.

Extract the flush-and-persist sequence into `GraphWithVectors::persist` (engine flush + `replace_graph_data`, both attempted, first error returned) and route the eviction-time `cache::flush_graph` and both flush mutations through it. Each caller keeps its own dirty-on-error policy, so the eviction/shutdown behaviour is unchanged; the only behavioural change is that the flush mutations now write the .meta file.

The new persist method does not call set_dirty on errors because flush_graph does not do so on errors, while flush does.

This was apparently also fixed in disk graph by arien? But it does seem to be present in OS raphtory, as the e2e test fails in `db_v4` right now

There was a bug where when calling updateGraph would not update the metadata

Some unit tests and one e2e test testing newGraph and updateGraph actually update the metadata

For newGraph and updateGraph since the e2e test verifies their behavior now there shouldn't be...

### DIFF
--- a/raphtory-graphql/src/model/mod.rs
+++ b/raphtory-graphql/src/model/mod.rs
@@ -15,7 +15,7 @@ use crate::{
         },
     },
     paths::{ExistingGraphFolder, ValidGraphPaths, ValidWriteableGraphFolder},
-    rayon::blocking_compute,
+    rayon::{blocking_compute, blocking_write},
     url_encode::{url_decode_graph_at, url_encode_graph},
 };
 use async_graphql::Context;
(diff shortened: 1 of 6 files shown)


raphtory-graphql/src/model/mod.rs
```diff
@@ -873,10 +873,15 @@ impl Mut {
         let data = ctx.data_unchecked::<Data>();
         let graph = data
             .get_graph_with_write_permission(ctx, &graph_path)
-            .await?
-            .graph()
-            .clone();
-        graph.flush()?;
+            .await?;
```
ljeub-pometry, line 876: wait, why does this even exist if we also have a flush on the MutableGraph endpoint?

ljeub-pometry (approved):
I like the simplification of the internal apis.

PR #2707, opened 2026-08-10
Remote py api tests

shivamka1:
Adds a differential test suite that runs the same code against a local `Graph` and a `RemoteGraph` and asserts they agree. The premise of `RemoteGraph` is that it is a drop-in replacement, which gives these tests a free oracle: one `build(g)` seeds both sides, one call runs on both, and the local graph is ground truth.

18 files, ~1580 cases, plus Hypothesis property tests.

### How it works

- **`graph_pair`** seeds a local graph and a server-backed remote one identically and yields both.
- **`canonical` / `assert_parity`** reduce a result to a comparable form and assert local == remote, including exception parity (both sides must raise, and raise the same type). The comparator bridges only the ways the two sides are unavoidably *different objects* — entity handles become identities, unordered entity collections are sorted — never a difference in the answers themselves.
- **`KNOWN_GAPS`** is the divergence ledger. An API not yet at parity is recorded with a reason and xfailed rather than silently skipped, so the suite doubles as an executable parity spec that shrinks as gaps close.

Coverage spans views, filters, writes (including a deep write-sequence file), properties, collections, `collect()` materialization, sorting, ids, types and RPC counts.

### Product bugs this surfaced

The suite was not only test code — it found real defects, each fixed here:

- **Integer node ids were broken over the wire.** The client stringified every node id before sending it, so a graph created with integer ids became string-indexed on the server. `node(5).id` came back as `'5'` instead of `5`, and the graph's own id-type enforcement disappeared. Node ids are now `GID` end to end and render as the server's typed `NodeId` scalar.
- **`edge.delete(t)` on an unlayered edge** behaved differently from `graph.delete_edge(t, src, dst)`: the no-layer branch of `resolve_layer` looked `_default` up but never created it, so the same intent spelled two ways diverged.
- **Non-finite floats, map ordering and property dtypes** are now pinned per carrier rather than assumed.

### Behaviour change

Writing an **integer node id into a string-indexed graph** (or the reverse) now raises on the remote side. Previously the client stringified the id, so the write silently succeeded against a string-indexed graph.

This matches the local API exactly — `Graph` has always raised `Storage error: Node id 1 does not have the correct type, expected String` for this — but code written against the older remote behaviour will now see an error where it previously saw a successful write.

### Client changes made for parity

- Node ids are `GID` throughout, rendered as the `NodeId` scalar (typed, not stringified).
- Write methods take properties the way local `AdditionOps` does — any `IntoIterator` of `(key, value)` — instead of requiring an exact `HashMap<String, Prop>`. `NO_PROPS` means none.
- Graph semantics use the schema's `GraphType` enum rather than a validated string.
- Write times use the `TryIntoInputTime` mechanism the local mutation API uses.

### Follow-ups filed

- #2716 — RemoteGraph has no base-graph vs view distinction
- #2717 — add `sorted()` to the local Nodes/Edges


db4-storage/src/pages/mod.rs
```diff
@@ -426,6 +426,12 @@ impl<
                     return;
                 }
             }
+            // A missing backing directory means the store's dir was already torn
+            // down (e.g. a temporary graph whose dir was removed first) — there is
+            // nothing left to persist, so skip quietly rather than aborting under
+            // panic-on-drop. Real flush failures, where the dir still exists, still
+            // surface loudly.
+            Err(err) if err.is_missing_path() => {}
```
ljeub-pometry: if your code is deleting a graph that is still in use, that is likely a problem!

db4-storage/src/lib.rs
```diff
@@ -139,6 +139,43 @@ pub mod error {
...
         }
+
+        /// True when this error is ultimately "the path does not exist" (ENOENT).
+        ///
+        /// Flush-on-drop paths use this to tolerate a backing directory that was
+        /// already removed (e.g. a temporary store whose directory was torn down
+        /// before the store itself dropped): there is nothing left to persist, so
+        /// the drop becomes a no-op instead of a hard failure. Genuine flush
+        /// errors — where the directory still exists — are unaffected.
+        pub fn is_missing_path(&self) -> bool {
```
ljeub-pometry: this needs to go, note that this silences way more errors than it claims in the docs, any missing file error gets silenced by this, not just the case where the entire directory was deleted!

python/tests/test_base_install/test_graphql/parity/_parity.py
```diff
@@ -0,0 +1,202 @@
...
+    build(local)
+    # TemporaryDirectory (outer) is torn down only after the server context
+    # (inner) has stopped and flushed — so the dir outlives every write-back and
+    # is then cleaned up, rather than leaked as `mkdtemp` would.
+    with tempfile.TemporaryDirectory() as work_dir:
+        with GraphServer(work_dir).start() as server:
+            client = server.get_client()
+            client.new_graph("g", graph_type)
+            remote = client.remote_graph("g")
+            build(remote)
```
ljeub-pometry: this should just be `remote = client.new_graph("g", graph_type)`

python/tests/test_base_install/test_graphql/parity/_parity.py
```diff
@@ -0,0 +1,202 @@
...
+
+@contextlib.contextmanager
+def graph_pair(build, graph_type="EVENT"):
+    """Seed a local ``Graph`` and a ``RemoteGraph`` with ``build``, yield both.
+
+    ``build`` takes a single graph handle and applies writes using only the
+    shared (drop-in) surface, so the exact same callable runs against each side.
+    The server is started on enter and torn down on exit.
+    """
+    local = Graph()
```
ljeub-pometry: this needs to handle `graph_type` here if you want that to work properly, if this is passed `"PERSISTENT"`, the local graph also needs to be a `PersistentGraph`

python/tests/test_base_install/test_graphql/parity/_parity.py
```diff
@@ -0,0 +1,202 @@
...
+            client = server.get_client()
+            client.new_graph("g", graph_type)
+            remote = client.remote_graph("g")
+            build(remote)
+            yield GraphPair(local=local, remote=remote)
+
+
+# --- canonicalization -------------------------------------------------------
+
+# Float comparison tolerance: serde round-trips can perturb the last bits.
```
ljeub-pometry: I thought we fixed the float round-trips in serde by enabling some features?

python/tests/test_base_install/test_graphql/parity/_parity.py
```diff
@@ -0,0 +1,202 @@
...
+_FLOAT_PLACES = 9
+
+
+def canonical(value):
```
ljeub-pometry, line 108: this needs to be handled by the apis, not the tests, ideally everything needs to roundtrip properly

ljeub-pometry: the node/edge comparison by id are ok

python/tests/test_base_install/test_graphql/parity/_parity.py
```diff
@@ -0,0 +1,202 @@
...
+    """Normalize a raphtory result into a stable, comparable Python value.
+
+    Absorbs the drift that would otherwise cause false diffs between an
+    in-process result and one that crossed the wire: collection ordering, float
+    precision, and datetime timezone. Entities are reduced to identity
+    (node name, edge ``(src, dst)``) so the two sides compare structurally.
+    """
```
ljeub-pometry: basically, no drift is allowed, the only thing this function is allowed to do is compare nodes and edges by id as these would not be expected to compare equal (they are objects pointing at different graphs after all)

python/tests/test_base_install/test_graphql/parity/_parity.py
```diff
@@ -0,0 +1,202 @@
...
+    ), f"value parity mismatch: local={local_result!r} remote={remote_result!r}"
+
+
+# --- divergence ledger ------------------------------------------------------
+
+# Known local↔remote gaps (see docs/remote-graph-api parity notes). A
+# parametrized case tagged with one of these keys is xfailed with its reason,
+# so the gap is *recorded* rather than silently skipped. Delete an entry when
+# the corresponding remote API lands.
+KNOWN_GAPS = {
```
ljeub-pometry, line 195: anything that is currently being handled by `canonical` should instead be here and tagged with an issue

python/tests/test_base_install/test_graphql/parity/_parity.py
```diff
@@ -0,0 +1,202 @@
...
+        "node-view-op plus GraphView, so Edges has no filter method"
+    ),
+    "filter.edge.filter": (
+        "RemoteEdge.filter has no local counterpart (as filter.edges.filter)"
+    ),
+    "filter.nested_edges.filter": (
+        "RemoteNestedEdges.filter has no local counterpart (as " "filter.edges.filter)"
+    ),
+    "filter.collection.select": (
+        "select() — the remote's explicit narrow-here-only form — has no local "
```
ljeub-pometry: this syntax should be the same?

python/tests/test_base_install/test_graphql/parity/_strategies.py
```diff
@@ -0,0 +1,401 @@
...
+from raphtory import filter as f
+
+# --- pools --------------------------------------------------------------------
+
+NODE_NAMES = ["n0", "n1", "n2", "n3", "n4", "n5"]
+LAYERS = ["alpha", "beta", "gamma"]
+NODE_TYPES = ["person", "bot", "org"]
+
+# One value type per key, values chosen so equality comparisons can bite.
+PROP_POOLS = {
```
ljeub-pometry: not very comprehensive in terms of data types...

python/tests/test_base_install/test_graphql/parity/_strategies.py
```diff
@@ -0,0 +1,401 @@
...
+_times = st.integers(min_value=0, max_value=12)
+# Explicit event ids sit above any auto-assigned id a short sequence produces
+# (autos count writes from 0), so explicit and auto ids never collide.
+_event_ids = st.sampled_from([None, 50, 51, 52])
+_names = st.sampled_from(NODE_NAMES)
+_maybe_layer = st.sampled_from([None] + LAYERS)
+
+
+@st.composite
+def _props(draw, pools=PROP_POOLS, max_size=3):
```
ljeub-pometry: you might be able to take some inspiration from the rust-side recursive props generator to make lists/maps more interesting

python/tests/test_base_install/test_graphql/parity/_strategies.py
```diff
@@ -0,0 +1,401 @@
...
+    keys = draw(
+        st.lists(st.sampled_from(sorted(pools)), unique=True, max_size=max_size)
+    )
+    return {key: draw(st.sampled_from(pools[key])) for key in keys}
+
+
+# --- write operations -----------------------------------------------------------
+
+# Each op is a tuple whose first element names the call; `apply_op` dispatches.
+_ADD_NODE = st.tuples(
```
ljeub-pometry: no private prefix?

python/tests/test_base_install/test_graphql/parity/_strategies.py
```diff
@@ -0,0 +1,401 @@
...
+
+def apply_op(g, op):
+    """Apply one generated op to a graph handle (local or remote).
+
+    Entity-scoped ops are guarded on existence: ``g.node(...)`` is ``None`` on
+    both sides for an absent node, and the guard decision is a function of the
+    op prefix alone, so both sides always skip (or apply) in lockstep.
+    """
+    tag = op[0]
+    if tag == "add_node":
```
ljeub-pometry, line 353: could be an idea to keep this in a global dict and make a function to create a new strategy and its corresponding operation to keep this in sync

python/tests/test_base_install/test_graphql/parity/_strategies.py
```diff
@@ -0,0 +1,401 @@
...
+@st.composite
+def _edge_prop_leaf(draw):
+    key = draw(st.sampled_from(_NUMERIC_KEYS + ["p_str"]))
+    if key == "p_str":
+        op = draw(st.sampled_from(_STRING_OPS + ["is_some", "is_none"]))
+    else:
+        op = draw(st.sampled_from(_CMP_OPS + ["is_in", "is_some", "is_none"]))
+    value = draw(_leaf_value(key, op))
+    return ("eprop", key, op, value)
+
```
ljeub-pometry, line 464: shouldn't node and edge properties support the same filters?

python/tests/test_base_install/test_graphql/parity/_strategies.py
```diff
@@ -0,0 +1,401 @@
...
+
+def _apply_comparison(target, op, value):
+    if op in ("eq", "ne", "lt", "le", "gt", "ge"):
+        return getattr(operator, op)(target, value)
+    if op == "is_some":
+        return target.is_some()
+    if op == "is_none":
+        return target.is_none()
+    return getattr(target, op)(value)  # contains / starts_with / ends_with / is_in
```
ljeub-pometry: if you use the internal names (such as `__eq__`, etc.) and pass in an empty tuple for the value for `is_some` and `is_none`, you only need the last line of code here...

python/tests/test_base_install/test_graphql/parity/test_parity_entities.py
```diff
@@ -0,0 +1,100 @@
...
+    assert_parity(rich_pair, fn)
+
+
+# --- PathFromNode (traversal) ----------------------------------------------
+
+PATH_READS = [
+    ("neighbours_names", lambda g: sorted(n.name for n in g.node("a").neighbours)),
+    # `.count()` is a remote-only extra (local PathFromNode has none) — count via
+    # iteration, which both sides support.
+    ("neighbours_count", lambda g: sum(1 for _ in g.node("a").neighbours)),
```
ljeub-pometry, line 98: hmm, we should add `count` locally...

python/tests/test_base_install/test_graphql/parity/test_parity_filters.py
```diff
@@ -0,0 +1,1297 @@
...
+#
+# The facts are also chosen to be sensitive to both things a filter can change:
+# *membership* (which entities remain) and *scope* (what each survivor answers,
+# since a collection filter narrows the view its members report through rather
+# than the collection itself).
+
+
+def _t(value):
+    """Reduce a time-ish value to its timestamp (remote yields ``EventTime``)."""
+    return None if value is None else getattr(value, "t", value)
```
ljeub-pometry: this should be the same type on both sides

python/tests/test_base_install/test_graphql/parity/test_parity_filters.py
```diff
@@ -0,0 +1,1297 @@
...
+        "layers": sorted(e.layer_names),
+        "earliest": _t(e.earliest_time),
+        "latest": _t(e.latest_time),
+        "valid": e.is_valid(),
+        "deleted": e.is_deleted(),
+    }
+
+
+def _edge_key(e):
+    # A single string, so `canonical` cannot reorder src and dst into agreement.
```
ljeub-pometry, line 188: canonical needs to be rethought, it should only be re-ordering neighbours/edges/nodes (it's a bit tricky)

python/tests/test_base_install/test_graphql/parity/test_parity_filters.py
```diff
@@ -0,0 +1,1297 @@
...
+        lambda: (f.Node.name() == "iso") | (f.Edge.property("weight") > 3.0),
+        "a node OR an edge predicate: each branch leaves the other entity "
+        "type unconstrained, so the disjunction admits everything",
+    ),
+    "universal.view_or": (
+        lambda: f.Graph.at(3) | f.Graph.at(5),
+        "a disjunction of two view scopes widens rather than narrows",
+    ),
+    "universal.view_not": (
+        lambda: ~f.Graph.layer("knows"),
```
ljeub-pometry, line 667: we might need to think about what this is supposed to actually mean (it is weird that this seems to be a no-op), should this be equivalent to excluding all events that are in the view from the filtered graph?

python/tests/test_base_install/test_graphql/parity/test_parity_filters.py
```diff
@@ -0,0 +1,1297 @@
...
+    ):
+        probed = canonical(probe(apply(side, build())))
+        _assert_discriminates(
+            side_name, f"{site}.filter({name})", probed, baseline[side_name], name
+        )
+
+
+# --- the `collection[expr]` sugar -------------------------------------------
+#
+# `nodes[expr]` / `edges[expr]` narrow membership at that step only. Remotely
```
ljeub-pometry, line 882: we need to fix the select in graphql to accept everything (this should be straight-forward)

python/tests/test_base_install/test_graphql/parity/test_parity_filters.py
```diff
@@ -0,0 +1,1297 @@
...
+
+def test_comparison_to_a_non_string_type_is_a_python_error(filter_pair):
+    """`node_type() == 5` never becomes an expression, on either side.
+
+    The comparison falls back to Python's default `==`, yielding a plain `bool`
+    rather than a `FilterExpr`, so the *call* is what fails. Worth pinning:
+    the failure has to be a `TypeError` from argument conversion on both sides,
+    not a filter that silently matches everything.
+    """
+    not_an_expr = f.Node.node_type() == 5
```
ljeub-pometry: this should raise an exception and not a return a bool, this is going to become a weird problem in the future otherwise if we ever allow bool throuhg as a filter.

python/tests/test_base_install/test_graphql/parity/test_parity_filters.py
```diff
@@ -0,0 +1,1297 @@
...
+        .src.name,
+    ),
+    (
+        "filter.nested_edges.filter",
+        lambda g: sorted(
+            sorted((e.src.name, e.dst.name) for e in sub)
+            for sub in g.nodes.edges.filter(f.Edge.property("weight") > 2.0)
+        ),
+    ),
+    # `select` is the remote's explicit narrow-here-only form; locally only the
```
ljeub-pometry: this is trivial to fix, make the remote accept the `[]` syntax

python/tests/test_base_install/test_graphql/parity/test_parity_generative.py
```diff
@@ -0,0 +1,498 @@
...
+@_gen_settings(max_examples=150)
+def test_generated_writes_full_state_parity(client, ops):
+    """Any generated write sequence leaves both sides in the same full state.
+
+    Rejections along the way are themselves parity-checked per op (both sides
+    must refuse, same type) and the sequence continues past them, so one
+    conflicting metadata write cannot mask what the rest of the sequence does.
+    """
+    pair, rejected = _fresh_pair(client, ops)
+    event(f"writes: rejected={min(rejected, 3)}")
```
ljeub-pometry: what is this min doing?

python/tests/test_base_install/test_graphql/parity/test_parity_generative.py
```diff
@@ -0,0 +1,498 @@
...
+    )
+    assert_parity(pair, probe_full_state)
+
+
+# --- P2: generated filters on generated graphs ---------------------------------
+
+# A seed prefix guaranteeing every filterable key, layer and node type occurs,
+# so a generated expression usually has something to bite on and the
+# discriminating-outcome rate below stays honest work rather than luck.
+_FILTER_SEED_OPS = [
```
ljeub-pometry: this seems like a bad idea, you probably want to test things against empty/almost empty graphs as that is a common edge case for things to break

python/tests/test_base_install/test_graphql/parity/test_parity_generative.py
```diff
@@ -0,0 +1,498 @@
...
+
+def _classify_filtered(filtered, unfiltered):
+    if not filtered["nodes"] and not filtered["edges"]:
+        return "empty"
+    if filtered == unfiltered:
+        return "universal"
+    return "discriminating"
+
+
+@given(ops=write_ops(max_size=8), expr=filter_exprs())
```
ljeub-pometry: might need more than 8 ops...

python/tests/test_base_install/test_graphql/parity/test_parity_generative.py
```diff
@@ -0,0 +1,498 @@
...
+    # already forces the remote answer to be identical.
+    kind = _classify_filtered(filtered, probe_membership(pair.local))
+    event(f"filter: {kind}")
+    target(1.0 if kind == "discriminating" else 0.0, label="discriminating filters")
+
+
+# --- P3: generated expressions at the collection subscripts ---------------------
+
+
+@given(ops=write_ops(max_size=8), expr=filter_exprs())
```
ljeub-pometry: probably more than 8 writes required

python/tests/test_base_install/test_graphql/parity/test_parity_generative.py
```diff
@@ -0,0 +1,498 @@
...
+@_gen_settings(max_examples=200)
+def test_generated_subscript_parity(client, ops, expr):
+    """``nodes[expr]`` and ``edges[expr]`` agree for generated expressions.
+
+    A pure edge-testing expression says nothing about node membership, so on
+    the node side both backends must *refuse* it — that expectation is
+    asserted outright (not just as optional exception parity) whenever the
+    generated tree's leaves are all edge-kind.
+    """
+    pair, _ = _fresh_pair(client, _FILTER_SEED_OPS + ops)
```
ljeub-pometry: no seeding so we actually cover all possible graphs

python/tests/test_base_install/test_graphql/parity/test_parity_generative.py
```diff
@@ -0,0 +1,498 @@
...
+        collected_edges = remote.edges.collect()
+        assert counter.value == 1, f"edges.collect() fired {counter.value} RPCs"
+        assert len(collected_edges) == local.count_edges() == _BIG_EDGES
+
+        # Content equality with local: the full node-name list is a single
+        # RPC; edges are spot-checked through the collected handles (each
+        # handle read is its own RPC, so the full 5000x2 sweep is out of
+        # budget — membership, counts and node identity carry the claim).
+        counter.reset()
+        remote_names = remote.nodes.name
```
ljeub-pometry, line 505: locally, this would be lazy

ljeub-pometry: not for fixing in this PR but this is where paging etc. needs to be hooked in for the remote apis

python/tests/test_base_install/test_graphql/parity/test_parity_props.py
```diff
@@ -0,0 +1,199 @@
...
+import math
+from decimal import Decimal
+
+import pytest
+from raphtory import Prop
+
+from _parity import assert_parity, graph_pair
+
+# A fixed instant, tz-aware, so datetime parity is not clock-dependent.
+_DT = datetime.datetime(2021, 3, 4, 5, 6, 7, tzinfo=datetime.timezone.utc)
```
ljeub-pometry, line 23: should check naive date time as well

python/tests/test_base_install/test_graphql/parity/test_parity_props.py
```diff
@@ -0,0 +1,199 @@
...
+
+# Insertion order is deliberately NOT alphabetical — see the map-order test.
+_MAP = {"zeta": 1, "alpha": 2, "mid": 3}
+
+_TYPED_PROPS = {
```
ljeub-pometry, line 31: this is missing a bunch of types

python/tests/test_base_install/test_graphql/parity/test_parity_props.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+@pytest.mark.parametrize("key", _TYPED_KEYS)
+def test_node_typed_property_dtype_parity(typed_pair, key):
+    # `repr(PropType)` rather than the object: it renders the full shape
+    # (`U8`, `Decimal(2)`, `List<I64>`, `Map{...}`), so a widened or erased
+    # type shows up in the diff instead of comparing equal by coincidence.
+    # `repr` and not `str`: `str` of a Map dtype iterates a hash map, so its
+    # key order varies between calls *on the same graph* — `repr` sorts.
+    assert_parity(typed_pair, lambda g: repr(g.node("a").properties.get_dtype_of(key)))
```
ljeub-pometry: map properties should be preserving insertion order, also, there is absolutely no reason to expect `repr` to be more reliable than equality comparison so might as well use that (and add some tests to make sure it works as expected)

python/tests/test_base_install/test_graphql/parity/test_parity_props.py
```diff
@@ -0,0 +1,199 @@
...
+def _build_non_finite(g):
+    g.add_node(
+        1,
+        "a",
+        properties={
+            "nan": float("nan"),
+            "pos_inf": float("inf"),
+            "neg_inf": float("-inf"),
+            "nan_f32": Prop.f32(float("nan")),
+            "pos_inf_f32": Prop.f32(float("inf")),
```
ljeub-pometry: "neg_inf_f32" missing

python/tests/test_base_install/test_graphql/parity/test_parity_props.py
```diff
@@ -0,0 +1,199 @@
...
+    comparator's sentinel folding if both sides degraded the same way."""
+    for side in (non_finite_pair.local, non_finite_pair.remote):
+        value = side.node("a").properties.get(key)
+        assert isinstance(value, float) and math.isnan(
+            value
+        ), f"expected NaN for {key!r}, got {value!r}"
+
+
+def test_infinities_are_infinite_on_both_sides(non_finite_pair):
+    for side in (non_finite_pair.local, non_finite_pair.remote):
```
ljeub-pometry: just doing an explict test like this for all the values is enough, no reason for also having a parity check (bonus points for parametrising it over node/edge/graph with properties and metadata)

python/tests/test_base_install/test_graphql/parity/test_parity_props.py
```diff
@@ -0,0 +1,199 @@
...
+        props = side.node("a").properties
+        assert props.get("pos_inf") == math.inf
+        assert props.get("neg_inf") == -math.inf
+
+
+# --- map key order ----------------------------------------------------------
+
+
+def test_map_property_key_order_parity(typed_pair):
```
ljeub-pometry: the order of keys in a map shouldn't really matter?

python/tests/test_base_install/test_graphql/parity/test_parity_rpc_counts.py
```diff
@@ -0,0 +1,767 @@
+"""Transport-contract tests: the RPC counts the remote client documents.
```
ljeub-pometry, line 1: this seems to be strangely obsessive over exact number of ipc calls, shouldn't we instead have a benchmark for the remote apis which verifies the actual performance?

python/tests/test_base_install/test_graphql/parity/test_parity_sorting.py
```diff
@@ -0,0 +1,168 @@
+"""Ordering parity for the remote-only `sorted(...)` surface, incl. nested keys.
```
ljeub-pometry, line 1: We should add sorting to the local apis...

python/tests/test_base_install/test_graphql/parity/test_parity_views.py
```diff
@@ -0,0 +1,387 @@
...
+  edge pairs, degrees, earliest/latest time, layer names, history and edge
+  validity, and
+* the **window bounds** the view installs — ``start`` / ``end`` /
+  ``window_size``, which are observable on every handle type.
+
+Every pair is also guarded against a *vacuous* pass: the op must demonstrably
+change what the handle reports versus the same handle unviewed, on both sides.
+Without that guard a remote that ignored every view op would pass the entire
+matrix.
+"""
```
ljeub-pometry, line 33: in general, the tests for the remote api should be able to assume that the local api is working and not dropping views completely?

python/tests/test_base_install/test_graphql/parity/test_parity_views.py
```diff
@@ -0,0 +1,387 @@
...
+    assert_parity(
+        matrix_pair,
+        lambda g: sorted(
+            (source.name, tuple(sorted(path.degree())))
+            for source, path in g.nodes.neighbours
+        ),
+    )
+
+
+def test_delete_edge_takes_an_event_id_on_both_sides():
```
ljeub-pometry: this test seems to be in the wrong place, should be in writes

python/tests/test_base_install/test_graphql/parity/test_parity_writes_deep.py
```diff
@@ -0,0 +1,1276 @@
...
+        # The auto-assigned `event_id` on that tombstone must match too.
+        assert_parity(pair, lambda g: _stamps(g.edge("x", "y").deletions))
+
+
+def test_edge_delete_without_a_default_layer_is_rejected_on_both_sides():
+    """``Edge.delete()`` with no layer needs a default layer to already exist.
+
+    A real asymmetry inside the write path, identical on both sides:
+    ``graph.delete_edge(t, src, dst)`` *creates* the default layer on demand,
+    while ``edge.delete(t)`` refuses when there isn't one. Pinned because the
```
ljeub-pometry: this seems like a bug, the edge needs to exist but the layer shouldn't have to

python/tests/test_base_install/test_graphql/parity/test_parity_writes_deep.py
```diff
@@ -0,0 +1,1276 @@
...
+    sides — the layer argument reaches the server rather than being dropped."""
+    with graph_pair(_build_layer_seed) as pair:
+        _assert_write_lands(pair, _layer_scoped_node_update, _node_update_layer_facts)
+
+
+@pytest.mark.parametrize("layer", WRITTEN_LAYERS)
+def test_layer_view_after_layered_writes_parity(layer):
+    """Reads under ``layer(name)`` agree, for every layer that was written."""
+    with graph_pair(_build_layered) as pair:
+        assert_parity(pair, _per_layer_probe(layer))
```
ljeub-pometry, line 447: I think it would make more sense for all the write tests to always probe across all layers (i.e., run _per_layer_ probe for every layer) and the full graph to make sure updates didn't go to a weird place (this is what the graph equality assertions to in rust)

raphtory-graphql/src/client/remote/remote_graph.rs
```diff
@@ -840,12 +840,13 @@ impl RemoteGraph {
         src: G,
         dst: G,
         layer: Option<String>,
+        event_id: Option<usize>,
```
ljeub-pometry: this is rust api, should use the into_input_time mechanism used by normal delete_edge...

ljeub-pometry: Note that this is handled badly by all the addition functions in here

raphtory-graphql/src/client/remote/remote_node.rs
```diff
@@ -504,13 +504,15 @@ impl RemoteNode {
         &self,
         t: T,
         properties: Option<HashMap<String, Prop>>,
+        layer: Option<String>,
         event_id: Option<usize>,
```
ljeub-pometry: rust api should use the `TryIntoInputTime` mechanism for handling this

raphtory-graphql/src/client/graphql_transport.rs
```diff
@@ -1191,17 +1184,17 @@ fn render_read_into(
             let _ = write!(out, " {{ filter(expr: {})", vars.add_filter(filter)?);
         }
         ReadExpr::SelectNodes { input, filter } => {
-            // Server field `select(expr: NodeFilter!)`: narrows the current
+            // Server field `select(expr: GqlFilter!)`: narrows the current
             // collection's membership only; downstream traversals see the
             // unfiltered graph.
             render_read_into(input, vars, out)?;
-            let _ = write!(out, " {{ select(expr: {})", vars.add_node_filter(filter)?);
+            let _ = write!(out, " {{ select(expr: {})", vars.add_filter(filter)?);
```
ljeub-pometry: silently ignoring errors here is weird, shouldn't we fail loudly if we can't parse the filter properly?

ljeub-pometry (changes requested):
- A lot of the comments should be turned into issues for future fixes to the remote apis
- Fix the trivial api discrepancies
- Remove the seeding from the generated graphs, testing against empty examples is important and should also help with narrowing down edge cases on failure
- The whole `FitlerTree`/`CompsiteFilter` apis are a mess but we probably need the expressions rework to fix that

python/tests/test_base_install/test_graphql/parity/test_parity_filters.py
shivamka1: Checked this against local rather than changing it — the server already takes the unified `GqlFilter` on both `Nodes.select` and `Edges.select`, and remote now accepts exactly what local accepts:

| expression on `nodes[...]` | local | remote |
|---|---|---|
| node predicate | `['a']` | `['a']` |
| mixed `and` (node & edge) | `Node filter expected` | `Node filter expected` |
| mixed `or` | `Node filter expected` | `Node filter expected` |
| `not` edge | `Node filter expected` | `Node filter expected` |
| edge predicate | `Node filter expected` | `Node filter expected` |

Same exception type, message and timing (at subscript, not first read). `edges.select(node_expr)` is accepted on both, matching local too. The remaining client-side gate only mirrors local's own restriction — widening it further would make remote accept what local rejects. If the intent is that `nodes[edge_expr]` should work at all, that's a change to local semantics first; happy to open an issue for it.

python/tests/test_base_install/test_graphql/parity/test_parity_filters.py
shivamka1: `[]` is already supported remotely — `__getitem__` exists on `RemoteNodes`, `RemoteEdges` and `RemoteNestedEdges`, each delegating to `select`, and it's exercised over the wire by the `GETITEM_SITES` matrix (every expression × nodes/edges × both backends).

The ledger text was misleading: it was describing `select()` having no *local* twin, not remote lacking `[]`. Removed those two entries so the subscript form is just an ordinary parity case — 2 fewer xfails, no coverage lost.

raphtory-graphql/src/client/graphql_transport.rs
shivamka1: Filter parse errors were already loud — `vars.add_filter(filter)?` propagates; the `let _ = ` was only discarding the `write!` result, which is infallible for a `String`. But it read as error-swallowing in 35 places, so rather than argue the point I've removed the appearance: added `From<fmt::Error> for ClientError` and every site is now `write!(...)?`.

raphtory-graphql/src/client/remote/remote_graph.rs
shivamka1: Switched the client write methods to `TryIntoInputTime` — same trait the local mutation API uses (`deletion_ops.rs`/`addition_ops.rs`). The separate `event_id` parameter is gone from the Rust API; `.t()` was discarding whatever the time argument carried and then asking the caller for it again.

Covers `add_node`, `create_node`, `add_edge`, `delete_edge`, `add_properties`, `Edge::add_updates`, `Edge::delete`, `Node::add_updates`.

Side effect worth noting: an unparseable time now fails client-side with a proper message rather than going to the server as a nonsense timestamp (needed `From<ParseTimeError> for ClientError`). The Python surface is unchanged — `event_id=` stays a kwarg there, since a bare timestamp can't distinguish "unset" from "0", so the binding assembles the `InputTime` and hands the client one value.

python/tests/test_base_install/test_graphql/parity/test_parity_generative.py
shivamka1: Seed prefix removed — every filter example now runs against a purely generated graph, so graphs *lacking* those keys/layers/types are reachable for the first time. `target()` still steers generation toward expressions that match part of the graph, which was the principled version of what the prefix was doing by force.

Measured trade-off (filter property, outcome shares):

| | seeded | unseeded |
|---|---|---|
| matched some (the case that proves anything) | 41.8% | 32.9% |
| matched none | 6.6% | 16.5% |
| matched all | 19.4% | 13.7% |

The rate drops, but each example got ~2.5x cheaper without the 6 replayed writes, so I raised the budgets (300→600, 200→400) — landing at roughly the original *absolute* coverage over a wider input space, with the whole file going 242s → 188s.

Also relabelled the outcomes from `discriminating`/`universal`/`empty` to `matched_some`/`matched_all`/`matched_none` so the statistics say what the filter actually did.

One thing this surfaced: **~37% of generated expressions are refused by both sides**, because they name a property the generated graph doesn't have. That's the generator spending most of its budget on inputs that can't test filtering, and it's the biggest remaining lever on example quality. Looking at that next.

python/tests/test_base_install/test_graphql/parity/test_parity_writes_deep.py
shivamka1: Done — `_containment_facts` (every layer's nodes/edges/count plus the graph totals, name-based so nothing can be masked) now runs after **every** write:

- `_assert_write_lands`: containment compared local-vs-remote, so a write that lands where asked but *also* leaks into `_default` or a sibling layer fails — the targeted probe alone couldn't see that.
- `_assert_write_rejected`: stronger — containment is snapshotted before and asserted **unchanged per side**, not merely equal across sides. A rejected write has no oracle problem (nothing may change anywhere), so even debris conjured identically on both backends fails.

One honest caveat: for *landed* writes this is parity, so a wrong-but-symmetric behaviour still passes — pinning absolute layer expectations there ties into the unlayered-delete question on the other thread.

python/tests/test_base_install/test_graphql/parity/_strategies.py
shivamka1: The value layer is now schema-first, modelled on the rust generator you pointed at (`db4-storage/src/pages/test_utils/props.rs`): each example draws a schema (property name → type), writes conform to it, and filter expressions draw their keys *from* it with same-typed comparison values.

- **Types**: 6 flat → 13 leaves (`i64 i32 u8 u16 u32 u64 f64 f32 bool str dtime ndtime decimal`) plus `List<T>`/`Map<T>`; an instrumented run showed zero type-related rejections, so the whole widened surface really executes.
- **Registry**: one `KIND_OPERATORS` table maps type-kind → operator → arity; `_apply_comparison` is now just `getattr(target, op)(*args)` with dunder names and an empty tuple for `is_some`/`is_none` — generation and application can't drift.
- **Correlation, not seeding**: no fixed content — a graph may contain any subset of its schema including none — but expressions no longer name properties that couldn't exist. Rejected outcomes dropped 43% → ~32%, and what remains is deliberate: ~10% missing-key leaves kept on purpose, layer-mismatch view atoms, and filters naming schema keys the random writes didn't write (which exercises absent-property behaviour as exception parity).

python/tests/test_base_install/test_graphql/parity/test_parity_views.py
shivamka1: The guard runs on the local side deliberately, and it isn't re-testing the local API — it's testing whether *this fixture* still has bite.

Parity compares two answers, so an op that happens to be a no-op on this particular graph makes both handles report the unviewed graph — they agree, the assertion passes, and nothing was proved. A remote that dropped every view op would sail through the same matrix. The local half of the guard is what notices that.

And when it fires, the cause is almost always the case rather than the engine: a window wide enough to cover every event, a layer every edge belongs to, an `at` on a timestamp where nothing happens. Those decay silently as the fixture drifts, so without the guard the suite gradually stops testing anything while staying green.

I've added that rationale to the module docstring (and to the matching guard in `test_parity_filters.py`) in a5dd7857a, so it doesn't read as redundant local coverage.

python/tests/test_base_install/test_graphql/parity/test_parity_rpc_counts.py
shivamka1: `graphql-bench` benchmarks the server via k6, so it never touches the Python client — it can't replace the RPC counts, which are about the client's round-trip contract, not speed. A client that fired three round trips per terminal would still answer correctly and still benchmark fine.

If you want client performance tracked too, that's a separate Python benchmark feeding the same `output.json` the k6 job already publishes. Happy to do it as a follow-up.

python/tests/test_base_install/test_graphql/parity/test_parity_filters.py
shivamka1: Agreed, it's under-defined. Right now negating a view scope is a no-op and the test just pins that, so if we change it later it's a deliberate change rather than a silent one.

Your suggestion (excluding the events that are in the view) feels like the right reading to me, but it's a semantics decision rather than something a test PR should settle. Want me to open an issue for it?

python/tests/test_base_install/test_graphql/parity/test_parity_generative.py
shivamka1: Yes, remote `collect()` is eager where local iteration is lazy. Server side paging already exists (`nodes.page(limit, offset)`), it's just not wired into the client collections yet, so this is the right seam for it.

Leaving it for the follow-up as you suggested.

python/tests/test_base_install/test_graphql/parity/test_parity_writes_deep.py
shivamka1: Fixed in b631ae24f, going with your reading: `edge.delete(t)` now creates the default layer when there isn't one, so it's equivalent to `graph.delete_edge(t, src, dst)`.

The cause was in `EdgeView::resolve_layer`. When you name a layer it honours the `create` flag, but the no-layer branch only ever looked `_default` up and errored if it was missing, so the same intent spelled two ways behaved differently. The no-layer branch now creates it too when the caller is writing.

Verified the two spellings leave identical state (same layers, same tombstone). The parity test that pinned the old rejection is inverted to assert the new behaviour.

Still odd, and not covered by this fix: the edge ends up listed in `layer_names` for a layer it was never added to, and `is_deleted()` reports true while it stays valid in `l1` and `l2`. Happy to raise that separately if you think it's worth changing.

python/python/raphtory/graphql/__init__.pyi
```diff
@@ -658,23 +847,82 @@ class RemoteGraph(object):
...
-        """View end bound. `None` for an unbounded view. Property — fires one RPC."""
+    def end(self) -> OptionalEventTime:
+        """
+        View end bound. `None` for an unbounded view. Property — fires one RPC.
+
+        Returns:
+            OptionalEventTime: the view end bound, or empty if unbounded.
+        """
+
+    def event_graph(self) -> RemoteGraph:
```
ljeub-pometry, line 863: we probably should have the same base graph versus view distinction on the remote side as well such that this is easier to document properly (make an issue for the future, not that critical)

python/tests/test_base_install/test_graphql/parity/_parity.py
```diff
@@ -44,69 +56,109 @@ def graph_pair(build, graph_type="EVENT"):
...
+    which kind to create and the local side is instantiated from the matching
+    class, so the pair always compares like with like.
     """
-    local = Graph()
+    try:
+        local_cls = _LOCAL_CLASS[graph_type]
+    except KeyError:
+        raise ValueError(
+            f"unknown graph_type {graph_type!r}, expected one of {list(GRAPH_TYPES)}"
+        ) from None
```
ljeub-pometry: this from isn't needed?

python/tests/test_base_install/test_graphql/parity/_parity.py
```diff
@@ -44,69 +56,109 @@ def graph_pair(build, graph_type="EVENT"):
...
-        return ("node", name)
+    identity = _identity(value)
+    if identity is not None:
+        return identity
+
+    if isinstance(value, dict):
+        # `dict.__eq__` already ignores key order, so keys are left alone; a
+        # test that cares about map *order* asserts it explicitly.
+        return {k: canonical(v) for k, v in value.items()}
+    if isinstance(value, (list, tuple)):
```
ljeub-pometry: this is the same as the `iter` case below, no need for an extra check

python/tests/test_base_install/test_graphql/parity/_parity.py
```diff
@@ -44,69 +56,109 @@ def graph_pair(build, graph_type="EVENT"):
...
+       reordering anything, so a real difference still fails.
+
+    Nothing else is touched. Float precision, datetime timezone, ``NaN``,
+    map key order and ``EventTime``-vs-``int`` all compare as they come: if a
+    value does not survive the round-trip exactly, that is a product bug for
+    ``KNOWN_GAPS`` and an issue, not something to paper over here.
     """
-    # bool is an int subclass — handle before int.
-    if value is None or isinstance(value, bool):
+    if isinstance(value, (str, bytes)) or value is None or isinstance(value, bool):
```
ljeub-pometry: only the str/bytes check here is necessary as they would otherwise be treated as iterables, everything else should already fall through

python/tests/test_base_install/test_graphql/parity/_parity.py
```diff
@@ -44,69 +56,109 @@ def graph_pair(build, graph_type="EVENT"):
...
         return value
-    if isinstance(value, float):
-        return "nan" if math.isnan(value) else round(value, _FLOAT_PLACES)
-    if isinstance(value, (int, str)):
+    if isinstance(value, (int, float)):
```
ljeub-pometry: not needed

python/tests/test_base_install/test_graphql/parity/_parity.py
```diff
@@ -44,69 +56,109 @@ def graph_pair(build, graph_type="EVENT"):
...
+
+    if isinstance(value, dict):
+        # `dict.__eq__` already ignores key order, so keys are left alone; a
+        # test that cares about map *order* asserts it explicitly.
+        return {k: canonical(v) for k, v in value.items()}
+    if isinstance(value, (list, tuple)):
+        return [canonical(v) for v in value]
 
-    # Anything iterable (collections, histories) → order-insensitive list.
     try:
```
ljeub-pometry: would be much more readable with `isinstance(Iterable) and not isinstance(value, (str, bytes))` check here and a final `return value` fall through at the end

python/tests/test_base_install/test_graphql/parity/_parity.py
ljeub-pometry: actually, get rid of this completely and add the str check later where the iterable case is handled

python/tests/test_base_install/test_graphql/parity/_strategies.py
```diff
@@ -1,58 +1,209 @@
...
 
 from hypothesis import strategies as st
 
+from raphtory import Prop
 from raphtory import filter as f
 
-# --- pools --------------------------------------------------------------------
+# --- name pools -----------------------------------------------------------------
 
 NODE_NAMES = ["n0", "n1", "n2", "n3", "n4", "n5"]
```
ljeub-pometry, line 52: we probably should cover integer node ids as well, things might go wrong there

ljeub-pometry: probably easiest to cover this as part of the schema generation

python/tests/test_base_install/test_graphql/parity/_strategies.py
```diff
@@ -1,58 +1,209 @@
...
+    "i32": ([-70000, 0, 3], Prop.i32, "orderable"),
+    "u8": ([0, 7, 255], Prop.u8, "orderable"),
+    "u16": ([0, 300, 65535], Prop.u16, "orderable"),
+    "u32": ([0, 70000], Prop.u32, "orderable"),
+    "u64": ([0, 2**40], Prop.u64, "orderable"),
+    "f64": ([-3.25, 0.5, 1.5, 2.5], None, "orderable"),
+    "f32": ([-3.25, 0.5, 1.5], Prop.f32, "orderable"),
+    "bool": ([True, False], None, "flag"),
+    "str": (["red", "green", "blue", "redish"], None, "text"),
+    "dtime": ([_DT_A, _DT_B], Prop.aware_datetime, "orderable"),
```
ljeub-pometry: `"dtime"` and `"ndtime"` should roundtrip without explicit conversion?

python/tests/test_base_install/test_graphql/parity/_strategies.py
```diff
@@ -1,58 +1,209 @@
...
+    "ndtime": ([_NDT_A, _NDT_B], Prop.naive_datetime, "orderable"),
+    "decimal": ([Decimal("3.14"), Decimal("-0.50")], Prop.decimal, "orderable"),
```
ljeub-pometry: `"decimal"` shouldn't need a converter

python/tests/test_base_install/test_graphql/parity/test_parity_comparator.py
```diff
@@ -0,0 +1,178 @@
...
+    assert not _agree(math.nan, math.nan)
+
+
+def test_datetime_timezone_is_not_normalized():
+    """A naive and an aware datetime for the same instant are not the same
+    answer, so the comparator must not reduce both to an epoch float."""
+    naive = datetime.datetime(2021, 3, 4, 5, 6, 7)
+    aware = datetime.datetime(2021, 3, 4, 5, 6, 7, tzinfo=datetime.timezone.utc)
+    assert not _agree(naive, aware)
+    assert _agree(aware, aware)
```
ljeub-pometry, line 124: missing `_agree(naive, naive)`

python/tests/test_base_install/test_graphql/parity/test_parity_comparator.py
```diff
@@ -0,0 +1,178 @@
...
+    assert not _agree({"a": 1, "b": 2}, {"a": 1, "b": 3})
+    assert not _agree({"a": 1}, {"a": 1, "b": 2})
+
+
+def test_scalar_types_are_not_coerced():
+    """No cross-type leniency: a stringified number is not the number."""
+    assert not _agree(1, "1")
+    assert not _agree(1.0, "1.0")
+    assert not _agree(None, 0)
+    assert not _agree(True, 1.5)
```
ljeub-pometry, line 172: interesting edge case: `_agree(True, 1)` which probably passes...
Likely not worth fixing that as it is tricky

python/tests/test_base_install/test_graphql/parity/test_parity_comparator.py
```diff
@@ -0,0 +1,178 @@
...
+
+# --- the assertion wrapper itself -------------------------------------------
+
+
+def test_assert_parity_fails_on_a_real_difference():
+    with pytest.raises(AssertionError, match="value parity mismatch"):
+        assert_parity(_pair(1, 2), lambda g: g)
+
+
+def test_assert_parity_requires_both_sides_to_raise_the_same_type():
```
ljeub-pometry, line 189: what happens if remote and local both raise but with a different exception type?

ljeub-pometry: keep this test but add another test where the two sides raise different exceptions

python/tests/test_base_install/test_graphql/parity/test_parity_filters.py
ljeub-pometry: open the issue so we don't forget about it, agreed that it isn't anything that needs to change here.

python/tests/test_base_install/test_graphql/parity/test_parity_filters.py
```diff
@@ -1090,17 +1093,26 @@ def test_edge_expr_in_a_node_subscript_is_refused_the_same_way(
 def test_comparison_to_a_non_string_type_is_a_python_error(filter_pair):
```
ljeub-pometry: this test doesn't belong here, it should go with the other tests for the filter builders

python/tests/test_base_install/test_graphql/parity/test_parity_generative.py
```diff
@@ -257,86 +257,100 @@ def _parity_with_outcome(pair, fn):
...
+    ops = case[2]
+    pair, rejected = _fresh_pair(client, case)
+    # These are Hypothesis `event` labels — they only shape the coverage
+    # summary, never the assertions. Both are bucketed rather than exact so the
+    # summary stays readable: an unbounded `rejected` count would print one
+    # line per distinct number, so 3+ collapses into a single "3" bucket.
+    event(f"writes: rejected={min(rejected, 3)}{'+' if rejected > 3 else ''}")
     event(
-        f"writes: ops={'few' if len(ops) <= 4 else 'several' if len(ops) <= 8 else 'many'}"
+        f"writes: ops={'none' if not ops else 'few' if len(ops) <= 4 else 'several' if len(ops) <= 8 else 'many'}"
```
ljeub-pometry: explicit category labels? `"0"`, `"1-4"`, `"5-8"`, `">8"` so one doesn't have to look at the code to figure out what e.g. `"few"` means?

python/tests/test_base_install/test_graphql/parity/test_parity_props.py
```diff
@@ -146,54 +166,94 @@ def non_finite_pair():
...
-    assert_parity(non_finite_pair, lambda g: g.node("a").properties.get(key))
+# Every place a property can live: (carrier, container) -> reader.
+_CARRIERS = {
+    "node.properties": lambda g: g.node("a").properties,
+    "node.metadata": lambda g: g.node("a").metadata,
+    "edge.properties": lambda g: g.edge("a", "b").properties,
+    "edge.metadata": lambda g: g.edge("a", "b").metadata,
+    "graph.properties": lambda g: g.properties,
+    "graph.metadata": lambda g: g.metadata,
+}
```
ljeub-pometry, line 183: should check the collections as well (`nodes`, `edges`, `neighbours`, etc.)

python/tests/test_base_install/test_graphql/parity/test_parity_types.py
```diff
@@ -0,0 +1,142 @@
+"""Return *types* must match, not just return values.
```
ljeub-pometry: this should be added to the parity assertions

python/tests/test_base_install/test_graphql/parity/test_parity_types.py
```diff
@@ -0,0 +1,142 @@
...
+@pytest.fixture(scope="module")
+def pair():
+    with graph_pair(_build) as p:
+        yield p
+
+
+# Accessors that yield a single optional time. Locally every one of these is an
+# `OptionalEventTime` — one static class whether or not a time exists — so the
+# remote must not substitute `EventTime`/`None`.
+SCALAR_TIME_ACCESSORS = {
```
ljeub-pometry: We should probably make these return OptionalEventTime locally, make an issue

raphtory-graphql/src/client/graphql_transport.rs
```diff
@@ -4609,13 +4839,13 @@ mod tests {
 
         for (name, score) in [("a", 10i64), ("b", 20), ("c", 30)] {
             let props: Map<String, Prop> = [("score".to_string(), Prop::I64(score))].into();
-            rg.add_node(1i64, name, Some(props), None, None, None)
+            rg.add_node(1i64, name, Some(props), None, None)
```
ljeub-pometry: any reason the property input is handled differently here compared to the local rust apis?

raphtory-graphql/src/client/op.rs
```diff
@@ -49,7 +49,15 @@ pub enum Op {
...
     /// Start of every read tree — names the graph.
-    Root { path: String },
+    Root {
+        path: String,
+        /// Optional graph-semantics override, rendered as the server's
+        /// `graphType:` argument (`EVENT` / `PERSISTENT`). `None` keeps the
+        /// stored graph's native flavour. Set by `RemoteGraph::event_graph` /
+        /// `persistent_graph` — the remote form of the local zero-copy
+        /// flavour conversions, applied at load time by the server.
+        graph_type: Option<String>,
```
ljeub-pometry: this should be an enum, not a string

ljeub-pometry (approved):
Some minor things to clean up but looks ok otherwise

python/tests/test_base_install/test_graphql/parity/_parity.py
shivamka1: Done. The scalar block at the top is gone, the str/bytes check now lives on the iterable branch, and everything else falls through to a final `return value`.

python/tests/test_base_install/test_graphql/parity/_parity.py
shivamka1: Removed along with the rest of the scalar block.

python/tests/test_base_install/test_graphql/parity/_parity.py
shivamka1: Removed, lists and tuples now go through the same iterable branch.

python/tests/test_base_install/test_graphql/parity/_parity.py
shivamka1: Done. The try/iter probe is now `isinstance(value, Iterable) and not isinstance(value, (str, bytes))` with `return value` as the final fallthrough.

python/tests/test_base_install/test_graphql/parity/_parity.py
shivamka1: Removed. It was only suppressing the chained KeyError in the traceback, which is harmless noise for a test helper.

python/tests/test_base_install/test_graphql/parity/test_parity_comparator.py
shivamka1: Added.

python/tests/test_base_install/test_graphql/parity/test_parity_comparator.py
shivamka1: Added `test_assert_parity_rejects_two_different_exception_types`: ValueError on one side and TypeError on the other now has a test pinning that the parity check fails.

python/tests/test_base_install/test_graphql/parity/test_parity_comparator.py
shivamka1: Confirmed, it passes: `True == 1` in Python so the comparator cannot see the difference by value. Agreed it is not worth special-casing bool here. Your other comment about folding type checks into the parity assertions would catch this for free though, since bool vs int is a class mismatch, so it should fall out of that change.

python/tests/test_base_install/test_graphql/parity/test_parity_generative.py
shivamka1: Done, the buckets are now `0`, `1-4`, `5-8`, `>8`.

python/tests/test_base_install/test_graphql/parity/test_parity_filters.py
shivamka1: Moved to `test_filters/test_node_filter.py` next to the other node_type tests. I dropped the trailing parity assertion when moving it since `nodes[expr]` over the wire is already covered by the subscript matrix in this file.

python/python/raphtory/graphql/__init__.pyi
shivamka1: Filed as https://github.com/Pometry/Raphtory/issues/2716.

python/tests/test_base_install/test_graphql/parity/test_parity_sorting.py
shivamka1: Filed as https://github.com/Pometry/Raphtory/issues/2717.

python/tests/test_base_install/test_graphql/parity/test_parity_filters.py
shivamka1: Filed as https://github.com/Pometry/Raphtory/issues/2718.

python/tests/test_base_install/test_graphql/parity/test_parity_writes_deep.py
shivamka1: Filed as https://github.com/Pometry/Raphtory/issues/2719.

python/tests/test_base_install/test_graphql/parity/_strategies.py
shivamka1: Checked what `canonical` was actually absorbing, and the answer for entity collections was nothing, so it is gone rather than ledgered.

It sorted nodes/edges/neighbours on the grounds that iteration order is unspecified. Measured: local and remote agree on the order of nodes, edges, neighbours, windowed and subgraph collections, and both sides are insertion-ordered rather than coincidentally name-sorted. With the sort removed the whole suite passes, generative properties included. So it is removed and there is a comparator case pinning that a reordered collection now fails.

That leaves two things it still does, and I would argue neither is a divergence to record:

- entity handles reduce to `("node", name)` / `("edge", src, dst)`, because a `Node` and a `RemoteNode` point at different graphs and could never compare equal by `==`
- containers like `History` vs `RemoteHistory` are listed, because they are distinct classes; contents and order both survive to the assertion

Everything else compares as it comes: float precision, NaN, datetime timezone, map ordering and now collection order. The negative cases in `test_parity_comparator.py` pin each of those.

PR #2709, opened 2026-08-13
Optimise filtering for select on nodes and edges

ljeub-pometry:

- Adds `domain` implementations for all the node ops
- Removes the default implementation for `domain` so we are less likely to miss obvious optimisations in the future
- Reworks the internal apis for path and edges select such that layers/subgraphs can be pushed down instead of scanning over all edges to apply the filter
- Fixed `EdgeFilter` and `ExplodedEdgeFilter` such that `EdgeFilter.window(...)`, `EdgeFilter.layer(...)`, etc. no longer apply the window/layer to the graph itself.
- As part of this, `EdgeFilter.window(...).is_self_loop()` no longer filters out self loops that are not included in the window (window/layer has no effect on is_self_loop).

performance and correctness

no, this should be a performance-only change

- the tests
- new test for select with layers

- we need a lot more tests for combining filters


raphtory/src/db/api/view/internal/time_semantics/persistent_semantics.rs
```diff
@@ -1075,21 +1117,31 @@ impl EdgeTimeSemanticsOps for PersistentSemantics {
         edge_alive_at_end(e, r.end, view)
     }
 
+    /// # Semantics
+    ///
+    /// - An edge is deleted if it is deleted in all layers
+    /// - An edge that has no visible deletion event is not deleted (if it has no visible event at
```
miratepuffin, line 1105: It might be worth flagging this to andrea to add in the Persistent semantics docs page

raphtory/src/db/graph/path.rs
```diff
@@ -118,16 +130,21 @@ impl<'graph, G: GraphViewOps<'graph>> PathFromGraph<'graph, G> {
...
-                    let node_type_id = base_graph.node_type_id(*v);
-                    node_types_filter[node_type_id]
-                })
-                .into_dyn_boxed()
-        })
+        PathFromGraph::new_filtered(
+            self.base_graph.clone(),
+            self.select.clone(),
+            self.nodes.clone(),
+            Arc::new(move |graph, vid| {
```
miratepuffin, line 137: Why do you need an Arc in here?

miratepuffin (approved):

raphtory/src/db/graph/path.rs
ljeub-pometry: The Arc is there to make sure everything is clone, it was always there just constructed in a different place.

raphtory/src/db/api/view/internal/time_semantics/persistent_semantics.rs
```diff
@@ -65,24 +65,6 @@ fn last_before<
...
 
-fn is_deleted_before<
-    'a,
-    G: GraphViewOps<'a>,
-    TSA: TimeIndexOps<'a, IndexType = EventTime, RangeType = TSA>,
-    TSD: TimeIndexOps<'a, IndexType = EventTime, RangeType = TSD>,
->(
-    additions: FilteredEdgeTimeIndex<'a, G, TSA>,
-    deletions: FilteredEdgeTimeIndex<'a, G, TSD>,
-    t: EventTime,
```
miratepuffin, line 76: These semantics are still the case yeah?

miratepuffin (approved):

PR #2733, opened 2026-08-24
Check for existing node types on disk when setting node type

fabubaker:
Check disk segments for an existing node type when setting a new type.

Node types, which are immutable, can be changed after a flush since the check will only read the empty in-memory segment.

No.

There is a test in https://github.com/Pometry/pometry-storage/pull/359 that adds a node type, flushes and tries to add a new type.

No.


miratepuffin (approved):
Looks good to me, but probably wants a once over from someone who works on the storage

ljeub-pometry (approved):
Took me a while to spot the actual fix with all the renames (new names are better). Looks good to me.

PR #2742, opened 2026-08-26
Check the sort keys are shared by using them, not by naming them twice

shivamka1:

`test_sort_key_types_are_shared` asserted `raphtory.NodeSortBy is raphtory.graphql.NodeSortBy`. That stopped holding when #2739 removed the duplicate export, so `db_v4` is currently red:

```
>       assert raphtory.NodeSortBy is gql.NodeSortBy
E       AttributeError: module 'graphql' has no attribute 'NodeSortBy'
```

The test now checks what that name was standing in for: the classes live in `raphtory` alone, are not duplicated under `graphql` (which is what made each stub declaration a redefinition of the other), and the same key object sorts a local and a remote collection identically.

`db_v4` fails this test today, so every PR branching from it inherits the failure. It is currently red on Pometry/pometry-storage#352, which changed nothing in this area.

No, test only.

## How was this patch tested?

The whole parity suite rather than the sorting subset — 1670 passed, 9 xfailed. Running only the sorting tests is what let the original break through, since the assertion lived in `test_parity_entities.py`.

### DIFF
--- a/python/tests/test_base_install/test_graphql/parity/test_parity_entities.py
+++ b/python/tests/test_base_install/test_graphql/parity/test_parity_entities.py
@@ -177,11 +177,22 @@ def test_edges_sorted_parity():
 
 
 def test_sort_key_types_are_shared():
-    """One set of sort-key classes serves both sides: `raphtory.NodeSortBy`
-    and `raphtory.graphql.NodeSortBy` are the same class."""
+    """One set of sort-key classes serves both sides.
+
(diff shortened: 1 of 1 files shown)


ljeub-pometry:
This test was completely useless, there are plenty of tests for the actual sorting behaviour. It is already deleted in #2705.

shivamka1:
Folded into #2736 — same repo, same target, and #2736 has to land for the RBAC fix anyway, so this is one fewer PR to sequence.

PR #2743, opened 2026-08-26
Add node type indexing

fabubaker:


db4-storage/src/api/node_type_index.rs
```diff
@@ -0,0 +1,35 @@
...
+    fn new(path: Option<&Path>, ext: Self::Extension) -> Self;
+
+    fn load(path: impl AsRef<Path>, ext: Self::Extension) -> Result<Self, StorageError>;
+
+    fn head(&self) -> RwLockReadGuard<'_, MemNodeTypeIndex>;
+
+    fn head_mut(&self) -> RwLockWriteGuard<'_, MemNodeTypeIndex>;
+
+    /// Returns the sorted, unique `VID`s of nodes whose type is in `type_ids`.
+    fn nodes_of_type(&self, type_ids: &[usize]) -> Vec<VID>;
```
ljeub-pometry: Is it possible to avoid this allocation?

db4-storage/src/pages/locked/nodes.rs
```diff
@@ -17,7 +17,7 @@ use rayon::prelude::*;
 use std::ops::DerefMut;
 
 #[derive(Debug)]
-pub struct LockedNodePage<'a, NS> {
+pub struct LockedNodePage<'a, NS: NodeSegmentOps> {
```
ljeub-pometry: Don't add bounds to the struct itself unless they are necessary, they just become a nuisance that you have to restate everywhere even when you are not using them

db4-storage/src/pages/node_page/writer.rs
```diff
@@ -201,86 +202,84 @@ impl<'a, MP: DerefMut<Target = MemNodeSegment> + 'a, NS: NodeSegmentOps> NodeWri
...
         let props = [(NODE_ID_IDX, gid)];
-        self.update_c_props(pos, layer_id, props);
+        self.update_c_props(pos, STATIC_GRAPH_LAYER_ID, props);
     }
 
-    pub fn store_node_type(&mut self, pos: LocalPOS, layer_id: LayerId, node_type: usize) {
+    pub fn store_node_type(&mut self, pos: LocalPOS, node_type: usize) {
         let props = [(NODE_TYPE_IDX, Prop::U64(node_type as u64))];
-        self.update_c_props(pos, layer_id, props);
+        self.update_c_props(pos, STATIC_GRAPH_LAYER_ID, props);
```
ljeub-pometry, line 259: this sets the type even if it is the default, unlike the other version above when `gid=None`

db4-storage/src/segments/node/segment.rs
```diff
@@ -668,11 +662,11 @@ mod test {
             Some(path.path().to_path_buf()),
             ext.clone(),
         );
-        let stats = GraphStats::default();
 
+        let stats = GraphStats::default();
```
ljeub-pometry, line 675: not used?

db4-storage/src/segments/node_type_index/index.rs
```diff
@@ -0,0 +1,211 @@
...
+        let sets: Vec<_> = type_ids
+            .iter()
+            .filter_map(|type_id| self.map.get(type_id))
+            .collect();
+
+        // No need to dedup after kmerge since a node can only have one type.
+        sets.iter()
+            .map(|set| set.iter().copied())
+            .kmerge()
+            .collect()
```
ljeub-pometry, line 96: We can avoid this collect, also we need a single type id version, there is quite a bit of overhead to forcing a kmerge() when it isn't needed!

db4-storage/src/segments/node_type_index/index.rs
```diff
@@ -0,0 +1,211 @@
...
+        type_ids.sort_unstable();
+
+        type_ids
+            .into_iter()
+            .filter_map(|type_id| {
+                self.map
+                    .get(&type_id)
+                    .map(|set| (type_id, set.iter().copied().collect()))
+            })
+            .collect()
```
ljeub-pometry: This is quite a bit of work, why do we need this?

db4-storage/src/segments/node_type_index/index.rs
```diff
@@ -0,0 +1,211 @@
...
+use raphtory_core::entities::VID;
+use std::{
+    collections::BTreeSet,
+    sync::atomic::{AtomicUsize, Ordering},
+};
+
+/// In-memory index that maps a node type id to `VID`s.
+#[derive(Debug)]
+pub struct MemNodeTypeIndex {
+    map: FxDashMap<usize, BTreeSet<VID>>,
```
ljeub-pometry, line 13: how does this compare to a `RwLock<Vec<RwLock<BTreeSet<VID>>>`? The assumption being that it is very rare for a new node type to appear?

raphtory-graphql/src/model/schema/node_schema.rs
```diff
@@ -86,17 +87,21 @@ impl NodeSchema {
...
-                    let mut node_types_filter =
-                        vec![false; self.graph.node_meta().node_type_meta().num_all_fields()];
+                    let num_type_ids = self.graph.node_meta().node_type_meta().num_all_fields();
+                    let mut node_types_filter = vec![false; num_type_ids];
                     node_types_filter[self.type_id] = true;
-                    let filter = TypeId.mask(node_types_filter.into());
+
+                    let filter =
+                        NodeTypeFilterOp::from_mask(node_types_filter.into(), self.graph.clone());
+
```
ljeub-pometry, line 96: No need to make a mask here since this actually prefers type ids now?

raphtory-graphql/src/model/schema/node_schema.rs
```diff
@@ -136,17 +142,21 @@ impl NodeSchema {
...
                 .filter_map(|(key, dtype)| {
-                    let mut node_types_filter =
-                        vec![false; self.graph.node_meta().node_type_meta().num_all_fields()];
+                    let num_type_ids = self.graph.node_meta().node_type_meta().num_all_fields();
+                    let mut node_types_filter = vec![false; num_type_ids];
                     node_types_filter[self.type_id] = true;
-                    let filter = TypeId.mask(node_types_filter.into());
+
+                    let filter =
+                        NodeTypeFilterOp::from_mask(node_types_filter.into(), self.graph.clone());
```
ljeub-pometry: no need for mask

raphtory-tests/tests/filter_tests/node_type_filter_index.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    let op = NodeTypeFilterOp::from_values(["Person"], &g);
+
+    match op.domain(storage) {
+        NodeList::List { elems } => {
+            let vids: Vec<_> = elems.iter().collect();
+            assert_eq!(vids.len(), 2);
+            assert!(vids.contains(&a));
+            assert!(vids.contains(&b));
+            assert!(!vids.contains(&c));
```
ljeub-pometry: the extra collect is not needed here?

raphtory/src/db/api/state/ops/filter.rs
```diff
@@ -392,18 +394,72 @@ where
...
+            .enumerate()
+            .filter_map(|(type_id, keep)| keep.then_some(type_id))
+            .collect();
+
+        NodeList::List {
+            elems: storage
+                .node_type_index()
+                .nodes_of_type(&type_ids)
+                .into_iter()
+                .collect(),
```
ljeub-pometry: You now have made a vec only to turn it back into an iterator to collect it back into a set. There is a chance the compiler can optimise it away in release mode maybe but we can trivially remove the first collect.

Ideally this wouldn't collect at all but we can figure that out in a future PR.

raphtory/src/db/graph/views/filter/model/node_filter/mod.rs
```diff
@@ -323,9 +322,10 @@ impl CreateFilter for NodeTypeFilter {
             .iter()
             .map(|k| self.0.matches(Some(k))) // TODO: _default check
             .collect::<Vec<_>>();
+
         Ok(NodeFilteredGraph::new(
-            graph,
-            TypeId.mask(node_types_filter.into()),
+            graph.clone(),
+            NodeTypeFilterOp::from_mask(node_types_filter.into(), graph),
```
ljeub-pometry: Constructing the filter only needs a reference to the graph, do it up front to avoid an extra clone.

raphtory/src/db/graph/nodes.rs
```diff
@@ -150,10 +150,13 @@ where
...
     pub fn node_list(&self) -> NodeList {
-        match self.nodes.clone() {
+        let list = match self.nodes.clone() {
             elems @ Index::Partial(_) => NodeList::List { elems },
             _ => self.graph.node_list(),
-        }
+        };
+
+        // TODO: Is this correct?
+        list.intersection(&self.predicate.domain(self.base_graph.core_graph()))
```
ljeub-pometry: this should already be handled when the predicate is applied, no need to do this on every call

raphtory/src/db/graph/nodes.rs
```diff
@@ -327,7 +331,12 @@ where
     }
 
     pub fn is_list_filtered(&self) -> bool {
-        !self.graph.node_list_trusted() || self.predicate.is_domain_filtered()
+        !self.graph.node_list_trusted()
+            || self.predicate.is_domain_filtered()
+            || !self // TODO: Is this correct?
```
ljeub-pometry: this should be handled by `self.predicate.is_domain_filtered()`?

ljeub-pometry (commented):
This seems to be mostly fine for now, make `nodes_of_type` return an iterator to avoid an extra allocation. We can leave the other allocation for now and can revisit that later.

db4-storage/src/segments/node/segment.rs
fabubaker: It's needed in the `NodeWriter::new` construction below.

db4-storage/src/segments/node_type_index/index.rs
fabubaker: Added a fast path for a single type_id, but that `collect` is needed to hold onto the guards from `self.map.get`.

db4-storage/src/segments/node_type_index/index.rs
fabubaker: As discussed offline, sticking with the `FxDashMap` for now.

raphtory-graphql/src/model/schema/node_schema.rs
fabubaker: As discussed, we still need the mask for use in `apply`.

raphtory-graphql/src/model/schema/node_schema.rs
fabubaker: As discussed, we still need the mask for use in `apply`.

raphtory/src/db/api/state/ops/filter.rs
fabubaker: As discussed we'll directly return an `IndexSet` from `nodes_of_type` and elide a `collect`.

db4-storage/src/api/node_type_index.rs
fabubaker: As discussed we'll directly return an `IndexSet` from `nodes_of_type` for now.

db4-storage/src/pages/locked/nodes.rs
fabubaker: They were a remnant from when the node type index was per segment instead of global, removed.

miratepuffin (approved):

PR #2747, opened 2026-08-28
Expose the auth surface an out-of-tree authorization plugin needs

shivamka1:

Four visibility changes, no behaviour: `pub mod auth`, `pub trait ContextValidation`,
`pub fn gql_error_with_code`, and the two `CODE_*` extension-code constants.

An authorization policy registered through `RegisterPlugin` lives outside this crate but guards
its resolvers, so it needs the same tools they use: `ContextValidation` for typed resolvers
(`require_jwt_write_access_dynamic` covers only dynamic ones), and the error helper with its code
constants so the plugin's denials carry the machine-readable codes the client already branches
on, rather than restating the strings and drifting.

These were public on the branch #2731 was developed on and lost visibility on the way to db_v4.
The auth plugin in the private repository already depends on them; without this it cannot compile
against db_v4.

No.

## How was this patch tested?

`cargo check --all-features` clean here; the private workspace that consumes the surface builds
against this commit.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/raphtory-graphql/src/auth.rs
+++ b/raphtory-graphql/src/auth.rs
@@ -436,7 +436,7 @@ fn roles_from_value(v: &serde_json::Value) -> Vec<String> {
     }
 }
 
-pub(crate) trait ContextValidation {
+pub trait ContextValidation {
     fn require_jwt_write_access(&self) -> Result<(), AuthError>;
 }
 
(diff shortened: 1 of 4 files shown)


ljeub-pometry:
equivalent changes were already merged

PR #2750, opened 2026-08-30
Add property indexing

miratepuffin:


raphtory/src/db/api/state/ops/filter.rs
```diff
@@ -224,12 +227,83 @@ impl<G> NodePropertyFilterOp<G> {
...
+    /// its exactness claim may be kept. Latest-flag candidates are a SUBSET
+    /// of what windowed or layer-restricted views need (a row's visible
+    /// latest can differ from its global latest), so restricted views fall
+    /// back to Ever candidates — a superset for every view — with exactness
+    /// off. `temporal().any()` is served by Ever directly; aggregating
+    /// chains are not served.
+    fn pushdown_semantics(&self) -> Option<(NodePropSemantics, bool)> {
+        use crate::db::api::view::internal::InnerFilterOps;
+        use crate::db::graph::views::filter::model::property_filter::Op;
+        let plain_view = !self.graph.window_filtered() && !self.graph.is_layer_filtered();
```
ljeub-pometry, line 307: We probably need per-layer index for a lot of queries? Also can handle Metadata separately as it is not affected by windowing and always has latest semantics?

raphtory/src/db/api/state/ops/filter.rs
```diff
@@ -224,12 +227,83 @@ impl<G> NodePropertyFilterOp<G> {
...
 }
 
 impl<G: GraphView> NodeOp for NodePropertyFilterOp<G> {
     type Output = bool;
 
-    fn domain(&self, _storage: &GraphStorage) -> NodeList {
+    fn domain(&self, storage: &GraphStorage) -> NodeList {
+        if let Some(candidates) = self.index_candidates(storage) {
+            // index candidates are ascending and deduplicated, as `from_sorted` requires
+            let list = NodeList::List {
```
ljeub-pometry, line 339: This materialises the list multiple times, any way to avoid this?

raphtory/src/db/api/state/node_state.rs
```diff
@@ -153,17 +223,69 @@ impl<K: Copy + Eq + Hash + Into<usize> + From<usize> + Send + Sync> Index<K> {
...
+            }
+            // a hash-set side carries no exactness claim
+            (Self::Sorted { keys: a, .. }, Self::Partial(b)) => Self::Sorted {
+                keys: a
+                    .iter()
+                    .copied()
+                    .filter(|k| b.contains(k))
+                    .collect::<Vec<_>>()
+                    .into(),
+                exact: false,
```
ljeub-pometry, line 269: this might be pessimistic for a subgraph, I think we can keep `exact` on an intersection

PR #2754, opened 2026-09-01
Node ID and OR filter bug

miratepuffin:

Fixes three filter-correctness bugs in Raphtory, all found while property-testing filter equivalence, and makes a filtered-read denial legible.

1. **Edge leak in `or` filters** (`filter.rs`) — an `or`-filtered node list was wrongly treated as trusted, so the edge iterator's fast path kept edges to excluded nodes.
2. **`.filter()` on a node collection dropped members** (`nodes.rs`) — `nodes.filter(name==…)` narrowed membership; it should keep every member (that's what `nodes[…]` is for). Membership is now read from the base graph, consistently across fields.
3. **A time view on an edge collection was ignored** (`edges.rs`) — `edges[before(t)]` (and `after`/`window`/`at`/`latest`) returned every edge instead of narrowing. It combined the filter against the unwindowed base graph, dropping the view; it now chains onto the current view like node collections do.

Plus: a filtered-read grant whose resolved filter can't be applied now denies with a clear, logged message instead of an opaque "internal error" (`data.rs`).

Plus: config-schema introspection — `GraphServer.config_schema()` in Python and `--print-config-schema` on the CLI return the full config schema as JSON: every field the server and every registered extension accept, including ones unset by default (`print_config` only shows set values). Backed by a new `config_schema` hook on server extensions (defaults to `to_json`).

All three produced silently wrong views; (3) failed open (an edge collection filtered by time processed every edge). None were caught because existing tests only assert on `.id` and only hit the paths that happened to work.

Behavioural fixes, no API change:

    nodes.filter(name=='a')       -> ['a','b']   (was ['a'])
    nodes[name=='a']              -> ['a']
    edges[Graph.before(t)]        -> narrowed     (was every edge)

Behaviour now matches the documented `.filter` vs `[...]` semantics; no doc changes needed.

New API surface: `GraphServer.config_schema()` (Python, documented in the stub) and the `--print-config-schema` server flag (documented in `--help`).

## How was this patch tested?

New unit and Python collection tests (`or`/and-of-`or` edge-endpoint invariants; node-collection membership; edge-collection filtering across every filter type and time view; access-filter denial logging), plus the full filter regression suite — all pass.

## Are there any further changes required?

Yes, tracked separately: time views don't compose reliably on edge collections — a *pure* time view is fixed here, but combining one with another filter (`and`/`or`/`not`) can still drop it and fail open. Same class as (3), in the composite filter path. Worth a parametrised suite over {collection path} × {filter type} × {and/or/not}, which currently has large gaps.


python/tests/test_base_install/test_filters/test_node_composite_filter_edge_endpoints.py
```diff
@@ -75,23 +75,56 @@ def check(graph):
     return check
 
 
+def _combo_offenders(graph, op, join):
```
ljeub-pometry: better name...

python/tests/test_base_install/test_filters/test_node_composite_filter_edge_endpoints.py
```diff
@@ -75,23 +75,56 @@ def check(graph):
...
         )
 
     return check
+
+
+@with_variants(_init)
+def test_and_of_or_keeps_edge_endpoints_and_node_set():
+    def check(graph):
+        # The shape behind the RBAC leak: a node-set restriction ANDed with a wide `or` that
+        # includes a node_type branch (whose domain is every node).
```
ljeub-pometry: clean up internal comments

raphtory/src/db/api/state/ops/filter.rs
```diff
@@ -301,21 +301,22 @@ where
...
+    fn const_value_in_domain(&self) -> Option<Self::Output> {
+        // An OR's domain is the union of its branches', so it is constant-true over that domain
+        // only when both branches are constant-true over theirs.
         match (
             self.left.const_value_in_domain(),
             self.right.const_value_in_domain(),
         ) {
-            (Some(true), _) | (_, Some(true)) => Some(true),
-            (Some(left), Some(right)) => Some(left || right),
+            (Some(true), Some(true)) => Some(true),
```
ljeub-pometry: this is pessimistic, if the domain of the constant filter is a superset of the domain of the non-constant filter, the filter is actually constant (`true` || `false` = `true`).

raphtory/src/db/api/state/ops/filter.rs
```diff
@@ -406,11 +407,105 @@ impl NodeTypeFilterOp {
...
+            NodeList::All
+        }
+        fn apply(&self, _storage: &GraphStorage, _node: VID) -> bool {
+            true
+        }
+        fn const_value(&self) -> Option<bool> {
+            self.cv
+        }
+        fn const_value_in_domain(&self) -> Option<bool> {
+            self.cvid
```
ljeub-pometry, line 470: With domain set to `NodeList::All`, `const_value` and `const_value_in_domain` should always be the same (the domain is everything), so this `Stub` is weird. There is a chance this test might start failing if we optimise things somewhere, even if the result is still correct (member is effectively constant everywhere). Change the domain here to be non-trivial to avoid this problem.

raphtory/src/db/api/state/ops/filter.rs
```diff
@@ -406,11 +407,105 @@ impl NodeTypeFilterOp {
...
+        };
+        assert_eq!(both.const_value_in_domain(), Some(true));
+
+        let widened = OrOp {
+            left: wide(),
+            right: member(),
+        };
+        assert_eq!(widened.const_value_in_domain(), None);
+
+        // The nested three-branch shape that triggered the RBAC edge leak.
```
ljeub-pometry: remove internal comments

ljeub-pometry (changes requested):
the `or` filter is now too pessimistic about its const-ness (which is correct but inefficient)

python/tests/test_base_install/test_filters/test_edges_collection_filter.py
```diff
@@ -0,0 +1,74 @@
...
+        "and": weight_gt_5 & Graph.layer("work"),
+        "or": (Edge.property("weight") > 15) | (Edge.src().name() == "a"),
+        "not": ~weight_gt_5,
+        "and_of_or": Graph.layer("work")
+        & ((Edge.property("weight") > 15) | (Edge.src().name() == "a")),
+    }
+
+
+def _edges(collection):
+    return sorted((e.src.name, e.dst.name) for e in collection)
```
miratepuffin: This can be e.id

python/tests/test_base_install/test_graphql/test_remote_graph_transport.py
```diff
@@ -2066,14 +2066,19 @@ def test_filter_nodes_preserves_membership():
         assert all_ids == ["alice", "ben", "bob", "hamza"]
 
 
-def test_filter_nodes_narrows_on_node_id():
-    """#2690: a node-id filter (name/id) is applied as a graph view, so it DOES
-    narrow collection membership — unlike the property filter above. Pins the
-    new behavior (and its parity with local) so a regression is visible."""
+def test_filter_by_node_id_keeps_membership_but_getitem_narrows():
+    """`.filter()` keeps every member for a name/id filter too — `[]` is the one that narrows.
```
miratepuffin: .filter only filters to the right, can also add the case where we do g.filer.nodes.id here

raphtory-graphql/src/python/server/server.rs
```diff
@@ -133,6 +133,15 @@ impl PyGraphServer {
         Ok(PyGraphServer(server))
     }
 
+    /// The full config schema as a nested dict: every field, including ones unset by default.
+    ///
```
miratepuffin, line 137: Need to add this to the PR description

raphtory-graphql/src/cli.rs
```diff
@@ -232,8 +238,14 @@ where
         Commands::Server(server_args) => {
             let port = server_args.port;
             let print_config = server_args.print_config;
+            let print_config_schema = server_args.print_config_schema;
             let server = GraphServer::new_from_args(server_args).await?;
-            if print_config {
+            if print_config_schema {
+                let schema = server.config().config_schema_json().map_err(|e| {
+                    std::io::Error::new(std::io::ErrorKind::InvalidData, e.to_string())
```
miratepuffin: tidy this up

raphtory/src/db/graph/nodes.rs
```diff
@@ -150,9 +150,14 @@ where
     }
 
     pub fn node_list(&self) -> NodeList {
+        // Membership is the candidate set (`self.nodes` / the base graph), filtered per node by
```
miratepuffin: don't need to this comment

shivamka1 (approved):

PR #2764, opened 2026-09-04
Graphql Scheduling changes and update benchmark

miratepuffin:

Two GraphQL server changes plus benchmark coverage:

1. **Cold graph loads off the query path, and bounded** — graphs decode on the blocking-IO pool
   via `blocking_load`, gated by a semaphore (`[concurrency] max_concurrent_loads`, default
   `cores / 4`, at least 2). A multi-second load no longer stalls queries or health, and a burst
   of cold loads can no longer hold an unbounded number of decoding graphs in memory at once.

2. **Read-only mode** (`[cache] read_only = true`) — every graph entering the cache is wrapped
   with the new `MaterializedGraph::read_only()`: storage segments are read-locked once for the
   server's lifetime, so every read on every path skips per-access segment locking. Graph
   mutations fail fast with the existing locked-graph error; role/permission management is
   unaffected.

Benchmark harness (`graphql-bench`): a generated 5M-node graph (string ids), an overlapping
heavy-scan + short-query scenario pair (`-e SCHEDULING_ONLY=1` runs just the pair), three
algorithm scenarios (single-source shortest path, PageRank, degree centrality), and an `-e ONLY=`
filter for running scenario subsets. The under-load pair reports steady completed rate instead of
the p95-gated max rate, which is meaningless under deliberate saturation. Note the three algorithm
ramps add ~15 minutes to the benchmark job.

- Health checks and short queries stalled behind cold graph loads: during one cold load,
  `/health` took the full load duration (measured 6.1s, returning 503) and 3ms after the change.
- Concurrent cold loads previously decoded with unbounded concurrency, so peak memory was
  concurrent decodes × graph size.
- Read-heavy deployments with no updates pay per-access lock traffic on every read: profiling a
  point-read mix showed ~27% of on-CPU compute time in shared-lock acquisition, capping the test
  box at ~2.2k req/s. In read-only mode the same box served a ramp to 24k req/s with p95 <1ms
  and zero shed load, and short queries hold p95 ~100ms under 24 concurrent full-graph scans.

Yes: two new config options, `[cache] read_only` (default `false`) and
`[concurrency] max_concurrent_loads` (default `cores / 4`), documented on the config fields. In
read-only mode graph mutations return the existing "graph is locked and cannot be mutated" error.

- New tests: the concurrent-load bound, and read-only cache behaviour (reads served, mutations
  rejected, writable control). The existing deadlock tests are unchanged and passing; the full
  `raphtory-graphql` suite is green.
- k6: the scheduling pair under 24 scan VUs (shorts p95 ~100ms, 100% success), the read-mix ramp
  comparison above, and the new algorithm scenarios (max sustainable rates on the reference
  graph: degree centrality 381 req/s, single-source shortest path 250 req/s, PageRank 68 req/s).
- Server-behaviour tests covering read-only mode with auth filters, probes, and algorithms run
  in both modes in the downstream suite.

- `nodes.page(offset, …)` reaches its offset via `iter().skip(offset)`, paying a locked node
  read per skipped element (~26k reads for a random offset on the reference graph); indexing
  directly into the node list when no filter is active is the follow-up with the largest
  measured win for paged reads.
- Outside read-only mode, point reads still take a segment read lock per node access;
  amortising the guard per query is the broader follow-up.

### DIFF
--- a/graphql-bench/process-k6-output.py
+++ b/graphql-bench/process-k6-output.py
@@ -34,8 +34,21 @@ def find_max_rate(scenario_name):
     # perf = valid_trend.loc[max_rate_index]
     # return pd.Series(perf, name=scenario_name)
 
+# The scheduling pair runs under deliberate saturation, so the p95-gated max-rate is meaningless
+# for it; report the steady completed rate instead (collapses if short queries starve).
+UNDER_LOAD_SCENARIOS = {"heavy_load", "short_queries_under_heavy_load"}
+
+def steady_rate(scenario_name):
+    scenario = output[output["scenario"] == scenario_name]
+    req_duration = scenario[scenario["metric_name"] == "http_req_duration"]
+    rate = req_duration["metric_value"].resample('1s').count()
+    return rate.iloc[10:].mean() # discard first 10 seconds
+
 scenarios = output["scenario"].unique()[1:] # first element is nan
-max_rates = [find_max_rate(scenario) for scenario in scenarios]
+max_rates = [
+    steady_rate(scenario) if scenario in UNDER_LOAD_SCENARIOS else find_max_rate(scenario)
+    for scenario in scenarios
+]
 
 results = [{"name": name, "unit": "req/s", "value": value} for (name, value) in zip(scenarios, max_rates)]
 df = pd.DataFrame(results)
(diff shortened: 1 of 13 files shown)


fabianmurariu (approved):
LGTM

Issue #2728, opened 2026-08-21
Filter-expression asymmetry: nodes[edge_expr] vs edges[node_expr]


shivamka1:
`edges[node_expr]` is accepted (edges whose endpoints match a node predicate) but `nodes[edge_expr]` raises `Exception("Node filter expected")`: the same kind of cross-kind question, answered in one direction only. The remote matches local exactly here (same acceptance, same refusal), so this is a local-semantics decision, not a client fix.

Two symmetric options:
- implement `nodes[edge_expr]` as "nodes with at least one matching edge" (the mirror of how `edges[node_expr]` treats endpoints), or
- reject both directions.

Raised as a question in review on the parity suite PR (#2707): "shouldn't node and edge properties support the same filters?"

Issue #2718, opened 2026-08-19
Negating a view-scoped filter is a no-op; the intended semantics are undefined


shivamka1:
Raised from review on #2707: https://github.com/Pometry/Raphtory/pull/2707#discussion_r3765462678

### What

A filter expression that scopes to a view can be negated, and the negation currently does nothing — the result is the same as the un-negated filter. The parity suite pins that behaviour so it cannot change silently, but nobody has decided what it *should* mean.

### The question

From the review:

> we might need to think about what this is supposed to actually mean (it is weird that this seems to be a no-op), should this be equivalent to excluding all events that are in the view from the filtered graph?

That reading — negation excludes the events the view selects — seems the most natural, but it is a semantics decision rather than something a test PR should settle, so the current no-op is pinned as-is.

### Why it matters

A no-op negation is a silent wrong answer: a caller who writes it gets an unfiltered result rather than an error, so the mistake is invisible. Either the operation should mean something specific, or it should be rejected.

### Suggestion

Decide between:

1. **Define it** — negating a view scope excludes the events in that view, and implement it on both sides.
2. **Reject it** — refuse the expression at build time, so a caller finds out immediately instead of getting an unfiltered result.

Whichever is chosen, the parity test that currently pins the no-op should be updated to pin the new behaviour.

shivamka1:
Verified against the current branch (post-db_v4 merge), and the behaviour is worse than this issue states. The description says the negation "does nothing — the result is the same as the un-negated filter", but those are two different outcomes and it is neither the second one:

```python
g = Graph()
g.add_edge(1, "a", "b"); g.add_edge(5, "c", "d"); g.add_edge(9, "e", "h")
view = filter.Graph.window(0, 6)

g.edges                       # a->b, c->d, e->h
g.filter(view).edges          # a->b, c->d          (filter applied)
g.filter(~view).edges         # a->b, c->d, e->h    (the WHOLE graph)
```

So negating a view scope discards the filter entirely and yields the unfiltered graph — not the same result as the un-negated filter, but the same result as passing no filter at all. A caller who writes `~view` expecting "everything outside the window" (`e->h` alone) silently gets everything instead, which is the maximally wrong answer rather than a merely surprising one.

Local and remote agree exactly here, so this is core semantics, not a client bug — `NotFilteredGraph`'s `filter_graph_view` returns the graph unchanged.

[...]

PR #2767, opened 2026-09-04
Fix filter correctness: layer aggregation and composite lowering

shivamka1:

Filter correctness across layers and composites. Five fixes and the test changes that pin them.

**1. `is_deleted` aggregates over an edge's layers** (Pometry/pometry-storage#378)

`IsDeletedGraph` implemented the predicate as a *per-layer* filter, and an edge passes a layer filter when **any** of its layers passes — so `edges[is_deleted()]` answered "some layer has a deletion" while `EdgeView::is_deleted` answers "no layer still holds this edge alive". The two diverge as soon as an edge's layers disagree, which the default delete path produces on its own: `delete_edge(t, src, dst)` with no layer always tombstones `_default`, leaving a layered edge alive on its own layer and deleted on another. On a `PersistentGraph` that also broke the `valid`/`deleted` partition — the same edge appeared in both. It now implements the whole-edge filter, which receives the view's full layer set.

The state itself is deliberate — recording a deletion on a layer an edge was never added to is what lets an out-of-order stream see a deletion before its addition — so the fix is in how the readers aggregate, not in rejecting the write.

**2. Filter composition on edge collections** (Pometry/pometry-storage#371)

`edges[...]` lowered a composite by nesting wrapper graphs, one per operand. That cannot express `or` or `not`, and loses a view operand entirely: a wrapper inherits its time semantics from the graph it wraps, so a windowed operand contributed no restriction at all, and each wrapper answers five filter hooks (`internal_filter_edge`, `internal_filter_edge_layer`, `internal_nodes_filtered`, the exploded pair) which then have to agree with each other for every combination of operand kinds. They did not.

A composite now lowers to a single per-edge boolean: each operand is asked the one question that already has a correct answer — "is this edge in your view?", via the composed `FilterOps::filter_edge` — and `&`, `|`, `~` combine those answers. One wrapper graph carries the result and inherits every other hook, so no sibling hook is left to contradict it. This is what `nodes[...]` already does with node ops, and why that path was the only one obeying set algebra.

Measured against set algebra over 69 (path, shape, operand-kind) groups that disagreed before: `edges[...]` now disagrees on **none**.

    edges[window(3,12) & weight>5]   ['bc']                    was every matching edge
    edges[weight>5 | layer('y')]     the union                 was every edge
    edges[~window(3,12)]             the complement            was every edge
    edges[~(weight>5 & layer('y'))]  the complement            was ~(weight>5) alone

**3. A node expression is one node test, lifted whole**

A node filter lifts onto an edge as "both endpoints pass". The question was how a *composite* of node predicates lifts. Lowering each leaf separately and composing at edge level makes `~` the complement of the matching edge set and `|` the union of two edge sets, which reads `~(Node.name == "a")` as "not (both endpoints are a)" — every edge merely touching `a` — and drops an N1-to-N2 edge from `N1 | N2`. A whole node expression is instead evaluated as a node set, and the edges are those whose endpoints both lie in it:

    edges[~(Node.name == "a")]   edges not touching a        was every edge
    edges[N1 | N2]               edges among N1-or-N2 nodes  was only edges with both ends in one branch

`edges[expr]` and `filter(expr).edges` agree on this, and so do `~expr` and `node: {not: expr}` — the expression-level tree collapses into the composite before lowering, and the two spellings previously disagreed, which the differential parity suite caught. This reverses the reading proposed as the fix in #371; the complement was the wrong target.

**4. `graph.filter(...)` lowers composites the same way, as an induced subgraph**

The entity path still nested wrapper graphs. It now builds one graph from what the expression says about nodes and about edges, read off the kinds of its leaves (`LeafKinds`). A node expression installs a node test and no edge test, so it behaves exactly like a single node filter: edges follow from their endpoints, and `node(x).filter(expr).out_neighbours` does not test `x`. An edge expression leaves nodes alone. Anything else installs both, and then nodes are decided first — an edge predicate is *unknown* for a node, so it excludes none — after which the edges are the edge-level set algebra induced on the nodes kept. The endpoint check every filtered graph performs stays on, so a filtered graph never holds an edge whose endpoint it dropped.

Four consequences worth calling out:

    filter(window(3,12) & weight>5).edges   narrowed          was: window ignored
    filter(P & N).nodes                     N's nodes         was: every node
    filter(~(P & N)).nodes                  every node        was: ~N's nodes, contradicting ~P | ~N
    filter(~window(0,6))                    what lies outside was: the whole graph (Pometry/Raphtory#2718)

Negating a view is therefore no longer a no-op. It is entity-level, matching the membership-only rule below: an entity with events on both sides of a window is excluded rather than kept with its outside events.

**5. Per-node and nested edge collections test both endpoints** (#371 class 5)

`node(x).edges[N]` and `nodes.edges[N]` never tested `x` itself. The select graph for a node expression was the node-filtered wrapper, and iterating a node's edges through one checks only the far endpoint — the traversal rule, where the node walked from is exempt. A collection subscript asks whether each *edge* passes, and an edge passes a node test when both endpoints do. `node("a").edges[N]` returned both of `a`'s edges when `a` failed `N`; it now returns none. Composites already took the boolean route and were unaffected.

All five classes in #371, and the negated-composite shapes recorded in its comments, are now fixed.

All of these produced silently wrong results, and all but one failed **open** — a filtered edge collection processed edges the filter should have excluded, which matters most for the access-control shapes (`property & window`) that motivated the audit. The `is_deleted` divergence made either answer unsafe to rely on, since the filter and the method are two spellings of one question.

Behavioural fixes, no API change for callers. Answers that change:

    # PersistentGraph, edge added on layer "work" then deleted with no layer named
    g.edges[Edge.is_deleted()]              -> []                     was [the edge]

    # any composite containing a view, on an edge collection
    edges[window(1,5) & pred]               -> narrowed               was: window ignored
    edges[~view], edges[a | b]              -> set algebra            was: every edge

    # a node expression applied to edges, every spelling and both paths
    edges[~N], filter(~N).edges             -> edges among the nodes failing N
                                               was: both endpoints fail / every edge touching one
    edges[N1 | N2]                          -> edges among N1-or-N2 nodes
                                               was: only edges with both ends in one branch
    edges[node: {not: ...}]                 -> same as ~N             was: a different answer to ~N

    # graph.filter with a composite
    filter(P & N).nodes                     -> N's nodes              was: every node
    filter(~(P & N)).nodes                  -> every node             was: ~N's nodes
    filter(~window(0,6))                    -> what lies outside      was: the whole graph

    # per-node and nested edge collections
    node("a").edges[N]                      -> [] when a fails N      was: all of a's edges
    nodes.edges[N]                          -> matches filter(N).nodes.edges

The node-expression rows are visible to GraphQL clients relying on the previous reading, and reverse the target stated in Pometry/pometry-storage#371 — the complement was the wrong reading, since it keeps every edge merely touching a failing node.

* Full Python suite 3362 passed / 17 skipped / 9 xfailed; filter suite 588 passed; `cargo test -p raphtory --lib` 70 passed; `cargo test -p raphtory-graphql` 254 passed; `cargo check --features python` and `cargo +nightly fmt` clean.
* Set algebra over every (path, shape, operand-kind) group on both graph models: **69 disagreeing groups before, none after** — `edges[...]`, `nodes[...]`, `filter().edges` and `filter().nodes` alike.
* Expectations are derived from a symbolic mirror of each expression rather than enumerated, so the negated-composite cells `~(A & view)` and `~(A | view)` are covered by construction. Those were reported separately on #371 and could not be reached by a pairwise grid.
* The expectations never come from the path under test — views reference the equivalent chained view, predicates are evaluated directly — so a filter that failed open cannot make its own expectation agree, and a case whose expected answer is the whole collection is skipped as non-discriminating.
* New invariants: a filtered graph never holds an edge whose endpoint it dropped; a node expression does not test the node stood on, for a composite exactly as for a single filter; an edge predicate neither admits nor excludes a node, so `~(P & N)` and `~P | ~N` agree.
* Local and remote compared across the composite shapes with no divergences. The parity suite's "universal" pins for `node | edge-pred`, `view | view` and `~view` moved into the discriminating matrix, since those expressions now exclude something; the layer case asserts the same set algebra against the un-negated view.
* Existing tests changed rather than deleted where they pinned the old behaviour: the `is_deleted` expectations are split by graph model, and every known-broken pin is gone because the classes it guarded are fixed.

Not for #371 — all five classes are fixed. Two semantics questions are left open deliberately, both narrowed by this work and neither blocking:

* **Pometry/Raphtory#2718** — negating a view is now entity-level. Whether it should be *event-level*, excluding the events inside the view and keeping an entity's events outside it, is undecided. It is not specific to negation: it is the membership-only rule for every composite containing a view, and `~snapshot_at(t)` has no event-level reading at all.
* **Pometry/Raphtory#2728** — `nodes[edge_expr]` still raises where `edges[node_expr]` is accepted, and `filter(P).nodes` now answers the same question. Three readings are defensible; the choice belongs on that issue.

A composite containing a view defines **membership** only; the events of the entities it admits come from the base graph, since a union or complement of time semantics is not definable. Single views still clip history exactly as before. Two cases keep their exact semantics rather than lowering to a boolean: a single filter, and a composite of exploded-edge filters, which narrow which *events* of an edge survive.


shivamka1:
Added two commits carrying over what was worth keeping from #2768 (now closed as a duplicate of the suites that came in with #2754):

**`Reference the filter set algebra independently, and cover negated composites`**

* The set-algebra references in `test_edges_collection_filter.py` and `test_nodes_collection_filter.py` were computed as `graph.edges[atom]` / `graph.nodes[atom]` — i.e. through the subscript those expectations exist to check. Where a single filter fails open, `EVERYTHING & X == X`, so a combination that dropped a term matches its expectation and the pins report a live bug as fixed. Views now reference the equivalent chained view and predicates are evaluated over the collection directly. Verified behaviour-preserving: all 18 atoms produce identical sets to the previous derivation on both graph models, so this changes what the expectations *depend on*, not what they assert.
* `~(A & B)` and `~(A | B)` are now generated as cases, with two representatives added to the pinned classes. The pairwise rules only ever negated a single atom, so the shapes recorded on Pometry/pometry-storage#371 — where negating a composite drops a view operand and `~(A & view)` degenerates to `~A` — had no coverage at all.

One weakness left deliberately untouched, worth a follow-up: in `test_nodes_collection_filter.py`'s fixture the view atoms select *every* node, so combinations involving them are partly vacuous (`name & window == name` regardless). Fixing that means reshaping a fixture several tests share, which does not belong in this PR.

Full suite: 3351 passed / 17 skipped / 9 xfailed. Filter suite: 585 passed.

raphtory/src/db/graph/views/filter/mod.rs
```diff
@@ -101,9 +231,97 @@ pub trait CreateFilter: Sized {
...
+
+    /// Whether this filter tests individual edge *events* rather than edges.
+    ///
+    /// Exploded-edge filters do, and composites of them must keep their wrapper
+    /// graphs: a per-edge boolean can only say whether an edge is in, not which
+    /// of its events survive.
+    fn is_exploded_edge_filter(&self) -> bool {
+        false
+    }
+
```
ljeub-pometry: we shouldn't have all these boolean flags, this needs to be handled properly

raphtory/src/db/graph/views/filter/mod.rs
```diff
@@ -101,9 +231,97 @@ pub trait CreateFilter: Sized {
...
+    /// boolean — while a filter that stands alone builds its wrapper graph and
+    /// asks whether an edge is in its view ([`leaf_filter_lowering!`]
+    /// writes that one out). A default could only be one of the two, and every
+    /// type that wanted the other would compile silently and be wrong: the
+    /// composition it was supposed to fix would quietly come back. Leaving it
+    /// required turns each of those into a compile error instead.
+    ///
+    /// Boxed rather than an associated type so an implementation is one line
+    /// instead of a type definition.
+    fn create_edge_filter<'graph, G: GraphView + 'graph, F: GraphView + 'graph>(
```
ljeub-pometry: give this an associated type, same as `create_node_filter`

raphtory/src/db/graph/views/filter/edge_op.rs
```diff
@@ -0,0 +1,224 @@
+//! Boolean composition of edge filters, as a per-edge boolean rather than nested graphs.
+//!
+//! A filter expression can be lowered two ways. Composing *graphs* — one wrapper
+//! per operand, each narrowing the last — is how the entity path works, and it
+//! cannot express `or` or `not`: a wrapper inherits its time semantics from the
+//! graph it wraps, so a view operand's restriction is simply absent from the
+//! composite, and each wrapper answers several hooks (`internal_filter_edge`,
+//! `internal_filter_edge_layer`, `internal_nodes_filtered`, the exploded ones)
```
ljeub-pometry: we need to fix the timesemantics for `AndFilteredGraph`, `OrFilteredGraph` and `NotFilteredGraph`, they are just plain wrong right now

raphtory/src/db/graph/views/filter/model/or_filter.rs
```diff
@@ -60,12 +63,33 @@ impl<L: CreateFilter, R: CreateFilter> CreateFilter for OrFilter<L, R> {
...
-        let l = self.left.filter_graph_view(filtered.clone())?;
-        let r = self.right.filter_graph_view(filtered.clone())?;
-        let left = self.left.create_filter(filtered.clone(), l)?;
-        let right = self.right.create_filter(filtered, r)?;
-        Ok(OrFilteredGraph { graph, left, right })
+    ) -> Result<Self::EntityFiltered<'graph, G, F>, GraphError>
+    where
+        Self: 'graph,
+    {
+        if self.is_exploded_edge_filter() {
```
ljeub-pometry: We should fix the `OrFilteredGraph` to actually work properly instead of all these hacks which will still have many broken corner cases. The `TimeSemantics` for the `OrFilteredGraph` should be the union of the windows of both sides. Overlapping windows can be merged into a single window such that the MultiWindow timesemantics only needs to support non-overlapping sorted windows for efficiency (all iterators can just be chained instead of merged). This method itself basically goes back to what it was before.

PR #2768, opened 2026-09-04
Check filter composition against set algebra

shivamka1:

A test module that checks filter *composition* against set algebra on every collection path, and records exactly which combinations are currently wrong.

Filtering is spelled several ways — `nodes[expr]`, `edges[expr]`, `graph.filter(expr)` — and each lowers an expression differently. They agree on single predicates and diverge on composites, so the module is the cross-product {atom kind} × {`&`, `|`, `~`, `~(A & B)`, `~(A | B)`} × {path}, over both graph models.

Two rules make the expectations mean something:

* **The oracle never comes from the path under test.** Each atom carries a reference set built another way — a view atom from the equivalent chained view (`graph.window(a, b).edges`), a predicate atom by evaluating it in Python over the collection — and composite expectations are set algebra over those. Deriving them from `edges[atom]` instead lets the suite agree with itself: where a single filter fails open, `EVERYTHING & X == X` is equally consistent with a correct `and` and with one that dropped a term.
* **A case that cannot discriminate is not a case.** Combinations whose correct answer is the whole universe are skipped, because an implementation that ignores the filter entirely would also match. Atoms whose reference set cannot discriminate on a given side are left out of the cases rather than silently weakening them — `layer` on the node side, because a layer view filters edges and keeps every node.

What it asserts:

* `nodes[expr]` obeys set algebra for **every** combination — positively, per case. That path lowers composites to per-node boolean operations, so `&` is intersection, `|` is union and `~` is complement by construction. It is both a permanent regression guard and the behaviour the other paths should reach.
* The two edge spellings agree with each other for every shape — including the shapes both get wrong, which is precisely why agreement between them proves nothing on its own.
* Single views narrow on every path, compared against the same view spelled as a chain.
* `~~A == A` for predicate atoms; node collections reject edge predicates with an error rather than quietly selecting everything.
* **A ledger**: the set of (path, shape, atom kinds) groups that disagree with set algebra is compared against `KNOWN_DISAGREEMENTS` — 69 groups today. It fails in both directions, naming the groups that moved.

The filter bugs found recently were all in composites, and the existing tests could not see them: they vary the *operators* but never the *operand kinds*, so every composite tested is property × property — which works. Swap one operand for a view and it breaks. Their expectations are also hardcoded lists written by observing output, which pin whatever the code did that day and cannot distinguish "correct" from "consistently wrong".

Set algebra is an oracle that needs no second implementation to compare against, and it is what found each of these. This module makes that check systematic, so the classes in Pometry/pometry-storage#371 have a scoreboard: as `create_edge_filter` lands, entries leave the ledger and the corresponding groups become positively asserted.

Every one of the 69 recorded groups maps to a mechanism already described in #371 — `and` with a time view, `or` with any view or mixed kinds, `~view`, `~node-filter` — or to the negated-composite shapes (`~(A & view)`) recorded on that issue, which a pairwise audit cannot reach because it only negates single atoms. `nodes[expr]` appears nowhere in the ledger.

No. Tests only.

* The module lands green: 62 passed, 4 skipped. Whole filter suite: 647 passed, 4 skipped.
* The ledger was checked in both directions by removing an entry and confirming it fails with the group named, rather than passing quietly.
* The discrimination guards were checked by construction: `layer` on the node side is excluded because its reference is the whole universe, and a test records that exclusion and its reason so a fixture change cannot quietly drop coverage.

Yes — the fix itself. Pometry/pometry-storage#371 proposes a per-edge boolean lowering (`create_edge_filter`) mirroring what `nodes[expr]` already does, which addresses every group in this ledger at once. That needs sequencing against the filter-expression rebase in #2753. Merging this first means the fix arrives with its acceptance criteria already in the tree.

### DIFF
--- a/python/tests/test_base_install/test_filters/test_filter_algebra.py
+++ b/python/tests/test_base_install/test_filters/test_filter_algebra.py
@@ -1,0 +1,533 @@
+"""Filter algebra across every shape and every collection path.
+
+Filtering is spelled several ways — `nodes[expr]`, `edges[expr]`,
+`graph.filter(expr)` — and the spellings lower an expression differently. They
+agree on single predicates and diverge on composites, so the interesting surface
+is the cross-product {atom kind} x {`&`, `|`, `~`, `~(A & B)`, `~(A | B)`} x
+{path}, checked against set algebra.
+
+Two rules keep the expectations trustworthy, both learned from filter bugs that
+the existing tests could not see:
+
+**The oracle never comes from the path under test.** Each atom carries a
+reference set built another way: a view atom from the equivalent chained view
+(`graph.window(a, b).edges`), a predicate atom by evaluating it in Python over
+the collection. Composite expectations are set algebra over those references.
+Deriving them from `edges[atom]` instead would let the suite agree with itself —
+where a single filter fails open, `EVERYTHING & X == X` is equally consistent
+with a correct `and` and with one that dropped a term.
+
+**A case that cannot discriminate is not a case.** Where the correct answer is
+the whole universe, an implementation that ignores the filter entirely still
+matches, so those combinations are skipped rather than counted. Real bugs have
+hidden behind exactly that coincidence.
+
+`nodes[expr]` is asserted positively for every combination: it lowers composites
+to per-node boolean operations, so set algebra holds there by construction, and
+it is the behaviour the other paths should eventually match.
+
+The other paths compose wrapper graphs instead, which loses a view operand's
+restriction, and they are currently wrong for many composites (see
+Pometry/pometry-storage#371). Rather than a pin per case, the disagreeing
+*groups* are recorded in `KNOWN_DISAGREEMENTS` and compared as a set, so this
+module lands green, states the boundary exactly, and fails with the specific
+groups that moved when it changes in either direction.
+"""
+
+from collections import defaultdict
+from itertools import combinations
+
+import pytest
+from raphtory import Graph, PersistentGraph, filter
+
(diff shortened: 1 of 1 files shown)


shivamka1:
Stacked on #2767 (the `is_deleted` aggregation fix), so the base is `fix/filter-followups` for now and the diff shows only the new module. GitHub will retarget this to `db_v4` automatically once #2767 merges.

The dependency is real rather than administrative: the ledger records `is_deleted`-adjacent behaviour, so generating it against a build without #2767 would bake in the pre-fix answers.

shivamka1:
Closing: this substantially duplicates the collection-filter suites that came in with #2754.
#2754 already covers combinations against set algebra on both edge paths *and* both node paths, plus the nested and per-node collections this module lacks, over 19 atoms rather than 7 and both graph models.

Three things here were genuinely new, and only two survive review:

* **independent oracles** — #2754 derives its reference sets from the path under test (`single = {n: frozenset(graph.nodes[e].name)}`), so on a build where a single filter fails open the derived expectations collapse and the pins misreport. Moving to #2767 alongside the degeneracy guard already there.
* **`~(A & B)` / `~(A | B)`** — #2754 only negates single atoms, so it cannot reach the shapes recorded on Pometry/pometry-storage#371 where negating a composite drops a view operand. Also moving to #2767.
* **a group-level ledger** — dropped: #2754's per-class rules and pin test already record the boundary and already fail when a class is fixed.

No loss of coverage, one less parallel suite to keep in step.

PR #2771, opened 2026-09-08
Aggregate is_deleted over an edge's layers, and reference filter set algebra independently

shivamka1:

Split out of #2767: the `is_deleted` fix and the test hardening that does not depend on the filter-lowering work, so they can land while the composite lowering is redesigned.

**1. `is_deleted` aggregates over an edge's layers** (Pometry/pometry-storage#378)

`IsDeletedGraph` implemented the predicate as a *per-layer* filter, and an edge passes a layer filter when **any** of its layers passes — so `edges[is_deleted()]` answered "some layer has a deletion" while `EdgeView::is_deleted` answers "no layer still holds this edge alive". The two diverge as soon as an edge's layers disagree, which the default delete path produces on its own: `delete_edge(t, src, dst)` with no layer always tombstones `_default`, leaving a layered edge alive on its own layer and deleted on another. On a `PersistentGraph` that also broke the `valid`/`deleted` partition — the same edge appeared in both. It now implements the whole-edge filter, which receives the view's full layer set.

The state itself is deliberate — recording a deletion on a layer an edge was never added to is what lets an out-of-order stream see a deletion before its addition — so the fix is in how the readers aggregate, not in rejecting the write. `ValidGraph` and `IsActiveGraph` stay per-layer, which is correct for them: "alive on some layer" and "active on some layer" are the intended readings.

**2. Set-algebra expectations no longer come from the path under test**

The edge- and node-collection suites derived their expected sets by reading single filters back through the same subscript they were checking. Where a single filter fails open, `EVERYTHING & X == X`, so a combination that silently dropped a term matched its own expectation. References now come from the equivalent chained view for view atoms and from direct evaluation for predicates, a degenerate baseline (a single filter selecting everything or nothing) is rejected up front, and the negated-composite shapes `~(A & view)` / `~(A | view)` are covered explicitly — they were reported on Pometry/pometry-storage#371 as a cell the pairwise grid cannot reach.

The combination classes that are broken on `db_v4` stay pinned as broken, by a discriminating representative each; those pins fail with instructions when a fix lands.

The `is_deleted` divergence made either answer unsafe to rely on, since the filter and the method are two spellings of one question. The oracle change is what let the remaining filter-composition bugs be measured honestly rather than masked.

One answer changes:

    # PersistentGraph, edge added on layer "work" then deleted with no layer named
    g.edges[Edge.is_deleted()]        -> []            (was [the edge])

On an event graph `is_deleted` still reports that a deletion event exists, regardless of which layer recorded it; the two models expect different results and the tests are split accordingly.

* New `test_deletion_layer_semantics.py`: the deletion's layer relative to the edge's layers, crossed over the three layouts and both graph models.
* `test_edge_filter.py` / `test_edge_filter_gql.py`: `is_deleted` expectations split by graph model.
* Collection suites: independent references, degenerate-baseline guard, negated-composite rows.
* Full Python suite 3355 passed / 17 skipped / 9 xfailed; filter suite 585 passed; `cargo test -p raphtory --lib` 83 passed; `cargo check --features python` and `cargo +nightly fmt` clean.

The composite filter lowering (#371 classes 1–5) is being redesigned in #2767 around proper time-semantics composition and will land separately.

### DIFF
--- a/raphtory/src/db/graph/views/is_deleted_graph.rs
+++ b/raphtory/src/db/graph/views/is_deleted_graph.rs
@@ -5,20 +5,17 @@ use crate::{
                 InheritEdgePropertySchemaOps, InheritNodePropertySchemaOps, InheritPropertiesOps,
             },
             view::internal::{
-                EdgeTimeSemanticsOps, Immutable, InheritEdgeFilterOps, InheritEdgeHistoryFilter,
-                InheritExplodedEdgeFilterOps, InheritLayerOps, InheritListOps, InheritMaterialize,
-                InheritNodeFilterOps, InheritNodeHistoryFilter, InheritStorageOps,
-                InheritTimeSemantics, InternalEdgeLayerFilterOps, Static,
+                EdgeTimeSemanticsOps, Immutable, InheritEdgeHistoryFilter,
+                InheritEdgeLayerFilterOps, InheritExplodedEdgeFilterOps, InheritLayerOps,
+                InheritListOps, InheritMaterialize, InheritNodeFilterOps, InheritNodeHistoryFilter,
+                InheritStorageOps, InheritTimeSemantics, InternalEdgeFilterOps, Static,
             },
         },
         graph::views::layer_graph::LayeredGraph,
     },
     prelude::GraphViewOps,
 };
-use raphtory_api::{
-    core::entities::{LayerId, LayerIds},
-    inherit::Base,
-};
+use raphtory_api::{core::entities::LayerIds, inherit::Base};
 use raphtory_storage::{core_ops::InheritCoreGraphOps, graph::edges::edge_ref::EdgeEntryRef};
 
 #[derive(Copy, Clone, Debug)]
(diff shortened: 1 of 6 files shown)


raphtory/src/db/graph/views/is_deleted_graph.rs
```diff
@@ -59,22 +56,32 @@ impl<'graph, G: GraphViewOps<'graph>> InheritNodeFilterOps for IsDeletedGraph<G>
...
+    fn internal_edge_list_trusted(&self) -> bool {
         false
     }
 
-    fn internal_filter_edge_layer(&self, edge: EdgeEntryRef, layer: LayerId) -> bool {
+    fn internal_filter_edge(&self, edge: EdgeEntryRef, layer_ids: &LayerIds) -> bool {
         let time_semantics = self.graph.edge_time_semantics();
-        time_semantics.edge_is_deleted(edge, LayeredGraph::new(&self.graph, LayerIds::One(layer)))
-            && self.graph.internal_filter_edge_layer(edge, layer)
+        time_semantics.edge_is_deleted(edge, LayeredGraph::new(&self.graph, layer_ids.clone()))
```
ljeub-pometry: this shouldn't be making a `LayeredGraph`, it should just ignore the input layer ids (no pushdown possible here)

ljeub-pometry (changes requested):

ljeub-pometry (approved):

PR #2773, opened 2026-09-08
Refactor node/edge iterator locking

fabianmurariu:
Take locks for node and edge iterators one segment at a time instead of taking them on every node


db4-storage/src/pages/edge_store.rs
```diff
@@ -644,13 +644,24 @@ impl<ES: EdgeSegmentOps<Extension = EXT>, EXT: PersistenceStrategy<ES = ES>>
...
-                })
+                let locked = Arc::new(page.locked());
+                (0..locked.num_edges())
+                    .map(LocalPOS)
+                    .map(move |pos| SegmentLockedEdgeEntry {
+                        locked: locked.clone(),
+                        pos,
+                    })
+                    .filter(move |edge| {
+                        let edge_ref: <<ES as EdgeSegmentOps>::ArcLockedSegment as LockedESegment>::EntryRef<'_> = edge.as_ref();
```
ljeub-pometry, line 662: The type annotation shouldn't be needed...

graphql-bench/Makefile
```diff
@@ -5,13 +5,44 @@ K6_IP=$(shell terraform output k6_ip | jq -r '.')
...
+# request/response sampling that is switched on emits a whole JSON blob per sample, which reached
+# 200MB uncompressed. Read it with `gzip -dc k6.log.gz | less` (or `| tail -n 100`).
+BUILD_LOG ?= build.log
+SERVER_LOG ?= server.log
+K6_LOG ?= k6.log.gz
+
 build:
-	pnpm install --frozen-lockfile
-	pnpm build
+	@{ pnpm install --no-frozen-lockfile && pnpm build ; } > $(BUILD_LOG) 2>&1 \
```
ljeub-pometry, line 22: shouldn't we keep the frozen, probably don't want to install random new versions of things...

graphql-bench/wait-for-server.sh
```diff
@@ -0,0 +1,31 @@
...
+# checkout and nothing listens on the port until that finishes, which is far longer than any sleep
+# worth hardcoding. k6's setup() calls fail() on the first refused connection, which aborts the
+# whole run before any scenario starts -- so the bench has to be sure the server is up first.
+set -eu
+
+URL="${RAPHTORY_URL:-http://localhost:1736}"
+TIMEOUT="${WAIT_FOR_SERVER_TIMEOUT:-1200}"
+INTERVAL="${WAIT_FOR_SERVER_INTERVAL:-2}"
+
+echo ">>> waiting up to ${TIMEOUT}s for ${URL}"
```
ljeub-pometry, line 14: we should build this into the raphtory cli...

raphtory/src/db/api/view/internal/list_ops.rs
```diff
@@ -128,6 +128,13 @@ impl List<VID> {
         }
     }
 
+    pub fn node_entries(self, g: &GraphStorage) -> impl Iterator<Item = NodeStorageEntry<'_>> {
+        match self {
+            List::All => Either::Left(g.node_entries()),
+            List::List { elems } => Either::Right(elems.into_iter().map(|vid| g.core_node(vid))),
```
ljeub-pometry, line 134: is it worth trying to avoid reacquiring the look for every node here as well by caching the guard of the last locked segment and reusing it if the next node is in the same segment?

raphtory/src/db/graph/nodes.rs
```diff
@@ -209,11 +211,17 @@ where
     ) -> impl Iterator<Item = VID> + Send + Sync + 'graph {
         let view = self.base_graph.clone();
         let selector = self.predicate.clone();
-
-        self.node_list().nodes_iter(&g).filter(move |&vid| {
-            g.try_core_node(vid)
-                .is_some_and(|node| view.filter_node(node.as_ref()) && selector.apply(&g, vid))
-        })
+        let node_list = self.node_list();
+        GenLockedIter::from(
```
ljeub-pometry, line 215: Could we avoid the lifetime issues here by passing in an owned GraphStorage instead of the reference for `node_entries`?

ljeub-pometry (approved):
Some minor comments but can be merged

raphtory/src/db/graph/nodes.rs
fabianmurariu: That won't work because of what node_entries returns

PR #2774, opened 2026-09-09
Compose graph views as a value, applied to the graph once

shivamka1:
Two commits build it: the first adds a set of time ranges and time semantics over several windows,
self-contained and independently tested; the second rewires the filter combinators onto it. Ten
more answer the review, one per comment — see *What the review changed* below.

## What this changes

A window means two unrelated things depending on where it sits, and the lowering could not tell
them apart:

- `Graph.window(0, 5)` — the window **is** the filter, and must narrow the result.
- `Node.window(0, 3).property("p") == 10` — the window is the **scope** the predicate is read in,
  and must restrict nothing else.

Both built a `WindowedGraph`, so no graph could say which it was, and the combinators treated every
operand as the second kind. A graph-level view therefore stopped restricting anything the moment it
was combined.

A view is now a value — `ResolvedView`, a set of time ranges and a set of layers. `and` intersects,
`or` unions, `not` complements, and the result is applied to the base graph **once**. The number of
ranges left picks the graph so it matches `TimeSemantics`: none is an empty window, one is a
`WindowedGraph`, more is a `MultiWindowedGraph` over the new `TimeSemantics::MultiWindow`, whose
semantics are the union of the windowed views over its ranges.

Because it is applied once, the result also carries the right time semantics — which is the second
half of the problem. A wrapper graph inherits its time semantics from the graph underneath, so no
arrangement of wrappers could have produced it.

## Behaviour changes

| expression | before | after |
|---|---|---|
| `window(0,5) & window(3,8)` | in either window | the overlap `[3,5)` |
| `window(0,5) \| window(6,10)` | the whole graph | both ranges, gap excluded |
| `~window(3,8)` | the whole graph | the ranges either side, **event level** |
| `window & predicate` | predicate read on the unwindowed graph; result carried the base graph's history | predicate read inside the window; result carries the window's history |
| `edges[window & predicate]` | same defect on the collection path | fixed with it |
| `window \| layer`, `~(window & layer)` | silently returned everything | `GraphError::InvalidGqlFilter` |

Two of these deserve spelling out.

**Negating a view is now event level.** An edge with events inside and outside a window survives
with the outside ones, rather than being dropped entirely. This is option 1 of #2718, and it
reverses the entity-level reading shipped in #2767.

**Mixing the axes under `|` or `~` is refused.** A resolved view is one set of times by one set of
layers — a rectangle. `window(0,4) | layer("work")` is an L-shape: three of the four (time, layer)
cells qualify, and the nearest rectangle is "everything", which admits events failing *both*
operands. A union is representable when the operands agree on one axis, a complement when at most
one axis is restricted; anything else errors with a message naming the workaround. A list of
rectangles would make this total — it needs per-layer time semantics, tracked in #2776, and turns
on an undecided question about layer-less events (`docs/view-composition.md` §5.2 has the argument).

## API changes

- `GraphFilter` is an empty entry point again, as `NodeFilter` and `EdgeFilter` are.
  `GraphFilter.window(0, 5)` returns a `ViewFilter`, which carries the chain as a list of ops. No
  `::new()` at call sites.
- `DynView` and `DynViewFilter` are removed, along with the parallel builder hierarchy beside them —
  a chain in a `Vec` needs no type-level accumulation.
- `CreateFilter` gains `view_bounds` (what the result contains) and `result_view` (the graph for
  that), beside the existing `filter_graph_view` (where a predicate reads from). The two questions
  are now asked separately.
- `LayerIds` and `Multiple` gain `PartialEq, Eq`. Both are sorted on construction, so structural
  equality is set equality.
- The GraphQL wire form round-trips `before`, `after` and `at` as themselves instead of collapsing
  them into windows.
- `AndFilter`, `OrFilter` and `NotFilter` return `DynGraphArc` rather than a precise type: `&` now
  returns either the composed view or an `AndFilteredGraph` depending on the expression's contents,
  and the view arm picks between three graphs from the number of ranges. Predicates and single views
  keep their precise types. #2776 is the fix — a graph carrying the resolved `TimeSemantics` is all
  four of those graphs at once — and the three declarations point at it.
- `NodeTimeSemanticsWindowOps` and `EdgeTimeSemanticsWindowOps` are new: the 39 `_window` methods
  that used to sit on `NodeTimeSemanticsOps`/`EdgeTimeSemanticsOps`, now implemented only by the
  base semantics.
- `TimeSemantics` gains `base()` and `from_ranges(base, ranges)`; `restrict` is gone, since a view
  resolves its ranges when it is built and then needs no intersection per call.

## Two bugs worth a second look

**An open-ended view left a sliver at the end of time.** `after(8)` was encoded as a window ending
at `i64::MAX` rather than as an open end, so the applied window stopped just short of
`EventTime::MAX`. Harmless until something complemented it — then `~after(8)` gained a second,
one-tick range, and on a **persistent** graph every edge still alive at the end of time fell into
it, so the complement returned the whole graph. `before`, `after` and `at` are now their own ops
calling the graph's own methods.

**A layer complement counted instead of enumerating.** Layer ids are not a contiguous range from
zero, so complementing `0..num_layers` named ids the graph never issued and missed real ones.
`ResolvedView::not` now takes the graph and uses the existing `diff` helper.

## What the review changed

Ten commits, each answering one comment. The theme running through them: **a view resolves its
bounds when it is built, so nothing downstream should resolve them again.**

- **The windowed semantics methods now live on their own traits**, implemented only by
  `EventSemantics`, `PersistentSemantics` and the `BaseTimeSemantics` that dispatches between them.
  `WindowTimeSemantics`, `MultiWindowTimeSemantics` and the `TimeSemantics` enum had to carry 38
  bodies each answering "and what if there is a further window on top?" — a case that cannot arise,
  because nesting is flattened at construction. Verified unreachable before deleting: zero call
  sites outside the module, `filtered_node`/`filtered_edge` included. **−1,331 lines net.**
- **A multi-window view folds the parent's ranges in once, in `new`**, instead of intersecting them
  into fresh `TimeSemantics` on every `node_time_semantics()` call — which runs per node and per
  edge.
- **Its six graph-level `_window` methods pass straight through to the parent**, as `WindowedGraph`
  has always done; they were clipping the caller's already-resolved window against the view's own
  ranges, applying the view twice.
- **`TimeSemantics::window` is allocation-free again from `Base` and `Window`.** My foundation
  commit had routed it through `restrict`, so every `WindowedGraph` semantics call — the commonest
  path in the codebase — allocated a range set and ran a set intersection to do two `max`/`min`.
  That was a regression this PR introduced, caught in review.
- **`TimeRanges` lives behind an `Arc<[Range<EventTime>]>`** with an owning index-based iterator, so
  cloning a range set is a refcount bump and iterating one copies nothing.
- **`union` merges two sorted lists** the way `intersect` sweeps them, instead of concatenating and
  re-sorting; **`earliest_time_global`/`latest_time_global` stop at the first and last range holding
  an event** instead of scanning every range for a `min`/`max`.

## Scope

This fixes the **view** half of Pometry/pometry-storage#371. The node-and-edge-kind half is
untouched and still pinned in `test_broken_combination_classes_are_still_broken`:

- lifting a node filter onto edges under `|` or `~` — `edges[~N]`, `edges[N1 | N2]`;
- Kleene node membership for mixed expressions;
- node filters on the per-node and nested edge paths when the anchor fails the predicate.

Those want deriving from the existing `internal_nodes_filtered` / `internal_edge_filtered` rather
than a declared flag set, and are separate work.

## Testing

Every expectation is built from something other than the path under test — a view from the
equivalent chained view, a predicate from a direct scan. Reading expectations back through the
subscript makes them agree by construction: where a single view filter fails open,
`EVERYTHING & X == X`, so a combination that dropped a term matches its expectation and the suite
reports a live bug as fixed. The oracles assert each baseline is a proper subset up front.

- A table-driven test over **all eight view ops** on both graph models: applying equals the chained
  view, negating equals the views either side, read off the applied view's own bounds.
- **Eight property tests**, 256 cases each: a single view is its chained view (membership and
  history); a conjunction is the two views chained; a disjunction is their union including per-edge
  history; a negation is what lies outside; negation is an involution; a view is idempotent; a
  disjunction commutes; a layer view and its negation match the graph over any subset; mixing the
  axes is refused. Watched failing before being trusted.
- `TimeRanges` against a brute-force point model by proptest, plus De Morgan and involution.
- Oracle rewrites: conjunction expectations now come from the composed view (`window & is_deleted`
  asks which edges are deleted *within* the window), and negated composites from the complement of
  what the composite selects — De Morgan does not hold over entity sets once an entity can belong
  to both a window and its complement.

| suite | result |
|---|---|
| Python, full base install | 3368 passed · 15 skipped · 9 xfailed |
| Rust filter tests | 651 passed |
| Rust graph tests | 198 passed |
| GraphQL library tests | 254 passed |
| `raphtory` unit tests | 108 passed |
| Rust view tests | 47 passed |

`cargo check --workspace --all-targets`, `cargo +nightly fmt` and `black` all clean.

## Notes for reading the diff

The part most worth reviewing is `admits_node` / `admits_edge` and `restricts_every_kind` in
`views/filter/mod.rs`. Four of the five defects found while building this were the same question
asked in different places — *which kinds of entity does this operand restrict, and does a view
count?* — answered inline three times, three different ways. It is now one named rule.

A longer design note lives at `docs/view-composition.md` (local, untracked): the full argument for
the rectangle limit, the five defects with repros, the type-erasure trade-off, and the open
questions.

Refs Pometry/pometry-storage#371, #2718, #2728, #2776.


raphtory/src/db/graph/views/multi_window_graph.rs
```diff
@@ -0,0 +1,663 @@
...
+            .into_iter()
+            .rev()
+            .flat_map(move |w| self.graph.temporal_prop_iter_window_rev(id, w.start, w.end))
+            .into_dyn_boxed()
+    }
+
+    /// The last update at or before `t` in any range: the latest range that
+    /// has one wins, so search the ranges from the back.
+    fn temporal_value_at(&self, id: usize, t: EventTime) -> Option<Prop> {
+        self.windows
```
ljeub-pometry, line 234: we should move these things to the time semantics (make an issue for that) such that we have `graph_time_semantics` here. We should then optimise this to not look at windows that are irrelevant based on the input time.

raphtory/src/db/api/view/internal/time_semantics/time_ranges.rs
```diff
@@ -0,0 +1,327 @@
...
+//! `[0,10)` — so each set has exactly one representation, and the number of
+//! ranges says which `TimeSemantics` variant it is: none is the empty window,
+//! one is `Window`, more is `MultiWindow`.
+
+use raphtory_api::core::storage::timeindex::EventTime;
+use std::ops::Range;
+
+/// Sorted, pairwise disjoint, non-adjacent, non-empty half-open ranges.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct TimeRanges(Vec<Range<EventTime>>);
```
ljeub-pometry: this should contain an `Arc<[Range<EventTime>]>` to be trivial to clone, the `into_iter` can just do `(0..self.len()).map(...)`, also, given how this is used, it might make sense to model this after the `LayerIds` enum (with variants for `All`, `Multi`, `Single`)

raphtory/src/db/api/view/internal/time_semantics/time_ranges.rs
```diff
@@ -0,0 +1,327 @@
...
+            } else {
+                j += 1;
+            }
+        }
+        TimeRanges(out)
+    }
+
+    /// Times in either set.
+    pub fn union(&self, other: &Self) -> Self {
+        let mut ranges = Vec::with_capacity(self.0.len() + other.0.len());
```
ljeub-pometry: this could be a lot more efficient with similar logic to intersect

raphtory/src/db/api/view/internal/time_semantics/multi_window_time_semantics.rs
```diff
@@ -0,0 +1,1144 @@
...
+}
+
+impl MultiWindowTimeSemantics {
+    pub fn windows(&self) -> &TimeRanges {
+        &self.windows
+    }
+
+    /// The ranges of the view, oldest first.
+    fn ranges(&self) -> Ranges {
+        self.windows.clone().into_vec().into_iter()
```
ljeub-pometry: see comment on `TimeRanges`, ideally this shouldn't allocate

raphtory/src/db/api/view/internal/time_semantics/multi_window_time_semantics.rs
```diff
@@ -0,0 +1,1144 @@
...
+    /// The ranges of the view, newest first, for the reverse iterators: the
+    /// per-range streams only concatenate into a sorted stream if the ranges
+    /// themselves come in the same order as the events.
+    fn ranges_rev(&self) -> Rev<Ranges> {
+        self.ranges().rev()
+    }
+
+    /// The ranges of the view intersected with a caller's window, for the
+    /// `_window` family, which asks about a range on top of the view's own.
+    fn clipped(&self, w: &Range<EventTime>) -> Ranges {
```
ljeub-pometry: we should try to avoid doing this on the fly, i.e., can we do this once on constructing the window rather than on every call

raphtory/src/db/api/view/internal/time_semantics/multi_window_time_semantics.rs
```diff
@@ -0,0 +1,1144 @@
...
+        })
+    }
+
+    fn node_earliest_time_window<'graph, G: GraphView + 'graph>(
+        &self,
+        node: NodeStorageRef<'graph>,
+        view: G,
+        w: Range<EventTime>,
+    ) -> Option<EventTime> {
+        self.windows.clipped_to(&w).iter().find_map(|w| {
```
ljeub-pometry: these should be pass-through to the underlying semantics, window nesting and resolution should always be handled when constructing the time semantics, we should never have a window on top of something that is already windowed

raphtory/src/db/api/view/internal/time_semantics/multi_window_time_semantics.rs
```diff
@@ -0,0 +1,1144 @@
...
+        node: NodeStorageRef<'graph>,
+        view: G,
+    ) -> Option<EventTime> {
+        self.windows.iter().rev().find_map(|w| {
+            self.semantics
+                .node_latest_time_window(node, view.clone(), w.clone())
+        })
+    }
+
+    fn node_earliest_time_window<'graph, G: GraphView + 'graph>(
```
ljeub-pometry: It seems that we should move all the `_window` variants to their own trait that is only implemented on the `BaseTimeSemantics`, nested windowing should always be resolved when creating the view, not when calling methods

raphtory/src/db/graph/views/filter/resolved_view.rs
```diff
@@ -0,0 +1,427 @@
...
+            time: self.time.intersect(&other.time),
+            layers: self.layers.intersect(&other.layers),
+        }
+    }
+
+    /// Either restriction. Representable as one (time, layers) pair only when
+    /// the operands agree on one dimension; otherwise the pair
+    /// `(T1 ∪ T2, L1 ∪ L2)` would admit `T1`-events on `L2` layers that neither
+    /// operand admits.
+    pub fn or(&self, other: &Self) -> Result<Self, GraphError> {
```
ljeub-pometry, line 80: we need per-layer time semantics to fix this, ok to error for now

raphtory/src/db/graph/views/filter/resolved_view.rs
```diff
@@ -0,0 +1,427 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::prelude::*;
+    use raphtory_api::core::entities::LayerId;
+    use raphtory_storage::core_ops::CoreGraphOps;
+
+    fn t(v: i64) -> EventTime {
```
ljeub-pometry: useless?

raphtory/src/db/graph/views/filter/resolved_view.rs
```diff
@@ -0,0 +1,427 @@
...
+        EventTime::start(v)
+    }
+
+    fn time(v: &[(i64, i64)]) -> TimeRanges {
+        TimeRanges::new(v.iter().map(|&(a, b)| t(a)..t(b)).collect())
```
ljeub-pometry: can use `EventTime::range(a..b)`, but also, why not use `&[Range<i64>]` as the input?

raphtory/src/db/graph/views/multi_window_graph.rs
```diff
@@ -0,0 +1,663 @@
...
+    fn temporal_value(&self, id: usize) -> Option<Prop> {
+        self.graph.temporal_value_at(id, self.hull_end())
+    }
+
+    /// Union: the per-range histories concatenated in range order.
+    fn temporal_iter(&self, id: usize) -> BoxedLIter<'_, (EventTime, Prop)> {
+        if self.is_empty() {
+            return iter::empty().into_dyn_boxed();
+        }
+        let windows = self.windows.as_slice().to_vec();
```
ljeub-pometry: this iterator borrows the lifetime, you should be able to just use `self.windows.iter()`

raphtory/src/db/graph/views/multi_window_graph.rs
```diff
@@ -0,0 +1,663 @@
...
+            .into_dyn_boxed()
+    }
+
+    /// Union, reversed: the ranges in reverse order, each reversed.
+    fn temporal_iter_rev(&self, id: usize) -> BoxedLIter<'_, (EventTime, Prop)> {
+        if self.is_empty() {
+            return iter::empty().into_dyn_boxed();
+        }
+        let windows = self.windows.as_slice().to_vec();
+        windows
```
ljeub-pometry: `self.windows.iter()`

raphtory/src/db/graph/views/multi_window_graph.rs
```diff
@@ -0,0 +1,663 @@
...
+            self.graph
+                .temporal_prop_ids()
+                .filter(|id| self.has_temporal_prop(*id)),
+        )
+    }
+}
+
+impl<'graph, G: GraphViewOps<'graph>> GraphTimeSemanticsOps for MultiWindowedGraph<G> {
+    fn node_time_semantics(&self) -> TimeSemantics {
+        self.graph
```
ljeub-pometry: we can't do this on the fly (this gets called in tight loops), the TimeRanges in this view should already include any windowing from parent views and this should just make a multi-windowed semantics without re-running the intersection logic

raphtory/src/db/graph/views/multi_window_graph.rs
```diff
@@ -0,0 +1,663 @@
...
+
+    /// Union: the earliest of the per-range earliest times.
+    #[inline]
+    fn earliest_time_global(&self) -> Option<i64> {
+        if self.is_empty() {
+            return None;
+        }
+        self.windows
+            .iter()
+            .filter_map(|w| self.graph.earliest_time_window(w.start, w.end))
```
ljeub-pometry: this can be a `find_map` as windows are ordered, no need for the `min` and iterating over everything

raphtory/src/db/graph/views/multi_window_graph.rs
```diff
@@ -0,0 +1,663 @@
...
+    }
+
+    /// Union: the latest of the per-range latest times.
+    #[inline]
+    fn latest_time_global(&self) -> Option<i64> {
+        if self.is_empty() {
+            return None;
+        }
+        self.windows
+            .iter()
```
ljeub-pometry, line 312: this can be `rev().find_map`, no need for `max` and iterating over everything

raphtory/src/db/graph/views/multi_window_graph.rs
```diff
@@ -0,0 +1,663 @@
...
+            .filter_map(|w| self.graph.latest_time_window(w.start, w.end))
+            .max()
+    }
+
+    /// Union over the ranges clipped to the caller's window: the earliest of
+    /// the per-range earliest times.
+    #[inline]
+    fn earliest_time_window(&self, start: EventTime, end: EventTime) -> Option<i64> {
+        self.windows
```
ljeub-pometry: this can be a straight passthrough, `self.graph.earliest_time_window`, any window nesting should be handled at construction, not on each call

raphtory/src/db/graph/views/multi_window_graph.rs
```diff
@@ -0,0 +1,663 @@
...
+            .clipped_to(&(start..end))
+            .iter()
+            .filter_map(|w| self.graph.earliest_time_window(w.start, w.end))
+            .min()
+    }
+
+    /// Union over the ranges clipped to the caller's window: the latest of the
+    /// per-range latest times.
+    #[inline]
+    fn latest_time_window(&self, start: EventTime, end: EventTime) -> Option<i64> {
```
ljeub-pometry, line 323: this can be a straight passthrough, `self.graph.latest_time_window`, any window nesting should be handled at construction

raphtory/src/db/graph/views/multi_window_graph.rs
```diff
@@ -0,0 +1,663 @@
...
+            .iter()
+            .any(|w| self.graph.has_temporal_prop_window(prop_id, w.clone()))
+    }
+
+    /// Union: the per-range histories concatenated in range order.
+    fn temporal_prop_iter(&self, prop_id: usize) -> BoxedLIter<'_, (EventTime, Prop)> {
+        if self.is_empty() {
+            return iter::empty().into_dyn_boxed();
+        }
+        let windows = self.windows.as_slice().to_vec();
```
ljeub-pometry: `self.windows.iter()`, this already borrows

raphtory/src/db/graph/views/multi_window_graph.rs
```diff
@@ -0,0 +1,663 @@
...
+            .flat_map(move |w| {
+                self.graph
+                    .temporal_prop_iter_window(prop_id, w.start, w.end)
+            })
+            .into_dyn_boxed()
+    }
+
+    /// Union over the ranges clipped to the caller's window: an update in any
+    /// of them is enough.
+    fn has_temporal_prop_window(&self, prop_id: usize, w: Range<EventTime>) -> bool {
```
ljeub-pometry, line 348: straight passthrough for windowing, see other comments

raphtory/src/db/graph/views/multi_window_graph.rs
```diff
@@ -0,0 +1,663 @@
...
+    }
+
+    /// Union over the ranges clipped to the caller's window, concatenated in
+    /// range order.
+    fn temporal_prop_iter_window(
+        &self,
+        prop_id: usize,
+        start: EventTime,
+        end: EventTime,
+    ) -> BoxedLIter<'_, (EventTime, Prop)> {
```
ljeub-pometry, line 357: straight passthrough for windowing, see other comments

raphtory/src/db/graph/views/multi_window_graph.rs
```diff
@@ -0,0 +1,663 @@
...
+            .into_dyn_boxed()
+    }
+
+    /// The same, reversed: the clipped ranges in reverse order, each reversed.
+    fn temporal_prop_iter_window_rev(
+        &self,
+        prop_id: usize,
+        start: EventTime,
+        end: EventTime,
+    ) -> BoxedLIter<'_, (EventTime, Prop)> {
```
ljeub-pometry, line 366: straight passthrough for windowing, see other comments

raphtory/src/db/graph/views/multi_window_graph.rs
```diff
@@ -0,0 +1,663 @@
...
+    }
+
+    /// The same, over the ranges clipped to the caller's window.
+    fn temporal_prop_last_at_window(
+        &self,
+        prop_id: usize,
+        t: EventTime,
+        w: Range<EventTime>,
+    ) -> Option<(EventTime, Prop)> {
+        self.windows
```
ljeub-pometry: straight passthrough for windowing, see other comments

raphtory/src/db/graph/views/multi_window_graph.rs
```diff
@@ -0,0 +1,663 @@
...
+                edge::EdgeView,
+                views::{deletion_graph::PersistentGraph, window_graph::WindowedGraph},
+            },
+        },
+        prelude::*,
+    };
+    use raphtory_api::core::storage::timeindex::{AsTime, EventTime};
+    use std::collections::BTreeSet;
+
+    fn r(a: i64, b: i64) -> std::ops::Range<EventTime> {
```
ljeub-pometry: replace with `EventTime::range(a..b)`

raphtory/src/db/api/view/internal/time_semantics/multi_window_time_semantics.rs
```diff
@@ -0,0 +1,1144 @@
...
+        self,
+        node: NodeStorageRef<'graph>,
+        view: G,
+        layer_ids: &'graph LayerIds,
+        w: Range<EventTime>,
+    ) -> impl Iterator<Item = EventTime> + Send + Sync + 'graph {
+        let semantics = self.semantics;
+        self.clipped(&w)
+            .flat_map(move |w| semantics.node_history_window(node, view.clone(), layer_ids, w))
+    }
```
ljeub-pometry: We should separate out the `_window` methods into their own trait which is only implemented on the `BaseTimeSemantics`, we do not have nested windows at the `TimeSemantics` level, they are always flattened out at construction. This is causing a lot of code that is unreachable.

raphtory/src/db/graph/views/filter/model/or_filter.rs
```diff
@@ -35,49 +36,62 @@ impl<L, R> ComposableFilter for OrFilter<L, R> {}
...
-        let r = self.right.filter_graph_view(filtered.clone())?;
-        let left = self.left.create_filter(filtered.clone(), l)?;
-        let right = self.right.create_filter(filtered, r)?;
-        Ok(OrFilteredGraph { graph, left, right })
+        _filtered: F,
+    ) -> Result<Self::EntityFiltered<'graph, G, F>, GraphError>
+    where
+        Self: 'graph,
+    {
+        let bounds = self.view_bounds(graph.clone())?;
```
ljeub-pometry, line 69: If we add the resolved `TimeSemantics` to the `AndFilteredGraph`, `OrFilteredGraph` and `NotFilteredGraph`, we no longer need to dispatch to different dynamic views here. To make that easy and avoid redundant code, this would mean that we move all the graph-related methods from `GraphTimeSemanticsOps` to a new trait that is implemented for `TimeSemantics` (Same as `NodeTimeSemanticsOps` and `EdgeTimeSemanticsOps`) and use that in the implementation for the graph properties.

ljeub-pometry (changes requested):

raphtory/src/db/api/view/internal/time_semantics/time_ranges.rs
shivamka1: `Arc<[Range<EventTime>]>` plus an owning index-based iterator in cdfa78967, so cloning a range set is a refcount bump and `MultiWindowTimeSemantics::ranges()` no longer copies into a `Vec`.

On modelling it after `LayerIds`: `TimeSemantics` is already that enum — `Base` (all time), `Window` (one range), `MultiWindow` (several) — and `TimeRanges` is the payload of the last arm, picked by `TimeSemantics::from_ranges` from the range count. Adding `All`/`Single`/`Multi` inside the payload would encode the same distinction in two places that have to be kept in step. Happy to build it if you would still rather have it there.

raphtory/src/db/graph/views/multi_window_graph.rs
shivamka1: Filed as #2776: the graph-level methods move onto `TimeSemantics` so views expose `graph_time_semantics()` and delegate, instead of each view answering the eleven temporal questions against its own notion of a window — and with the ranges inside the semantics, a windowed question can binary-search the sorted set rather than clipping every range. Referenced here in a0a4638e9.

raphtory/src/db/graph/views/filter/model/or_filter.rs
shivamka1: Filed as #2776, with this as the reason it matters: the erasure exists because a resolved view becomes one of `WindowedGraph`, `MultiWindowedGraph`, `LayeredGraph` or the graph itself, and no single associated type names all four. A graph carrying the resolved `TimeSemantics` is all four at once, so `EntityFiltered` can be concrete and the vtable goes. Referenced at all three declarations in a0a4638e9.

raphtory/src/db/graph/views/filter/resolved_view.rs
shivamka1: Kept as an error, and captured in #2776: the representation that would make it total is a list of `(TimeRanges, LayerIds)` rectangles, which needs per-layer time semantics. The issue also records the decision that gates it — node property updates live on the reserved static layer, so whether a layer view restricts layer-less events decides whether `layer("work")` and `window(0,4) | layer("work")` can agree about a node property update outside the window. Referenced at the refusal in a0a4638e9.

PR #2777, opened 2026-09-11
make hashing work with the new relaxed property comparison semantics

ljeub-pometry:

- Wrapper type for hashing `Prop` with semantics that are compatible with `Prop::equals`

- filtering with `is_in` and `is_not_in` was extremely slow with the new relaxed semantics


fabianmurariu (approved):
LGTM

PR #2783, opened 2026-09-14
Reworking the label propagation algorithm

rodrigolece:

The label propagation algorithm is getting:
- A new initialisation. We can pass `init_state` as a map of node labels, with the idea that we will typically want to do this with an incomplete list of nodes. In this case all the nodes not in the list are considered as having `NO_LABEL` which is a sentinel. Unlabelled nodes don't cast a vote.
- An automatic stopping based on the `rel_tol` and `patience` parameters. The algorithm returns after `patience` iterations where the number of nodes changing isn't more than the `rel_tol` fraction.
- A more sophisticated way of breaking up vote ties. Initially, we were using a hash to randomly break ties between votes, and because this was inefficient it's now combination of hashing and a affine mapping.
- A fast version, which is a complete re-write that doesn't use the Tasks API (`ATask`, `TaskRunner`, etc..).

Both for performance and usability. The partial initialisation covers an important use case for us.

Yes, the signature of the `label_propagation` function is changing. We are also adding the function `label_propagation_fast`. It's documented in the docstrings.

Some of the changes rely on the existing `lpa_test` (initialised with a different seed in order to recover the expected partitions), while the new fast version has a new test `lpa_fast_matches_lpa`.

Not that we are aware of, but surely there will be things to clean up.

### DIFF
--- a/raphtory/src/algorithms/community_detection/label_propagation.rs
+++ b/raphtory/src/algorithms/community_detection/label_propagation.rs
@@ -139,17 +139,20 @@ where
     // `patience` iterations since the improvement was no better than `rel_tol`.
     let rel_tol = rel_tol.unwrap_or(3e-4);
     let patience = patience.unwrap_or(10);
-    // (best, stale): Check is Fn + called once/iter single-threaded, so the Mutex is uncontended
-    let convergence_state = Arc::new(Mutex::new((usize::MAX, 0usize)));
+    // (best, stale, n_iter): Check is Fn + called once/iter single-threaded, so the Mutex is uncontended
+    let convergence_state = Arc::new(Mutex::new((usize::MAX, 0usize, 0usize)));
     let step4 = Job::Check(Box::new(move |state: &GlobalState<ComputeStateVec>| {
         let diff = state.read(&global_diff);
-        let (best, stale) = &mut *convergence_state.lock().unwrap();
+        let (best, stale, n_iter) = &mut *convergence_state.lock().unwrap();
+        *n_iter += 1;
         // check for improvement
         let improved = (diff as f64) < (*best as f64) * (1.0 - rel_tol);
         *best = (*best).min(diff);
         *stale = if improved { 0 } else { *stale + 1 };
         // Stop once fully converged (diff == 0) or the changed-node count has plateaued.
         if diff == 0 || *stale >= patience {
+            let pct = 100.0 * diff as f64 / num_nodes as f64;
+            println!("label_propagation: stopped after {n_iter} iters; diff={diff} ({pct:.2}%)");
             Step::Done
         } else {
             Step::Continue
--- a/raphtory/src/algorithms/community_detection/label_propagation.rs
+++ b/raphtory/src/algorithms/community_detection/label_propagation.rs
@@ -16,12 +16,18 @@ use crate::{
 };
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc,
+};
 
 #[derive(Clone, PartialEq, Serialize, Deserialize, Debug, Default)]
 pub struct LabelPropState {
     #[serde(skip)]
     nbors: HashMap<usize, usize>,
     pub community_id: usize,
+    #[serde(skip)]
+    is_changed: bool, // derive(Default) initializes to false
 }
 
 /// Computes components using a label propagation algorithm
@@ -52,6 +58,10 @@ where
     let global_diff = accumulators::sum::<usize>(2);
     ctx.global_agg_reset(global_diff);
 
+    let num_nodes = g.count_nodes();
+    let active: Arc<Vec<AtomicBool>> =
+        Arc::new((0..num_nodes).map(|_| AtomicBool::new(false)).collect());
+
     let step1 = ATask::new(move |s| {
         let id = s.node.index();
         let state: &mut LabelPropState = s.get_mut();
@@ -59,10 +69,29 @@ where
             .as_ref()
             .and_then(|map| map.get(&id).copied())
             .unwrap_or(id);
+        state.is_changed = true; // the actual initialization
         Step::Continue
     });
 
+    let active_step2 = Arc::clone(&active);
     let step2 = ATask::new(move |s: &mut EvalNodeView<_, LabelPropState>| {
+        if s.prev().is_changed {
+            for nbor in s.neighbours() {
+                active_step2[nbor.state_pos].store(true, Ordering::Relaxed);
+            }
+        }
+        Step::Continue
+    });
+
+    let active_step3 = Arc::clone(&active);
+    let step3 = ATask::new(move |s: &mut EvalNodeView<_, LabelPropState>| {
+        // Gate: consume this node's activation flag atomically.
+        if !active_step3[s.state_pos].swap(false, Ordering::AcqRel) {
+            s.get_mut().is_changed = false;
+            // NB: this leaves community_id unchanged
+            return Step::Continue;
+        }
+
         let prev_id = s.prev().community_id;
         let nbor_iter = s.neighbours();
         let state: &mut LabelPropState = s.get_mut();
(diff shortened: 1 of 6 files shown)


python/python/raphtory/algorithms/__init__.pyi
```diff
@@ -70,6 +70,7 @@ __all__ = [
     "hits",
     "balance",
     "label_propagation",
+    "label_propagation_fast",
```
ljeub-pometry, line 73: We shouldn't have two versions? Who would want to run the slow one?

raphtory/src/algorithms/community_detection/label_propagation.rs
```diff
@@ -110,3 +281,369 @@ where
...
+                .enumerate()
+                .map(|(slot, &label)| (label, slot))
+                .collect();
+            (labels, slot_of)
+        }
+        // Unseeded
+        // The position -> VID table, which is what makes `tie_rank` hash the VID here exactly as
+        // the task version does. Filled from the same `(pos, vid)` pairs every other pass uses,
+        // rather than from `Index::iter` (whose order would have to be assumed to agree) or from
+        // `Index::value` (which is `iter().nth(i)` on a full index, so O(n) per lookup).
```
ljeub-pometry, line 408: I'm pretty sure that is optimised to `O(log(n))` (not that you should use it but this comment is probably wrong)

raphtory/src/algorithms/community_detection/label_propagation.rs
```diff
@@ -110,3 +281,369 @@ where
...
+                .iter()
+                .enumerate()
+                .map(|(slot, &label)| (label, slot))
+                .collect();
+            (labels, slot_of)
+        }
+        // Unseeded
+        // The position -> VID table, which is what makes `tie_rank` hash the VID here exactly as
+        // the task version does. Filled from the same `(pos, vid)` pairs every other pass uses,
+        // rather than from `Index::iter` (whose order would have to be assumed to agree) or from
```
ljeub-pometry, line 407: `Index::iter` can be assumed to agree with the `par_iter` output (that is the whole point...). The actual initialisation below is fine, just delete these comments.

raphtory/src/algorithms/community_detection/label_propagation.rs
```diff
@@ -110,3 +281,369 @@ where
...
+    // Unseeded leaves every entry at 0, deliberately. The task version does not set a confidence
+    // when it seeds that case either, so an `iter_count` of 0 reports 0.0 on both sides; and from
+    // the first sweep on, every entry is overwritten, because a labelled node always casts a
+    // self-vote and so never takes the `total == 0` branch.
+    if let Some(map) = &init_state {
+        index.par_iter().for_each(|(pos, vid)| {
+            if let Some(slot) = map.get(&vid.index()).and_then(|l| slot_of.get(l)) {
+                prev[pos].store(*slot, Ordering::Relaxed);
+                // A seed's label is GIVEN, not inferred, so it starts fully confident: 1 of 1.
+                votes[pos].store(1 << 32 | 1, Ordering::Relaxed);
```
ljeub-pometry, line 463: We might have to worry about a node having more than 4 billion neighbours at some point. Given that this is only stored into, you could just use `to_bits` and `from_bits` to pack a float into this atomic instead.

raphtory/src/algorithms/community_detection/label_propagation.rs
```diff
@@ -110,3 +281,369 @@ where
...
+    // pass to collect. Unseeded: everything, matching the task version setting `is_changed` on
+    // every node, so there is nothing to collect and the pass is skipped.
+    let mut active_cur: Vec<AtomicBool> = (0..n).map(|_| AtomicBool::new(!seeded)).collect();
+    let mut active_next: Vec<AtomicBool> = (0..n).map(|_| AtomicBool::new(false)).collect();
+    if seeded {
+        index
+            .par_iter()
+            .for_each_init(Vec::new, |nbors, (pos, vid)| {
+                if prev[pos].load(Ordering::Relaxed) != NO_LABEL {
+                    collect_nbors(vid, nbors); // collects into nbors
```
ljeub-pometry: Why the hell are we collecting this into a vec? We just seem to be iterating over it...

raphtory/src/algorithms/community_detection/label_propagation.rs
```diff
@@ -110,3 +281,369 @@ where
...
+///
+/// # Returns
+///
+/// A `TypedNodeState` mapping each node to its `LabelPropState`, exactly as [`label_propagation`].
+pub fn label_propagation_fast<G>(
+    g: &G,
+    iter_count: usize,
+    seed: Option<u64>,
+    threads: Option<usize>,
+    init_state: Option<HashMap<usize, usize>>,
```
ljeub-pometry, line 374: this input is bizarre for a public api, this should be a map from nodes to labels, something like `HashMap<V, usize>` where `V: AsNodeRef`

raphtory/src/algorithms/community_detection/label_propagation.rs
```diff
@@ -110,3 +281,369 @@ where
...
+                };
+                (counts, Vec::new(), Vec::new())
+            },
+            |(counts, touched, nbors), (pos, vid)| {
+                let prev_slot = prev[pos].load(Ordering::Relaxed);
+                // Gate: consume this node's activation flag atomically. A node that does not run
+                // still has to carry its label over, which the task version gets for free from the
+                // runner copying `prev` into `cur` between supersteps.
+                if !active_cur[pos].swap(false, Ordering::AcqRel) {
+                    cur[pos].store(prev_slot, Ordering::Relaxed);
```
ljeub-pometry, line 510: This version is still touching all the nodes on each iteration, so whatever overhead there is in the task apis, it is something else.
These two values already agree if a node didn't change its label, thus this update is only needed on nodes that were active in the previous step but are not active this step. I think we can make this even faster by pushing newly activated nodes into a sharded list and only iterate over those on the next iteration. We would also have to iterate over the nodes that were active in the current iteration but are not active in the next and copy over their labels.

raphtory/src/algorithms/community_detection/label_propagation.rs
```diff
@@ -110,3 +281,369 @@ where
...
+
+    // Synchronous LPA never reaches global_diff == 0 on graphs with locally-bipartite pockets
+    // (results in ~period-2 oscillations), so the stopping criterion we use is to wait for
+    // `patience` iterations since the improvement was no better than `rel_tol`.
+    let pool: Arc<rayon::ThreadPool> = threads.map(custom_pool).unwrap_or_else(|| POOL.clone());
+    pool.install(|| {
+        let mut best = usize::MAX;
+        let mut stale = 0usize;
+        for n_iter in 1..=iter_count {
+            let diff = sweep(&prev, &cur, &active_cur, &active_next);
```
ljeub-pometry, line 594: Just inline the code for sweep here! We are not using it anywhere else and thus there is zero point in making it a closure!

ljeub-pometry (changes requested):

ljeub-pometry:
To add to the review comments above, what needs to happen to merge this:
 - Remove the old slow version if it is actually slower (what benchmark are we using to test this?)
 - Fix the input for the seeded version (it shouldn't rely on the algorithm-internal positions). In an ideal case, you should be able to put the output of a run of this algorithm back in as a seed with minimal fuss...
 - Inline the pointless sweep closure
 - Fix the handling of the confidence as per the comment so we don't have to worry about it breaking on some monster node.

The other comments are future optimisation suggestions.

rodrigolece:
Thank you very much for your review @ljeub-pometry !

From the final list, I think I've addressed all the points except the one about removing the slow version. To answer your question, the benchmark that I was using is a simple python script running in our remote box, against one of our larger graphs (starts with the letter E). I'd much rather we do this with a proper benchmark, and I can see that the repo has a surface to do this. I will pick your brain about it, and then I'm happy to sunset the slow version once we have properly benchmarked it.

PR #2787, opened 2026-09-18
quick fix for the subgraph creation deadlock

ljeub-pometry:

Fix deadlock in the server when creating a subgraph for a graph that is not currently in the cache.

No

Added test that reproduced the deadlock

We need to review the mutation locking and dirty path cleanup logic. The root cause is that the read path currently also has to acquire the read lock for the mutations as it cleans up dirty paths.


shivamka1 (approved):
LGTM. Probably we should have an issue to fix this.

fabianmurariu (approved):
LGTM

fabianmurariu (approved):
LGTM

PR #2796, opened 2026-09-18
Fixed latest time event id

AndreaTocchettiPometry:


fabianmurariu:
We won't fix this issue like this, we need a graph wide time history.

PR #2797, opened 2026-09-18
Fix exploded edge latest timestap

AndreaTocchettiPometry:


ljeub-pometry (changes requested):
The fix here is wrong, returning `EventTime::end` is semantically appropriate as the exact event id is unknown (we do not have a global history on the graph right now). Returning event id 0 here would mean that there are likely events in the graph that happen after the end of this edge which should not be the case. If this is inconsistent with `graph.latest_time`, we should fix that there instead.

